PR #122, opened 2017-01-12
adds std arch draft

cc10512:
@dandaly: adding the source for the PSA spec as a separate directory in the p4-16 spec tree.

@jnfoster, @ChrisDodd, @mbudiu-vmw, @vgurevich, @gbrebner, @chkim4142: FYI.


jafingerhut:
Is this PR obsoleted by the following one on p4lang/p4c?  https://github.com/p4lang/p4c/pull/569

cc10512:
@jafingerhut: I synchronized the two. I was hoping to get doxygen to produce the spec, but unfortunately P4 is not a language understood by doxygen and the result is very poor. However, I believe we need both a formal spec and a commented architecture file, so I will keep them in sync until I find a better way of generating the spec and the header out of a single source.

PR #165, opened 2017-04-18
Minor clarification that default_action is not mandatory

jafingerhut:
except to prevent undefined behavior as later described.

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -3325,7 +3325,7 @@ table t(inout bit<32> z) {
 
 The default action of a table is an action that is invoked automatically by the match-action unit whenever the lookup table does not find a match for the supplied key.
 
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -3325,7 +3325,7 @@ table t(inout bit<32> z) {
...
+The initial value for the default action is supplied as a value for the ```default_action``` property; if you wish to avoid the undefined behavior described below, this property must be present for a table. The default action _must_ appear after the ```action``` property and may be declared as ```const```, indicating that it cannot be changed dynamically by the control-plane. The default action _must_ be one of the actions that appear in the actions list. In particular, the expressions passed as ```in, out``` or ```inout``` parameters must be syntactically identical to the expressions used in one of the elements of the ```actions``` list.
```
jnfoster: We should reword this -- it's not undefined behavior anymore, but a no-op.

jafingerhut: I would be in favor of closing this PR in favor of a different one, @jnfoster, if you are working on one that addresses the latest working group meeting decisions.

jnfoster: Sure, we can close this PR. I will create a new one (not until tomorrow morning though.)

Issue #163, opened 2017-04-16
Why have implicit casts for assignment?


jafingerhut:
The code snippet below gives no error or warning from the current compiler, and the spec even implies that it should be OK with the statement: "Implicit casts are allowed ... In assignments when the right-hand side has a different type from left-hand side." (Section 8.9.2 "Implicit Casts"):
```
    bit<16> y;
    bit<8> x;
    x = y;
```
This surprises me quite a bit.  Why do we think programmers will make fewer mistakes of mismatching widths during assignment than in all the other situations where the spec and compiler say it is an error, i.e. during arithmetic ops, or passing arguments to actions, control blocks, etc.?

I for one would prefer it if such assignments were an error.

jnfoster:
Close by #173

PR #173, opened 2017-04-25
Change assignments to use implicit cast only when right-hand side is int

jafingerhut:


jnfoster:
Closes #163

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -1962,7 +1962,7 @@ Unlike C, P4 allows a very limited number of implicit casts. The reason is that
 Implicit casts are allowed in P4 only when their meaning is completely unambiguous:
 
 - To convert an ```int``` value to a fixed-width type.
-- In assignments when the right-hand side has a different type from left-hand side.
+- In assignments when the right-hand side has type ```int```.
```
jafingerhut: @jnfoster - The current p4c gives a warning about the RHS not being able to fit in the LHS, which makes me happy.  If you want, I can add a sentence in the spec recommending that the compiler should give a warning in this case.

jnfoster: Sure, I think a note suggesting (maybe not mandating) a warning is fine.

PR #126, opened 2017-03-13
v1.0.4

jnfoster:
Draft proposal for v1.0.4 maintenance release.

PR #174, opened 2017-04-25
Proposed updates for v1.0.4 for modify_field and friends

jafingerhut:
Indicating that they leave the destination with an undefined value if
any of the source operands are fields in invalid headers.

There are also a few TBD comments added in places where I wonder if
the text is correct according to the current P4-14 compiler and bmv2
behavior.

### DIFF
--- a/p4-14/v1.0.3/tex/p4.tex
+++ b/p4-14/v1.0.3/tex/p4.tex
@@ -556,7 +556,7 @@ with the \texttt{metadata} keyword) differ only in their validity. Packet header
 maintain a separate valid indication which may be tested explicitly. Metadata 
 is always considered to be valid. This is further explained in 
 Section~\ref{sec:testing}.  Metadata instances are 
-initialized to 0 by default.
+initialized to 0 by default.  (TBD: Is this true?  Or are metadata instances uninitialized and contain undefined values initially?)
 %%, but initial values may be specified in their declaration.
 
 The BNF for header and metadata instances is:
@@ -658,7 +658,7 @@ represent that field's validity is a reasonable work around.
 Only valid packet header fields may result in a match (when a value is
 specified for exact or ternary matches against the field), although a
 match operation may explicitly check if a header instance (or field)
-is valid. Only valid packet headers are considered for deparsing (see
+is valid (TBD: Should this be updated as asked elsewhere?). Only valid packet headers are considered for deparsing (see
 Section~\ref{sec:deparsing}).  
 
 \SUBSUBSECTION{Header Stacks}{headerstacks}
@@ -798,7 +798,7 @@ The \texttt{output_width} value is in bits.
 
 A field instance is excluded from the calculation (i.e., it is treated as 
 if the instance is not listed in the input list) if the field's header 
-is not valid.
+is not valid.  (TBD: Is this true, or should the result of the calculation be undefined, since it will be fed an undefined value for that header field?)
 
 The algorithm is specified as a string.  The following algorithms are
 defined with the given names, and targets may support others.
@@ -886,7 +886,7 @@ The \texttt{verify} option indicates that the parser should calculate the expect
 value and check if that value is stored in the indicated field. If the value 
 is not equal, then a \texttt{p4_pe_checksum} exception is generated; see 6.6.1. 
 Standard Parser Exceptions. This check occurs at the end of parsing and is 
-performed only if \texttt{field_ref} is valid.
+performed only if \texttt{field_ref} is valid.  (TBD: True, or should this be changed to say that a verify operation on a field in an invalid header has undefined behavior?)
 
 The \texttt{update} option indicates that the system should update the value of the 
 field if changes are made to any fields on which it depends. The update to 
@@ -1832,7 +1832,7 @@ Add a header to the packet's Parsed Representation
 { % Description
 If the \texttt{header_instance} is not an element in a header stack, the indicated 
 header instance is set valid. If the header instance was invalid, all its 
-fields are initialized to 0. If the header instance is already valid, it is 
+fields are initialized to 0 (TBD: Should this be updated to 'all its fields contain unspecified values'?). If the header instance is already valid, it is 
 not changed.
 
 If \texttt{header_instance} is an element in a header stack, the effect is to push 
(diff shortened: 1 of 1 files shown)


jnfoster:
@antoninbas are you able to comment on these TBDs?

antoninbas:
I will comment with respect to bmv2

p4-14/v1.0.3/tex/p4.tex
```diff
@@ -556,7 +556,7 @@
 maintain a separate valid indication which may be tested explicitly. Metadata 
 is always considered to be valid. This is further explained in 
 Section~\ref{sec:testing}.  Metadata instances are 
-initialized to 0 by default.
+initialized to 0 by default.  (TBD: Is this true?  Or are metadata instances uninitialized and contain undefined values initially?)
```
antoninbas: bmv2 simple_switch initializes all metadata fields to 0 to be compatible with P4 programs and apps that were written with this assumption.

p4-14/v1.0.3/tex/p4.tex
```diff
@@ -658,7 +658,7 @@
 Only valid packet header fields may result in a match (when a value is
 specified for exact or ternary matches against the field), although a
 match operation may explicitly check if a header instance (or field)
-is valid. Only valid packet headers are considered for deparsing (see
+is valid (TBD: Should this be updated as asked elsewhere?). Only valid packet headers are considered for deparsing (see
```
antoninbas: I'm in favor of amending this statement ("Only valid packet header fields may result in a match") for the following reasons:
- P4_14 offers a way to explicitly match of header validity
- this seems to be the direction p4_16 has chosen
- bmv2 does not check for validity in a match unless explicitly asked in the P4 program

p4-14/v1.0.3/tex/p4.tex
```diff
@@ -798,7 +798,7 @@
 
 A field instance is excluded from the calculation (i.e., it is treated as 
 if the instance is not listed in the input list) if the field's header 
-is not valid.
+is not valid.  (TBD: Is this true, or should the result of the calculation be undefined, since it will be fed an undefined value for that header field?)
```
antoninbas: bmv2 skips invalid fields / headers in calculations, as this has proved very convenient in P4 programs which hash on L4 fields
this is also consistent with deparsing. IMO a field list calculation is just: 1) serializing the fields to a buffer (just like deparsing), and 2) applying a calculation over the buffer.

p4-14/v1.0.3/tex/p4.tex
```diff
@@ -1832,7 +1832,7 @@
 { % Description
 If the \texttt{header_instance} is not an element in a header stack, the indicated 
 header instance is set valid. If the header instance was invalid, all its 
-fields are initialized to 0. If the header instance is already valid, it is 
+fields are initialized to 0 (TBD: Should this be updated to 'all its fields contain unspecified values'?). If the header instance is already valid, it is
```
antoninbas: Maybe we should try to guarantee that a valid header does not have undefined field values, by initializing them to 0.

p4-14/v1.0.3/tex/p4.tex
```diff
@@ -2885,7 +2892,7 @@
 of \texttt{1} will match when the header is valid; \texttt{0} will match when the header is 
 not valid. As a reminder, the table does not have to explicitly include a 
 match on a field's validity to safely match on its value - invalid fields 
-will never match on a table entry that includes it. Note that metadata fields 
+will never match on a table entry that includes it (TBD: Is this true?  Or is it more accurate to say something like ``invalid fields cause the search key to contain an undefined value in those bit positions, and whether they match a table entry or not is thus unspecified''?). Note that metadata fields
```
antoninbas: I think I commented on this above already.

p4-14/v1.0.3/tex/p4.tex
```diff
@@ -886,7 +886,7 @@
 value and check if that value is stored in the indicated field. If the value 
 is not equal, then a \texttt{p4_pe_checksum} exception is generated; see 6.6.1. 
 Standard Parser Exceptions. This check occurs at the end of parsing and is 
-performed only if \texttt{field_ref} is valid.
+performed only if \texttt{field_ref} is valid.  (TBD: True, or should this be changed to say that a verify operation on a field in an invalid header has undefined behavior?)
```
antoninbas: I have maybe a small preference for the existing text.

jnfoster:
@jafingerhut Minor point, but I believe the .tex is generated from the .pt file (see https://github.com/p4lang/p4-spec/blob/master/p4-14/v1.0.3/tex/Makefile#L7)

I'm not sure if this PR was intended to be merged or if it's just scratch space for a discussion that will eventually go onto #126?

-N

p4-14/v1.0.3/tex/p4.tex
jafingerhut: @jnfoster updated this text already in his v1.0.4 branch, to delete the sentence that began "As a reminder, ...".    Looks good to me.

p4-14/v1.0.3/tex/p4.tex
jafingerhut: Both P4-14 and P4-16 specs say that the value of fields in an invalid header are undefined.  I think there is strong motivation from multiple P4 implementers not to change that.

p4-14/v1.0.3/tex/p4.tex
jafingerhut: OK

p4-14/v1.0.3/tex/p4.tex
jafingerhut: OK

p4-14/v1.0.3/tex/p4.tex
jafingerhut: I will submit a PR updating this statement, to say that the fields of a search key created from header fields in invalid headers contain undefined values.

p4-14/v1.0.3/tex/p4.tex
jafingerhut: Sounds OK to me.

jafingerhut:
@jnfoster I believe all of my proposed changes for P4-14 v1.0.4 spec are now pointed to by comments in PR https://github.com/p4lang/p4-spec/pull/126

This one was useful for the discussion of some of my questions, but we don't need to keep it around once all the changes on the other PR are done.

PR #184, opened 2017-05-08
removes priorities from table entries.

### LINKED ISSUE #168: Restrict constant tables to all explicit or no explicit priorities across all entries?
I would guess that when creating a constant ternary or range table, people would adopt one of two possibilities:

+ No explicit @priority annotations for any entry of the table, allowing the order of entries in the code to specify relative priority.
+ Every entry of the table has an explicit @priority statement.

Allowing a mix of some of each seems only to raise questions of precisely what it means for them to be mixed together in the same table.  The existing example seems to imply that the i-th entry (i starting from 1 and incrementing with each new entry) is the default value of @priority if none is given, and @priority overwrites that, but it wasn't clear from my reading if that is exactly what is intended.

If people think a mix of explicit and implicit @priority values is useful, I don't have a strong objection to allowing them -- it just seems like a poor coding practice that imposes unnecessary mental effort on one trying to understand the precise behavior.

Maybe the summary is: If this is allowed, I want a lint tool that recommends I avoid mixing such entries if it finds I am doing so.

cc10512:
Fixes #168.

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -3415,15 +3415,9 @@ table keys (see Sec. [#sec-table-props]). The table key type must match the
 type of the element of the set. ```actionRef``` must be an action which appears
 in the table actions list, with all its arguments bound.
 
-Entries in a table are matched in a priority order, where lower numbers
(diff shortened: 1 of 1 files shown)


cc10512:
@jafingerhut: indeed, that's exactly what I did.

PR #188, opened 2017-05-09
Clarify the behavior of direct type invocation.

cole-barefoot:


jnfoster:
Looks great, thanks!

cole-barefoot:
Creating a new local instantiation per type application will result in multiple instances with the same control API name, which is wrong.  I think this is a bug in the current implementation, actually.

Instead of rejecting duplicate type applications in the same scope, we could instead say they all refer to the same local instantiation.  I don’t really have a preference—do you?

mbudiu-vmw:
Controls themselves are not exposed to the control plane. Only if a control contains objects visible to the control plane this matters. Controls can be used just as functions.

The restriction on unique naming already takes care of that case, so perhaps we don't need to make it illegal here.

PR #189, opened 2017-05-09
Add paragraph explaining that table properties can change their seman…

jafingerhut:
…tics

Also replaced the word 'attribute' when it was referring to a table
'property', to make the word 'property' consistently used for that
thing everywhere.

Fixed a couple occurrences of 'action' that should have been the
standard table property name 'actions'.

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -2300,7 +2300,7 @@ consequence, all constructor arguments must also be expressions that can be
 evaluated at compilation time.
 
 The following example shows a constructor invocation for setting the
-target-dependent implementation attribute of a table:
+target-dependent implementation property of a table:
 ~ Begin P4Example
 extern ActionProfile {
     ActionProfile(bit<32> size);  // constructor
@@ -3239,7 +3239,7 @@ The standard table properties are the following:
 
 We proceed discussing each of these.
 
-A property marked as ```const``` cannot be changed dynamically by the control-plane. The key and actions properties are always constant, so the ```const``` keyword is not needed for these.
+A property marked as ```const``` cannot be changed dynamically by the control-plane. The ```key``` and ```actions``` properties are always constant, so the ```const``` keyword is not needed for these.
 
 ###	Table properties { #sec-table-props }
 
@@ -3287,13 +3287,13 @@ match_kind {
 
 These identifiers correspond to the P4~14~ match kinds with the same names. The semantics of these annotations is actually irrelevant for describing the behavior of the P4 abstract machine; how they are used influences only the control-plane API and the actual implementation of the look-up table. From the point of view of the P4 program, a look-up table is an abstract finite map that is given a key and produces as a result either an action or a "miss" indication, as described in Section [#sec-mau-semantics].
 
-If a table has no key property, then it contains no look-up table, just a default action, which is always executed (i.e., the associated lookup table is always the empty map).
+If a table has no ```key``` property, then it contains no look-up table, just a default action, which is always executed (i.e., the associated lookup table is always the empty map).
 
 Each key element can have an optional ```@name``` annotation which is used to synthesize the control-plane visible name for the key field.
 
 ####	The list of actions { #sec-table-action-list }
 
-A table must declare all possible actions that may appear within the associated lookup table or in the default action. This is done with the ```action```s attribute; the value of this attribute is always an ```actionList```:
+A table must declare all possible actions that may appear within the associated lookup table or in the default action. This is done with the ```actions``` property; the value of this property is always an ```actionList```:
 
 ~ Begin P4Grammar
 actionList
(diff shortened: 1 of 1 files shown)


jafingerhut:
The most significant change here is the new 2nd to last paragraph in the diff starting with "Additional properties are permitted".  Everything else is as described in the commit comments above.  I can back out all of the other changes if this is too off-topic.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -3484,11 +3484,13 @@ program, with decreasing priority levels.
...
 
-A ```table``` declaration indicates the interfaces expected from a ```table```: keys and actions. However, the best way to implement a table is actually dependent on the nature of the entries that will populate the table at runtime (for example, tables could be dense or sparse, could be implemented as hash-tables, associative memories, tries, etc.) An implementation attribute could also be used to pass additional information to the compiler back-end. The value of this attribute could be an instance of an ```extern``` block chosen from a suitable library of components. For example, the core functionality of the P4~14~ table ```action_profile``` constructs could be implemented on architectures that support this feature using a construct such as the following:
+Additional properties are permitted to change the semantics of a table lookup. For example, an architecture could define a property ```default_action_undefined``` that, when given the value ```true```, causes the behavior of the data plane to be undefined for any packet that found no match when searching the table.
```
mbudiu-vmw: We have to be careful with the semantics of table lookup, if it can be arbitrary, then the compiler cannot really reason about the behavior of programs. The compiler assumes the semantics in the diagram.

vgurevich:
Mihai,

I agree. The best way would be to periodically standardize more "popular"
properties. Additionally it might make sense to agree on the naming for
vendor-specific properties, so that people can clearly see them in the code.

Thanks,
Vladimir

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Andy's text is consistent with what was discussed at the language design meeting yesterday. Can you suggest an appropriate rewording that would let the compiler do its job?

jnfoster:
@vgurevich I don't see how to standardize vendor-specific properties and be open to extensions (as is the intention of this feature). What we can do is restrict which parts of the semantics of table lookup they are allowed to affect.

mbudiu-vmw:
I think that the semantics of the lookup should never change; the lookup is given a key and returns an action and a hit bit.
What can change is how this lookup is done, or how the actual table is implemented.

jnfoster:
I don't see how to implement @jafingerhut's `default_action_undefined` with those types. What action would be returned -- e.g., for a table with nothing in its action list?

mbudiu-vmw:
It is not clear the compiler should accept a table with no actions.
The compiler assumes that one of the actions will run, and nothing else.
The dataflow analysis for tables makes a join of the dataflow produced by all actions.
If that's not true all bets are off.

jnfoster:
You're right: the grammar does not allow the empty list of actions.

Still, what action should be produced if the table misses and there is no default action? If the `default_action_undefined`, would the onus be on the target to interpret a "0" hit bit and an arbitrary action as undefined (or target-specific) behavior? I suppose that could work, although it's kind of kludgy since the types are being abused ;-)

mbudiu-vmw:
Whatever it is, the compiler will assume it is one in the list. Otherwise the meaning of the program is undefined. Which is perhaps what the user wants.

jnfoster:
So maybe the right thing to say is that targets may support other properties that control how the results of table lookups (action and hit) are interpreted. For example, default_action_undefined...

vgurevich:
@jnfoster ,

The way to standardize is to adopt certain conventions on how to name these additional properties, e.g. by suggesting that vendors always name their properties as VENDOR_XXX_<property> or something. This can be published later in a separate "best practices" doc.  Separately, we can maintain a list of commonly recognized properties, such as "counters", "meters", etc.

jnfoster:
This kind of convention is fine, but it doesn't resolve the issue that @mbudiu-vmw raised: if the compiler relies on all tables to have a certain set of behaviors to do its job, then a vendor cannot  modify that behavior or the compiler will be broken. Hence, we must restrict what vendor extensions can do. (Exception: if the behavior with these extensions becomes completely undefined -- i.e., not just an undefined value, but C-style [undefined behavior](http://blog.regehr.org/archives/213) then we could have our cake an eat it too.) 

I think the solution proposed above resolves it -- i.e., vendor-specific properties can change whether/how an action and hit/miss metadata are interpreted on the packet state. Mihai would have to confirm that this is a strong enough guarantee for the compiler.

vgurevich:
Here is an example of a table with a slightly different behavior that started the whole conversation. A directly indexed table, which has 2^n entries and indexed by an n-bit-wide key. In many implementations, entries in such tables do not have a "valid" bit and therefore there can be no miss and thus no default action per se. From the control plane standpoint, there are two approaches that the driver designer can take:
1. Provide the same add/delete APIs as for regular table, but every time a delete happens, the action portion of the actual entry is overwritten by the default action. That's why there is no default/default -- the default action has to always be chosen.
2. Provide set() APIs instead, meaning that all the entries are always present in the table and the control plane can only change them, but there is no "delete"

jnfoster:
Neither of these approaches are inconsistent with the semantics we adopted: no default action means a miss in a table does not change the packet state (because false implies anything).

Issue #160, opened 2017-04-12
Is default_action required or not?


jafingerhut:
The latest spec seems inconsistent on this.  Early in section 12.2.1.3 it says:

    this property [default_action] must be present for all tables.

Later in that same section it says:

    If the user does not specify a value for the default action, the runtime behavior of the
    program is undefined until the control-plane initializes the default_action to one of
    the legal values. The compiler should produce a warning in this case.

If default_action must be present, then the later statement should be removed.  If the later statement is true, then the earlier phrase should be removed.

In case it matters, the 2017-Apr-08 compiler behavior for both p4test and p4c-bm2-ss is to warn about a table with no default_action, but no error.

vgurevich:
I think we need to do the following (as was discussed at the last meeting):

1. If default_action is not specified, let's make the default to be NoAction(). It will also be the default action in case the default_action attribute is  specified in a manner that is not fully resolvable at the compile-time. 
   -- The rationale: this is by far the most typical case and we do not want to burden people with the requirement to specify NoAction() in 99% of all the tables in their programs: it will serve no informational purpose, will make reading the programs more difficult and will also open the door for more bugs or compiler warnings. In addition, it is more compatible with P4-14. This was decided at the last meeting.

2. It might make sense to add another attribute, that would allow P4 programmers to specify that there is no such "default-default" action for a given table. Alternatively, the attribute can specify that there can be no miss. It can also be an annotation, but I would prefer an attribute.
   -- The rationale: it allows important optimizations on a wide variety of hardware. Essentially a traditional "indexed" table is typically the one to be expressed in such a way. This is an addition I'd like to propose.

mbudiu-vmw:
You should submit a pull request against the spec with these changes

jnfoster:
>
> I think we need to do the following (as was discussed at the last meeting):
>
>    1.
>
>    If default_action is not specified, let's make the default to be
>    NoAction(). It will also be the default action in case the default_action
>    attribute is specified in a manner that is not fully resolvable at the
>    compile-time.
>    -- The rationale: this is by far the most typical case and we do not
>    want to burden people with the requirement to specify NoAction() in 99% of
>    all the tables in their programs: it will serve no informational purpose,
>    will make reading the programs more difficult and will also open the door
>    for more bugs or compiler warnings. In addition, it is more compatible with
>    P4-14. This was decided at the last meeting.
>
>
Actually, I believe we agreed on was to take a declarative approach and
[...]

PR #195, opened 2017-05-10
Clarify semantics of default_action

jnfoster:
* Allow `default_action` to be omitted.
* Clarify that a table-miss on a table without a default action behaves like a no-op.
* Remove some trailing whitespace.

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -1427,17 +1427,17 @@ introduce a header stack called ```mpls``` containing 10 entries, each of type `
 
 ### Header Unions {#sec-header-unions}
 
-A header union represents an alternative between different headers. Header unions can be used to represent "options" in protocols like TCP and IP. They also provide hints to P4 compilers that only one alternative will be present, allowing it to conserve resources. 
+A header union represents an alternative between different headers. Header unions can be used to represent "options" in protocols like TCP and IP. They also provide hints to P4 compilers that only one alternative will be present, allowing it to conserve resources.
 
 A header union is defined as:
 ~ Begin P4Grammar
 headerUnionDeclaration
     : optAnnotations HEADER_UNION name
-      '{' structFieldList '}'         
+      '{' structFieldList '}'
     ;
 ~ End P4Grammar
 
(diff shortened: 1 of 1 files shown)


jnfoster:
Fixes #160

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -3552,7 +3557,7 @@ table t {
...
 
-The initial value for the default action is supplied as a value for the ```default_action``` property; this property must be present for all tables. The default action _must_ appear after the ```actions``` property and may be declared as ```const```, indicating that it cannot be changed dynamically by the control-plane. The default action _must_ be one of the actions that appear in the actions list. In particular, the expressions passed as ```in, out``` or ```inout``` parameters must be syntactically identical to the expressions used in one of the elements of the ```actions``` list.
+If present, the ```default``` action property _must_ appear after the ```action``` property. It may be declared as ```const```, indicating that it cannot be changed dynamically by the control-plane. The default action _must_ be one of the actions that appear in the actions list. In particular, the expressions passed as ```in, out``` or ```inout``` parameters must be syntactically identical to the expressions used in one of the elements of the ```actions``` list.
```
jafingerhut: typo?  default -> default_action for the name of the property

mbudiu-vmw:
Looks good, but the PR also has some stuff about unions in it.

jnfoster:
That's because I set my editor to delete white spaces as someone (you suggested). I added a comment in the pull request explaining.

I can go back and fix it up, but perhaps we can let this slide?

mbudiu-vmw:
Not a real problem.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Good catch. Fixed in 9333756

jnfoster:
Yeah, it took me forever to figure out what was going on. I carefully created two pull requests and kept getting these changes to the unions text. So I deleted them (you may have noticed), thinking I was really a Git-idiot, and maybe I had to start fresh. Then I realized my editor was now making these changes on save.

Sigh. At least all the trailing whitespace is gone so this won't be an issue going forward.

PR #196, opened 2017-05-10
Clarify limitations on archictecture-specific table properties

jnfoster:
This attempts to resolve the issue raised in #189.


mbudiu-vmw:
The paragraph about properties is OK, but any properties added should not change the data-plane semantics, they may change the implementation or the control-plane API.

The example with the missing default action is too specific. I would rather not say anything and relegate such comments to specific architecture back-ends documentation.

jnfoster:
I am happy walking back some of the text. From the email discussion last night, it @vgurevich's example doesn't actually require this feature, unless I misunderstood something.

jafingerhut:
Unless I am misunderstanding @vgurevich's example, I think these things are true:

Suppose a P4 target wishes to implement an exact match table with an N-bit search key and 2^N entries using a direct-index SRAM.

If the P4 program explicitly specifies a default_action for such a table, then the driver software must fill in _something_ into all 2^N SRAM entries at init time.  One possibility would be to write the encoding of the default action into every entry.  In addition, every time table_delete is called on that table, the driver software must write that entry with the same default action contents.

If the P4 program doesn't use the .hit attribute of the table, then the above is enough.

If the P4 program does use the .hit attribute of the table, the implementation needs a way to distinguish whether an entry read from the SRAM had table_add or table_delete called on it last.  At least in some cases that might require an additional bit in the table entry to distinguish between entries that were table_add'd with the default action, vs. entries that were table_delete'd or never table_add'd.  It could be a separate valid bit, or it could be one more encoding in a log(N)-bit encoding.

If the P4 program leaves out a default_action, and the proposed 'no side effects on a miss' semantics are implemented, one correct implementation is to behave as if 'default_action = NoAction' was explicitly specified.  All of the above still applies, including the wrinkle about implementing the .hit attribute (but most P4 programs use that for very very few tables, that I have seen).

jnfoster:
@mbudiu-vmw I pushed a new commit (1507d2e) to clarify. Can you take a look?

@jafingerhut I don't think I agree with this. P4 is a declarative language. We don't mandate the physical-level implementation strategy. Hence, 

> If the P4 program explicitly specifies a default_action for such a table, then the driver software must fill in something into all 2^N SRAM entries at init time.

This is one way to do things, but it seems like there are other choices. For example, it could add the extra bit to distinguish the add/delete (and therefore default) cases.

jafingerhut:
I agree one could add the extra bit to distinguish the add/delete cases, but then that bit must be initialized in all 2^N table entries, and falls under my statement "then the driver software must fill in _something_ into all 2^N SRAM entries at init time".

I am not trying to mandate any implementation here, or suggesting additions to the spec.  I am trying to determine why @vgurevich thinks there are benefits to having undefined behavior around this case.  I don't see it yet, but maybe by interacting with him I can learn.

jnfoster:
@jafingerhut You're right, that's a good point. I was substituting "some action" for "somehow."

Merging this now in the interest of converging, but the discussion of @vgurevich's example can continue.

Issue #150, opened 2017-03-30
Global instances


cole-barefoot:
The issue of global instances---i.e. top-level control or parser instantiation---came up at the LDWG meeting on Monday, March 27th, 2017.

## The Problem

What should the P4_16 spec say about global instantiation, if anything?  Here are some situations to consider.

1. If a control is instantiated at the top level but (a) not used in another control, and (b) not passed as an argument to a package, then is that instance dead code?  Can the control plane install rules to its tables?

1. If a control or parser is instantiated at the top level and used exactly once in another control or parser, is this equivalent to instantiating it locally in the namespace where it is used, and can the compiler inline it?

1. If a control or parser is instantiated at the top level, can it be used more than once?  Does this differ from how a control or parser instantiated locally (inside another control or parser) can be used?

jnfoster:
Do we need to revisit this issue at the 5/8 meeting? We've now resolved #129 differently.

jnfoster:
Closed by #197

PR #197, opened 2017-05-10
Add text restricting top-level instantiations

jnfoster:
First attempt to capture the restriction on top-level instances discussed in previous LDWG meeting and by email on 5/9.

Specifically, we restrict instantiations to a single top-level package, along with the parsers/controls mandated by the architecture. 

Added a short example to illustrate.

Note that declaration constants remain -- they are harmless.

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -2600,7 +2600,7 @@ control c(...) {
 }
 ~ End P4Example
 
-A P4 program may only instantiate the ```parser```s and ```control```s mandated by the architecture, as well as a single top-level package. This restriction is designed to ensure that any state either resides in the architecture itself, or is local to a ```parser``` or ```control```. For example, the following declarations would not be legal,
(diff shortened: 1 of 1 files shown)


jnfoster:
Fixes #150

jnfoster:
I think you're probably right but I'd like to look at the example. Can you give me a more detailed pointer?

Does the P4_14 converter emit top-level extern instances?

-N

mbudiu-vmw:
If I remember correctly the converter only generates local instances.

mbudiu-vmw:
However, the converter may need to generate top-level instances if one wants something like two tables with a shared action profile where one table is ingress and the second one in egress. I know this does not make sense, but it is legal code according to P4-14. In this case our converter is broken too.

jnfoster:
Okay, I can live with this. Intuitively, an extern object instance at the top-level is easier for the compiler back-end to handle: either it can be mapped to the available resources provided by the target or not. This is not necessarily the case with parsers and controls, which can have code inside (and complex control flow between them). 

I just pushed a new version that clarifies this and updates the example. Let me know what you think.

PR #201, opened 2017-05-11
Reformat Madoko sources to fit in 80 columns

jnfoster:
Although it touches almost every line, there are no material changes in this commit. 

Some lines are still longer than 80 columns when adding breaks would  cause a Madoko code block \`\`\` ... \`\`\` to appear at the start of a line. I also had to tweak the phrasing of a small number of paragraphs which began with a code block for the same reason.

PR #202, opened 2017-05-11
Fix section label clashes

### LINKED ISSUE #199: P4 Spec ToC: Sections 12.1 Actions and 12.2 Tables at wrong page number
Hey All,

I noticed in the table of contents that section 12.1 Actions is listed at page 93 (should be 75) and 12.2 Tables is listed at page 92 (should be 76).  I think the problem is that there are two sections listed for each of these.

Sections 12.1 and 16.3.1.3 are both "Actions"

Sections 12.2 and 16.3.1.1 are both "Tables"

I'm happy to fix this, but I wasn't sure what which we should change, or how we should change it.

jnfoster:
Do not merge until after #201.

Add labels to subsections of control-plane names that were causing
conflicts with `sec-tables` and `sec-actions` labels elsewhere. Also
added labels to other subsections (e.g. `sec-cp-keys`) to avoid the
possibility of the same error in the future.

Fixes #199

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -5850,7 +5850,7 @@ the global scope, have the same local and fully-qualified names.  The
 local names of controllable entities and enclosing constructs are
 derived from the syntax of a P4 program as follows.
 
-#### Tables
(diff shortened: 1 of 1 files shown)


jnfoster:
Rebased and now the diff is readable.

PR #206, opened 2017-05-12
Bad Madoko formatting, and eliminate one error in example P4 program

jafingerhut:


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -3448,8 +3446,8 @@ or set their validity bits directly.
 ~Begin P4Example
 U u;
 u.h1.setValid();     // u and u.h1 are valid
-H1 my_h1 = u.h1;         // my_h1 contains an undefined value
```
jnfoster: Good catch. I touched a lot of lines of text last night. I'm sure there are a few more.

PR #211, opened 2017-05-12
Fix a couple more instances of incorrect Madoko markup

jafingerhut:
I think this fixes all occurrences of this particular kind of incorrect Madoko markup, which was causing the { #cross-ref-label } to appear in the PDF output.


akeep:
+1

Noticed this earlier, and I was about to fix it.  Glad you beat me to it.

PR #214, opened 2017-05-12
Add header_union to list of derived types that create scopes

jafingerhut:
Also to the list of typeKeywords in p4.json, which controls how color
highlighting is done in P4 code samples.

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -1371,9 +1371,8 @@ do not require terminating semicolons).
 Various constructs act as namespaces that create local scopes for
 names:
 
-- Derived type declarations (```struct```, ```header```,
-  ```header_union```, ```enum```) introduce local scopes for the field
-  names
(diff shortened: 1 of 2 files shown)


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -1371,8 +1371,9 @@ do not require terminating semicolons).
 Various constructs act as namespaces that create local scopes for
 names:
 
-- Derived type declarations (```struct```, ```header```, ```enum```)
-  introduce local scopes for the field names
+- Derived type declarations (```struct```, ```header```,
+  ```header_union```, ```enum```) introduce local scopes for the field
```
cc10512: you want to keep this on a single line, even if it is longer than 80 chars. Otherwise madoko interprets header_union as a language and emits a warning. Sometimes also messes up formatting.

jafingerhut: OK, good catch.  I will update the PR.

PR #216, opened 2017-05-12
Several small fixes

mbudiu-vmw:


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2268,46 +2293,18 @@ integer, without a width specified.
...
-Rationale: ```int<W>``` is sufficient to allow a signed integer as a
-member, and it has easily-determined storage requirements, unlike an
-infinite precision ```int```.  ```match_kind``` values are not useful
-to store in a variable, as they are only used to specify how to match
-fields in table search keys, which are all declared at compile time. ```void```
-is not useful as part of another data structure.  Headers must have
-precisely defined formats as sequences of bits in order for them to be
-parsed or deparsed.
+Rationale:```int``` does not have precise storage requirements, unlike
+``bit<>``` or ```int<>``` types.```.
```
jafingerhut: extra ``` after types. ?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -1176,7 +1176,7 @@ produced by different programmers at different times:
 Similar to the C standard library, the P4 language specification
 defines a core library that declares useful P4 constructs. A
 description of the P4 core library is provided in Appendix
-[#sec-p4-core-lib]. Most P4 programs will include the core
+[#sec-p4-core-lib]. All P4 programs must include the core
```
jnfoster: I vaguely recall that we changed this from "All/must" to "Most/will." And `p4test` certainly accepts programs that do not include `core.p4`... 

(I don't feel strongly about it.)

mbudiu-vmw: Sometimes the compiler will synthesize code using verify even if you don't use it in your program. In this case if you haven't included core.p4 you will get very mysterious errors. So it should be "must".

jnfoster: Another solution (which I believe was discussed previously) would be to treat `verify` like a built-in function, more like "+" or "==", and not as part of a library.

mbudiu-vmw: There are other reasons to include core.p4, for example the standard errors are defined there. So I think it's safe to say "must". If you really know what you are doing you can do without it, but that's too hard to explain.

jnfoster: Sure, this is fine.

jnfoster: Does the `checkCorelibMethods` helper in the type checker validate *all* of the elements of `core.p4`? Or do we need to open an issue on `p4c`?

mbudiu-vmw: I am working on lookahead now.

jnfoster: I meant should we open a placeholder issue on p4c so we don't forget :-)

jnfoster:
@mbudiu-vmw I updated this pull request so it could be merged (including resolving two tiny conflicts). I hope that's okay. Can you confirm that I did so correctly?

mbudiu-vmw:
Looks fine to me.

jnfoster:
Okay. I just had to detangle eight more conflicts.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2556,7 +2589,8 @@ expression
     ;
 
 expressionList
-    : expression
+    : /* empty */
```
jnfoster: @mbudiu-vmw Just checking, we weren't relying on `expressionList` being non-empty elsewhere?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2175,6 +2178,28 @@ struct Parsed_headers {
 }
 ~ End P4Example
 
+### Tuple types { #sec-tuple-types }
```
jnfoster: When I detangled conflicts, I couldn't tell if this was meant to be deleted. So I kept it. Can you confirm @mbudiu-vmw

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2690,7 +2725,7 @@ The ```?:``` expression behaves as in C, e.g.:
 (x == 0) ? e0 : e1;
 ~ End P4Example
 
-The first argument is boolean, and the second and third arguments must
+The first argument is Boolean, and the second and third arguments must
```
jnfoster: @mbudiu-vmw Andy Keep had already lowercased all occurrences of the two-element type that shall not be named. I don't care which way we go, but let's not keep editing over each other.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5320,12 +5346,6 @@ the ```udpSupport``` parameter, as in the following example:
 GenericParser(false) TopParser;
 ~ End P4Example
 
-## Optional constructor parentheses
-
-For the sake of readability, type definitions with no constructor
```
jnfoster: Isn't this still true? Did this paragraph move elsewhere?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -6130,7 +6139,7 @@ table t {
 The ```@name``` annotation directs the compiler to use a different
 local name when generating the external APIs used to manipulate a
 language element from the control plane. It must have a string literal
-argument, which shall not contain the "```.```" character.  In the
+argument.  In the
```
jnfoster: Why did you remove this restriction? It is important.

jnfoster: @cole-barefoot can you also weigh in?

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: @cole-barefoot weigh in?

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: It is moved in the document before the table.

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: No, if we allow empty structs we have to allow empty lists too.

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: The dictionary says it's uppercase.

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: Yes, it is important, a name annotation can still be relative and contain a path. There are absolute paths `@globalname` and relative paths `@name`. But they are paths, not names. The compiler creates paths when inlining, for example.

jnfoster: I was not deeply involved with hashing this out, but I thought in the design we'd converged on `@name` was only supposed to modify the local component of the path.

mbudiu-vmw: When you inline a block into another you have to prepend the name of the callee to all names of the inner instances. If the callee has a relative name, the results are relative paths. I don't see how it can be done otherwise.

mbudiu-vmw: When you inline a block into another you have to prepend the name of the callee to all names of the inner instances. If the callee has a relative name, the results are relative paths. I don't see how it can be done otherwise.

cole-barefoot: I'm fine removing it---the important point is that `@name` can't introduce names that conflict outside its scope.

jnfoster: Okay.

jnfoster:
@mbudiu-vmw Let's get this in. Will you detangle and merge this or should I?

mbudiu-vmw:
I can only detangle this when you stop making changes.
Also, you have several comments to address from my reviews.

jnfoster:
You have the token! Go for it.

jnfoster:
Okay, I'll merge this! Thanks!

PR #221, opened 2017-05-12
Update core.p4 in appendix to include comments now in p4c version

jafingerhut:


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -6300,31 +6300,58 @@ predefined ```packet_in```\ and ```packet_out``` extern objects, used
...
+    /// @T can be a header type, a header stack, a header_union, or a struct
+    /// containing fields with such types.
     void emit<T>(in T hdr);
+    /// If @condition is true write the specified @data into the output packet,
+    /// advancing cursor.  @T can be a header type, a header stack, a header_union,
+    /// or a struct containing fields with such types.  If @condition is false,
+    /// do nothing.
     void emit<T>(in bool condition, in T data);
 }
+@name("NoAction")
```
jnfoster: Can this be removed? It's redundant.

jnfoster:
@jafingerhut you removed `verify`, per the comment in the `p4c` version?

jafingerhut:
@jnfoster - Yes, I did not copy 'verify' from the version of core.p4 currently in p4c, because of the comment in that version saying it should be removed.  If you want it kept in the P4-16 spec version, let me know.

cc10512:
Until we find a good home (ideally this should be a parser method or a compiler builtin primitive operation) I believe we need to keep it.

mbudiu-vmw:
We should remove it in the spec. The spec will be frozen today, we can't change it later.
We could also completely remove core.p4 from the spec, or indicate that it is a snapshot in time which may change.

jnfoster:
I agree with Mihai here. The spec is clear that `verify` is a built-in function. It doesn't mandate where it comes from. So it's safe to leave it out of `core.p4`.

jnfoster:
I'm going to merge this now. We can make further changes if necessary in new PRs.

PR #227, opened 2017-05-13
Third round of polishing

jnfoster:


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -1682,8 +1657,9 @@ baseType
 
 ### The void type
 
-The void type is written as ```void```. It contains no values. It can
-only appear in few restricted places in P4 programs.
+The void type is written as ```void```. It contains no values. It is
+not included in the set of base types as it can only appear in few
```
jafingerhut: Saying here that void is not included in the set of base types seems at odds with void being included in the list above.

jnfoster: Ahh, good catch. Just pushed a fix -- that sentence is referring to the production rule `baseType` not the set of base types.

PR #229, opened 2017-05-13
Add pseudocode for push_front and pop_front

jafingerhut:


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -3363,6 +3363,39 @@ manipulate the elements at the front and back of the stack:
...
+    for (int i = this.size-1; i >= 0; i -= 1) {
+        if (i >= count) {
+            this[i] = this[i-count];
+        } else {
+            this[i].setInvalid();
+        }
+    }
+    this.nextIndex = this.nextIndex + count;
+    if (this.nextIndex > this.size) this.nextIndex = this.size;
+    // Note: this.last, this.next, and this.lastIndex adjust with this.nextIndex
```
mbudiu-vmw: I don't think we have a lastIndex.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -3363,6 +3363,39 @@ manipulate the elements at the front and back of the stack:
   be a positive integer that is a compile-time known value. The return
   type is ```void```.
 
+The pseudocode below defines the behavior of ```push_front``` and ```pop_front```.
+
+~ Begin P4Pseudo
+void push_front(int count) {
```
mbudiu-vmw: Can you make count unsigned so we don't have to discuss this issue either?

jafingerhut: Sure.  I will propagate that change to the other places where push_front and pop_front have and 'int' argument for consistency.  Hopefully that is what you are expecting.

mbudiu-vmw:
In P4 it has to be int, because we don't have unsigned. But in pseudocode it must be positive.
You can at least add a comment that the compiler will reject negative values, and that the values have to be compile-time constants.

p4-16/spec/P4-16-draft-spec.mdk
jafingerhut: There is not a separate data value, but the spec mentions that lastIndex is nextIndex-1.  My comment is there simply to ensure that if some implementer decides to make the .next .last and .lastIndex as separate data items, they should be adjusted as and when nextIndex is updated.

jafingerhut:
Must the language spec restrict the argument of push_front and pop_front to compile time constants?  Can't that be left as an optional target-specific restriction?

mbudiu-vmw:
The problem is that if these are runtime values we have to say what their type is, it must be a bit<N> type for some N.

jafingerhut:
The spec already says that the argument to push_front and pop_front must be positive, and must be compile time constants, a few lines above these proposed additions.  I don't think it should be necessary to repeat it a few lines later.

p4-16/spec/P4-16-draft-spec.mdk
jafingerhut: OK, given your later comment, I am not planning on changing this.  Github still thinks there are changes requested, though.  Not sure how to change that.

PR #230, opened 2017-05-13
Add semantics of assignments between header unions, and isValid()

jafingerhut:


jafingerhut:
Not sure if this requires a wider audience for review, given that it could be considered more than just an editorial change.

mbudiu-vmw:
The semantics is the expected one: Copy all fields. So the explanation can be short.

jnfoster:
I agree with @mbudiu-vmw and I believe this was discussed. It just didn't
make it into the document.

To be safe, we could email p4-design but maybe that's overkill. What do you
think?

N

jafingerhut:
I suppose the description of header union assignment could be even shorter than what I wrote.  Given that `u = hi` has perhaps-surprising no-op behavior if `hi` is invalid and `u.hj` is valid for a different member header of `u`, it seemed wise to be explicit that this is not the case for assignments between header union variables.

jnfoster:
Looks great. Thanks Andy!

jnfoster:
@jafingerhut, I agree with @mbudiu-vmw. Want to add one more sentence at the beginning, then let's merge?

jafingerhut:
@jafingerhut @mbudiu-vmw I think it should be as you suggest now.  Take a look.

mbudiu-vmw:
Looks good.

PR #233, opened 2017-05-14
More Madoko markup fixes

jafingerhut:
I happened to notice that whenever Madoko has warnings like this early
in its output:

  warning: unable to read language definition: parsed_packet
  warning: unable to read language definition: table
  warning: unable to read language definition: control
  warning: unable to read language definition: parser

those occur because it found ```parsed_packet``` at the beginning of a
line somewhere in the input .mdk file.

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -783,8 +783,8 @@ preprocessor ```#include``` directives (see Section [#sec-preprocessor]).
 []{tex-cmd: "\indent"}
 The parser attempts to recognize an Ethernet header followed by an IPv4 header.
 If either of these headers are missing, parsing terminates with an
-error. Otherwise it extracts the information from these headers into a
-```Parsed_packet``` structure. The match-action pipeline is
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
We should file a bug against madoko.
Is there a difference in the output?

jnfoster:
Weird, I thought I'd eliminated these. Maybe more got introduced by further
edits.

I don't think it's a bug in Madoko. You shouldn't start a line with treble
backticks unless the next characters are the name of a language it knows
about. We should be sure that lines don't do this, unless we mean to have a
code display. And the warnings let us check that.

jafingerhut:
@mbudiu-vmw It does lead to a difference in the PDF output, at least.  It starts a new paragraph.

cc10512:
The main reason for treble quotes was the syntax highlighting for p4. I experimented with single quotes and it is still highlighting since we set p4 as the default language. That should relieve the restriction with the position in the line.

jnfoster:
@cc10512 great find! Let's update the document to avoid treble backticks later. It will be much less fragile and easier to maintain.

PR #234, opened 2017-05-15
Fourth round of polishing

jnfoster:
Partway through expressions. Mostly trivial changes just streamlining and making the prose internally consistent.

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -2505,10 +2505,10 @@ created type.
 
 # Expressions { #sec-exprs }
 
-This section describes all computations that can be performed in P4,
-grouped by the type of the values than can be processed.
+This section describes all expressions that can be used in P4,
+grouped by the type of value they produce.
 
-The grammar for general expressions is given by:
+The grammar production rule for general expressions is as follows:
 
 ~ Begin P4Grammar
 expression
@@ -2590,43 +2590,41 @@ typeArgumentList
 
 ~ End P4Grammar
 
-See also the complete P4 grammar in Appendix [#sec-grammar].
-
-An additional semantic check is required for the right shift to check
-that there is no space between the two consecutive greater-than signs ```>>```.
-This rule is required to allow parsing for both the right
-shift operators and specialized types, such as in ```function<bit<32>>```.
+See Appendix [#sec-grammar] for the complete P4 grammar.
 
 This grammar does not indicate the precedence of the various
 operators. The precedence follows exactly the C precedence
 rules. Concatenation (```++```) has the same precedence as infix
 addition. Bit-slicing ```a[m:l]``` has the same precedence as array
 indexing (```a[i]```).
+An additional semantic check is required for right shift to check
+that there is no space between the two consecutive greater-than signs ```>>```.
+This rule is required to allow parsing for both the right
+shift operators and specialized types, such as in ```function<bit<32>>```.
 
-In addition to these expressions, ```select``` expressions (described
-in Section [#sec-select]) may be used only in parsers.
+In addition to these expressions, P4 also supports ```select``` expressions (described
+in Section [#sec-select]), which may be used only in parsers.
 
 ## Expression evaluation order { #sec-expr-eval-order }
 
-Given a complex expression, the order in which sub-expressions are
-evaluated can be important if these sub-expressions can produce
+Given a compound expression, the order in which sub-expressions are
+evaluated is important when the sub-expressions have
 side-effects. P4 expressions are evaluated as follows:
 
-- Boolean operators ```&&``` and ```||``` are evaluated short-circuit:
+- Boolean operators ```&&``` and ```||``` use short-circuit evaluation---i.e.,
   the second operand is only evaluated if necessary.
-- The conditional operator ```?:``` evaluates its first argument, and
-  based on its values it evaluates the second or the third.
+- The conditional operator ```e1 ? e2 : e3``` evaluates ```e1`, and then
+  either evaluates `e2` or `e3`.
 - All other expressions are evaluated left-to-right as they appear in
   the source program.
-- Function and method calls are evaluated as described in Section
+- Method and function calls are evaluated as described in Section
   [#sec-calling-convention].
 
-## Expressions on error values { #sec-error-exprs }
+## Operations on expressions of type ```error``` { #sec-error-exprs }
 
 Symbolic names declared by an ```error``` declaration belong to the ```error```
-namespace.  The ```error``` type only supports comparisons
-for equality and difference.  The result of a comparison is a boolean
-value.
+namespace.  The ```error``` type only supports equality (`==`) and inequality (`!=`) comparisons.
+The result of such a comparison is a boolean value.
 
 For example, the following operation tests for the occurrence of an
 error:
@@ -2636,9 +2634,9 @@ error errorFromParser;
 if (errorFromParser != error.NoError) { ... }
 ~ End P4Example
 
-## Expressions on enum values { #sec-enum-exprs }
+## Operations on expressions of type ```enum``` { #sec-enum-exprs }
 
-Symbolic names declared by an ```enum``` do not belong to the
+Symbolic names declared by an ```enum``` declaration do not belong to the
 top-level namespace, but to a newly introduced namespace.
 
 ~ Begin P4Example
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2590,43 +2590,41 @@ typeArgumentList
...
 
-## Expressions on error values { #sec-error-exprs }
+## Operations on expressions of type ```error``` { #sec-error-exprs }
 
 Symbolic names declared by an ```error``` declaration belong to the ```error```
-namespace.  The ```error``` type only supports comparisons
-for equality and difference.  The result of a comparison is a boolean
-value.
+namespace.  The ```error``` type only supports equality (`==`) and inequality (`!=`) comparisons.
+The result of such a comparison is a boolean value.
```
cc10512: assignment?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2636,9 +2634,9 @@ error errorFromParser;
 if (errorFromParser != error.NoError) { ... }
 ~ End P4Example
 
-## Expressions on enum values { #sec-enum-exprs }
+## Operations on expressions of type ```enum``` { #sec-enum-exprs }
 
-Symbolic names declared by an ```enum``` do not belong to the
+Symbolic names declared by an ```enum``` declaration do not belong to the
 top-level namespace, but to a newly introduced namespace.
```
cc10512: Symbolic names declared by an ```enum``` belong to the namespace introduced by the enum declaration rather than the top-level namespace.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
-An ```enum``` value can only be compared for equality/difference using ```== and !=```. ```enum```
-values cannot be cast to or from any other types.
+Similar to errors, ```enum``` expressions only support equality (`==`)
+and inequality (`!=`) comparisons. Expressions whose type is an `enum`
+cannot be cast to or from any other type.
 
-When ```enum``` values appear in the control-plane API the compiler
-back-end has to choose a suitable serialization data type and
+Note that if an ```enum``` value appears in the control-plane API, the
+compiler must select a suitable serialization data type and
```
cc10512: Then what is the reason for limiting casting?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
 
-The ```?:``` expression behaves as in C, e.g.:
+### Conditional operator
 
-~ Begin P4Example
-(x == 0) ? e0 : e1;
-~ End P4Example
+A conditional expression of the form ```e1 ? e2 : e3``` behaves the
+same as in languages like C. The expression `e1` is evaluated first,
+and either `e2` or `e3` is evaluated depending on the result.
```
cc10512: Repeat of the text in line 2616

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
+- Negation, denoted by ```!```, and
+- Equality and inequality tests, denoted by ```==``` and ```!=``` respectively.
 
-Operator precedence is similar to C. Operator evaluation is
-short-circuit.
+The precedence of these operators is similar to C and uses short-circuit evaluation.
 
-There are no implicit casts from bit-strings to booleans or
-vice-versa. As a consequence, a C program fragment such as:
+P4 does not implicit cast from bit-strings to booleans and
```
jafingerhut: Improve grammar here.  Maybe "P4 does not implicitly cast"?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
+vice versa. As a consequence, a program is valid in a language like C such as,
```
jafingerhut: replace "is valid" with "valid" or "that is valid" ?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
+- Test for inequality between bit-strings of the same width,
   designated by ```!=```. The result is a boolean value.
 - Unsigned comparisons ```<,>,<=,>=```. Both operands must have the
   same width; the result is a boolean value.
 
-All the following operations produce bit-string results when applied
-to bit-strings. All these operations require both operands to have the
+Each of the following operations produces a bit-string result when applied
+to bit-strings of the same width:
 same width.
```
jafingerhut: second occurrence of 'same width' should be deleted

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
+  from ```l``` (which becomes the least significant bit of the result) to ```m``` (the
+  most significant bit of the result) from the source operand. The conditions ```0 <= l < W```
   and ```l <= m < W``` are checked statically (where ```W``` is
   the length of the source bit-string). Note that both endpoints of
   the extraction are inclusive. The bounds are required to be
   compile-time known values so that the result width can be computed
-  at compile time.
-- Slices are l-values: one can assign to a slice: ``` e[m:l] = x ```
-  This statement sets bits ```m``` to ```l``` of ```e``` to the
+  at compile time. Slices are also l-values, which means taht P4 supports asssigning to a slice: ``` e[m:l] = x ```.
```
jafingerhut: taht -> that

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
+  The effect of this statement is to sets bits ```m``` to ```l``` of ```e``` to the
```
jafingerhut: sets -> set

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
+for some ```W````. Recall that the ```int<W>``` denotes signed ```W```-bit integers,
 represented using 2's complement.
 
-"Underflow" or "overflow" produced by arithmetic cannot be detected:
-operations "wrap around", similar to C operations on unsigned values
-(i.e., representing a large value on W bits will only keep the
-least-significant W bits of the value)
+In general, P4 arithmetic operations do not detect "underflow" or "overflow":
+operations simply "wrap around", similar to C operations on unsigned values.
+Hence, attempting to representing large values using `W` bits will only keep
```
jafingerhut: representing -> represent

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: I'm pretty sure comparison...

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Nice. Added.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Seems not fundamental. @mbudiu-vmw ?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
 
 All binary operations (except shifts) require both operands to have
-the same exact type (signedness) and width; supplying operands with
-different widths or signedness produces an error at compile time. No
+the same exact type (signedness) and width and supplying operands with
+different widths or signedness produces a compile-time error. No
 implicit casts are inserted by the compiler to equalize the
-types. There are no binary operations that combine signed and unsigned
-values (except shifts).
+types. With the exception of shifts, P4 does not have any no binary operations that combine signed and unsigned
```
jafingerhut: "any no" -> "any"

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Added "As described above, ..."

In general, this section is repetitive but I think that's not terrible. We're sacrificing readability for the end-to-end reader, but making it easier to flip open to a random subsection and get a full picture of how particular features behave.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Fixed, thanks.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Ugh. Thanks.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2647,196 +2645,191 @@ X.v1  // reference to v1
...
 - Shifting with a negative amount does not have a clear semantics:
-  while in C the result is undefined, in P4 the type system makes it
+   the P4 type system makes it
   illegal to shift with a negative amount.
-- In C, shifting with an amount larger or equal to the number of bits
-  has an undefined result (unlike the P4 definition).
-- Finally, shifting may require doing work which is exponential in the
-  number of bits of the right-hand-side operand.
+- Unlike C, shifting by an amount larger or equal to the number of bits
+  has a well-defined result (unlike C).
```
jafingerhut: delete one of the two occurrences of 'unlike C' in this sentence

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Ugh. Thanks.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Ugh. Can you tell I was being hurried home for dinner? Sigh.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Thanks.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Thanks.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Thanks.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Thanks. In general, I think we should avoid comparing P4 to C/C++/Java so much. There are sections in the spec where it seems to be every other paragraph and I don't think it helps promote understanding. I've been removing these as I go when it seems sensible.

p4-16/spec/P4-16-draft-spec.mdk
jafingerhut: If you did not join the mother of your children for Mother's day, you would be in much much much more trouble than if a little awkward English gets into the P4-16 1.0 spec!  Think long term, man :-)

jnfoster: Ha. I had the kids from wake-up until mid-afternoon so she could go ride up Montebello with her friends (and sacrificed "my" exercise window to do a bit of polishing).

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: It is fundamental, enums are completely abstract.
How would you propose casting enums and to what?

jnfoster: Let's leave this as-is.

PR #241, opened 2017-05-15
Fifth round of polishing

jnfoster:


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2841,134 +2841,126 @@ bit<16> y;
...
-  types ```T``` and ```S``` can be cast back and forth if they
-  represent base types.
 
-### Implicit casts { sec-implicit-casts }
-Unlike C, P4 allows a very limited number of implicit casts. The
-reason is that often the implicit casts have a non-trivial semantics,
-which is invisible for the programmer.
+The following casts are legal in P4:
+
+- ```bit<1> <-> bool```: converts the value ```0``` is converted to ```false```, the value ```1``` is converted ```true```, and vice versa.
```
jafingerhut: "is converted" -> "converted"

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -3982,26 +3980,31 @@ parserStates
...
+Hence, used in the context of a `parserDeclaration` the production
+rule `parserTypeDeclaration` should not yield type parameters.
+
 At least one state, named ```start```, must be present in any ```parser```.
-A parser cannot contain two states with the same name. A parser cannot
-define the ```accept``` and ```reject``` states; these are logically
-outside of the parser.
+A parser may not define two states with the same name.
+It is also illegal for a parser to give explicit definitions for
+the ```accept``` and ```reject``` states explicitly---those
```
jafingerhut: Might want to delete one of "explicit" and "explicitly" in the same sentence

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -4038,34 +4039,36 @@ ParserModel {
...
     : optAnnotations STATE name
      '{' parserStatements transitionStatement '}'
     ;
 ~ End P4Grammar
 
+Each state has a name and a body. The body consists of a sequence of
+statements that describe the processing performed when the parser
+transitions to that state including:
+
+- Mocal variable declarations,
```
jafingerhut: Mocal -> Local

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -4038,34 +4039,36 @@ ParserModel {
...
+- Assignment statements, and
+- Method calls, which can serve several purposes:
+  * Invoking functions (e.g., using ```verify``` to check the validity of data already parsed), and
+  * Invoking methods (e.g., extracting data out of packets or computing checksums) and other parsers (see Section [ #sec-invoke-subparser ]).
+
+The syntax for parser statements is given by the following grammar rules:
```
jafingerhut: You left out transition statements from the list, which should be kept, I think?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -4084,26 +4087,21 @@ parserBlockStatement
...
-The state body contains a sequence of:
-
-- local variable declarations
-- assignment statements
-- method calls. These can serve multiple purposes:
-  * ```verify``` calls
-  * method invocations (e.g., for extracting data out of packets),
-  * invocations of other parsers.
+Architectures may place restrictions on the expressions and statements
+that can used in a parser---e.g., they may forbid the use of
```
jafingerhut: "can used" -> "can be used"

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Thanks. Done.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Done.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Done.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Done.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: The rationale was that `transition`s are syntactically distinct from parser statements and are described in another section.

I'll restore and add a parenthetical with a pointer to the next section.

Issue #161, opened 2017-04-14
Types supported by emit


mihaibudiu:
The spec mentions that emit can handle headers, arrays or structs, but does not say what happens if a bit<X> field is passed as argument.

PR #243, opened 2017-05-15
Sixth and final round of polishing

jnfoster:


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -4876,22 +4877,18 @@ match_kind {
 ~ End P4Example
 
 These identifiers correspond to the P4~14~ match kinds with the same
-names. The semantics of these annotations is actually irrelevant for
-describing the behavior of the P4 abstract machine; how they are used
-influences only the control-plane API and the actual implementation of
-the look-up table. From the point of view of the P4 program, a look-up
-table is an abstract finite map that is given a key and produces as a
-result either an action or a "miss" indication, as described in
-Section [#sec-mau-semantics].
```
jnfoster: @mbudiu-vmw I see what this paragraph is meant to convey, but worry it might cause confusion. Given the same control plane operations, two tables with different match_kinds on their keys may behave quite differently -- e.g., ternary vs. lpm. So it's not just the implementation of the look-up table, but it's semantics too.

Since this doesn't seem like a critical point from the perspective of defining the language, maybe just cut it?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -4876,22 +4877,18 @@ match_kind {
...
+names. From the point of view of the P4 program, a look-up table is an
+abstract finite map that is given a key and produces as a result
+either an action or a "miss" indication, as described in Section
+[#sec-mau-semantics].
 
 If a table has no ```key``` property, then it contains no look-up
-table, just a default action, which is always executed (i.e., the
```
jnfoster: This is no longer true so I deleted it.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -4928,20 +4927,18 @@ table smac {
...
-names; e.g., the following program fragment is illegal:
+Each action in the list of actions for a table must have a distinct
+name---e.g., the following program fragment is illegal:
 
 ~ Begin P4Example
 action a() {}
 control c() {
     action a() {}
     // Illegal table: two actions with the same name
-    table t { actions = { a; .a; } }
```
jnfoster: I see what this is doing, but the "." is subtle. I wonder if it's better just to leave it off...

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5444,41 +5458,30 @@ packet_out {
...
-        if (!cond) return;
-        this.data.append(data);
+    // append the data to the packet
+    void raw_emit<T>(T data) {
+        really this.data.append(data);
         this.lengthInBits += data.lengthInBits;
     }
-}
+    // append data to the packet if valid
+    void emit<T>(T data) {
```
jnfoster: This definition needs to be carefully vetted.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5711,11 +5713,11 @@ As an example, consider the following program fragment:
 
 ~ Begin P4Example
 // architecture declaration
-parser p(...);
-control c(...);
-control d(...);
+parser P(...);
```
jnfoster: Updating for consistency with our metavariable conventions.

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: The whole point of the spec is to clarify such subtle points.

jnfoster: I don't feel strongly. We can put it back.

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: I think that this is still true. There is nothing else that could be in this table except the default_action. The compiler synthesizes such tables from straight-line code on architectures which only support tables.

jnfoster: I think the following table is legal,
```
table t = {
  key = { }
  actions = { }
}
```
and behaves like a no-op when applied.

To clarify:
* We no longer require the `default_action.` We borrowed text from P4_14 to clarify that it behaves like a no-op if the table misses and there is no `default_action.`

* The discussion of table semantics says that looking up returns either:
  * Miss, or
  * Hit and an action.

(Note that the miss case does not have an action.)

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: The tables all implement the same data-plane interface ITable<TKey, TAction>, where TKey is a tuple. The data-plane semantics of a table is given solely by this interface, period. The interface, you will notice, does not depend on the match_kind. The implementation does.

Maybe the control-plane semantics depends on the match_kind, but the data-plane semantics does not.

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: The text is about a table without a key, not a table with an empty key.
The table can still have lots of actions in the actions list.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: I'm confused. 

Consider these two tables:
```
table t1 {
  key = { ipv4.dstAddr : ternary }
  actions = { fwd; drop; }
  const entries = { 
     prefix : fwd;
  } 
}
```
and 
table t1 {
  key = { ipv4.dstAddr : exact }
  actions = { fwd; drop; }
  const entries = { 
     prefix : fwd;
  } 
}
```
And suppose that prefix1 encodes `10/8` and prefix2 encodes `192.168/16`.

Applying `t1` will hit on all addresses  in `10/8`, while applying `t2` will hit only for the single address `10.0.0.0`.

Isn't that a difference in data-plane semantics?

Comment (mbudiu-vmw) [inline p4-16/spec/P4-16-draft-spec.mdk]: The semantics is that given a key you get back an action. How that is done is irrelevant.

I can use two tables exactly like t1, if I populate them with different entries I will get different behaviors. That's what you are doing in your example. But both tables take a key and return an action - that is the dataplane semantics.

File: p4-16/spec/P4-16-draft-spec.mdk (hunk 2 above)
Comment (jafingerhut) [inline p4-16/spec/P4-16-draft-spec.mdk]: Can't a table have many actions in the actions list regardless of whether it has a key defined, and regardless of whether that list of fields in the key is empty or not?

File: p4-16/spec/P4-16-draft-spec.mdk [hunk 6]
```diff
@@ -4671,20 +4670,30 @@ controlBody
...
-used to build parameterized control blocks, see Section
-[#sec-parametrization].
+Hence, used in the context of a `controlDeclaration` the production
+rule `controlTypeDeclaration` should not yield type parameters.
 
-We start by describing the core components of a ```control``` block,
+P4 does not support exceptional control-flow within a ```control```
+block. That is, there is no equivalent of the ```verify``` statement
+or the ```reject``` state from parsers. Hence, all error handling must
+be performed explicitly by the programmer.
```
jafingerhut: Perhaps it does not qualify as an exception to the statements above, but 'exit' can be invoked from 'deep within a call stack' of controls and actions, and cause the control block to exit immediately.  It must be explicitly used by the programmer for it to occur, so it is not hidden from them, hence my reason for saying that it is perhaps not needed to be mentioned here.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5011,17 +5007,19 @@ entry matches a given packet, then the table does not affect the
...
-entries that enable the execution of the algorithm in the P4 source,
-allows the compiler to infer how the table is actually used and
-potentially make better placement decisions for tagets with constraint
+While table entries are typically installed by the contol plane,
+tables may also be initialized at compile-time with a set of entries.
+This is useful in situations where tables are
+used to implement fixed algorithms---defining table entries statically
+enables expressing these algorithm directly in P4,
+which allows the compiler to infer how the table is actually used and
+potentially make better allocation decisions for tagets with limited
```
jafingerhut: "tagets" -> "targets"

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5112,30 +5108,30 @@ control ingress(inout Header_t h, inout Meta_t m, inout standard_metadata_t stan
...
-the semantics of table lookups, which always produce either a ```hit```
-and an action or a ```miss```---they can only change how those results
-are interpreted on the state of the data plane. This restriction is
-needed to ensure that it is possible to reason about the behavior of
-tables during compilation.
+#### Additional properties {#sec-additional-table-properties}
+
+A ```table``` declaration defines its essential control and data plane
+interfaces---i.e., keys and actions. However, the best way to
+implement a table may actually dependend on the nature of the entries
```
jafingerhut: dependend -> depend ?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5156,20 +5152,21 @@ table t {
...
 ~ End P4Example
 
-(An action profile specifies the fact that, although the table is
-expected to have a large number of entries, only a small number of
-distinct entry values are expected. This can lead to an optimized
-implementation of the table, using an indirection level to share
-identical entries. The ```ActionProfile``` ```extern``` object in the
-previous example is meant to convey this information.)
+Here the action profile might be used to optimize for the case where
+the table has a large number of entries, but the actions assocated
```
jafingerhut: "assocated" -> "associated"

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5444,41 +5458,30 @@ packet_out {
...
-        this.emit(header.valid$, header);
-    }
-
-    // append the data to the packet if the condition is true
-    void emit<T>(bool cond, T data) {
-        if (!cond) return;
-        this.data.append(data);
+    // append the data to the packet
+    void raw_emit<T>(T data) {
+        really this.data.append(data);
```
jafingerhut: What does "really this" mean?

p4-16/spec/P4-16-draft-spec.mdk
jafingerhut: It looks good to me.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Sure. 

However, the table need not have a default action, so I still think the original text was wrong. For example:

```
table t = {
  actions = { a1; a2; ... }
}
```

jafingerhut: Perhaps the original statement could be correct by simply removing the word "default"?  i.e. this statement sounds true to me:
"If a table has no ```key``` property, then it contains no look-up table, just an action, which is always executed."

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Earlier we defined the data plane semantics as a function from vectors of bits to vectors of bits. And I think that's the plain English reading of the term.

(Minor nit: as I wrote elsewhere, since `default_action` is no longer mandatory, I don't think it's the case case that given a key you get back an action. You either get a Miss, or a Hit and an action.)

I am worried that some readers will interpret this text as follows: if I take a snapshot of the state populated by the control plane (or use constant entries, as in this example), then the data plane will behave the same (i.e., have the same semantics) no matter what kind of `match_kind` I use with my table. And that's would plainly be false. 

Does this sentence serve an important definitional purpose? If so, then I could be talked into keeping it. If not, I think it's probably better to remove it.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Given my `t` above, couldn't the control plane change the action by modifying the `default_action`? Or is that precisely the case we are trying to rule out?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -2424,7 +2424,7 @@ representing the received packet that is processed.
 parserTypeDeclaration
     : optAnnotations PARSER name optTypeParameters
      '(' parameterList ')'
-    ;
+    ;    ;
```
mbudiu-vmw: Why an extra semicolon here?

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -4628,25 +4628,24 @@ the standard error ```error.ParserTimeout```.
...
-
-A ```control``` block may start with declarations of ```action```s, ```table```s,
-constants, variables, and instantiations.
+Syntactically, a ```control``` block is declared with a name,
+parameters, optional type parameters, and a sequence of declarations
+of constants, variables, ```action```s, ```table```s, and other
+instantiations:
 
 ~ Begin P4Grammar
+controlTypeDeclaration
```
mbudiu-vmw: controlTypeDeclaration has its own section, it does not need to be duplicated here.

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: Exit is indeed rather exceptional. It was introduced in the language after this paragraph was written.
Maybe this paragraph should be reorganized.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Restored.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Added "We also use standard `for` notation to
iterate through the elements of a stack `(e : data)` and list of
fields for header stacsk and unions `(f : data.fields$)`."

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: Yes, the data-plane does not care about the match_kind; it does not appear in the definition of the interface I described above.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Added "The only statement which
has a non-local effect on control flow is `exit`, which causes
execution of the enclosing control block to immediately terminate."

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Fixed.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Fixed.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Fixed.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Fixed.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -4928,20 +4927,18 @@ table smac {
...
+Each action in the list of actions for a table must have a distinct
+name---e.g., the following program fragment is illegal:
 
 ~ Begin P4Example
 action a() {}
 control c() {
     action a() {}
     // Illegal table: two actions with the same name
-    table t { actions = { a; .a; } }
+    table t { actions = { a; a; } }
```
mbudiu-vmw: The new example does not have the same semantics as the old one.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5282,6 +5279,11 @@ use of constructor parameters.
 Consider again the parser declaration syntax:
 
 ~ Begin P4Grammar
+parserTypeDeclaration
```
mbudiu-vmw: There is a separate section for parser type declarations.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5420,21 +5422,33 @@ the stack in order of increasing indexes.
...
+unions:
+
+- When applied to a header, `emit` appends the data in the header to
+  the packet if it is valid and otherwise behaves like a no-op.
+- When applied to a header stack, `emit` recursively invokes itself to
+  each element of the stack.
+- When applied to a `struct` or header union, `emit` recursively
+  invokes itself to each field.
+
+Appending a value such as a bit-string or integer value to a
```
mbudiu-vmw: We have agreed in the design meeting not to allow arbitrary bit-strings.
Headers may have additional constraints, such as being byte-aligned.

p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -5444,41 +5458,30 @@ packet_out {
...
+        if (isHeader(T))
+            if(data.valid$)
+                raw_emit(data);
+        else if (isHeaderStack(T))
+            for (e : data)
+                 emit(e);
+        else if (isStruct(T) || isHeaderUnion(T))
+            for (f : data.fields$)
+                 emit(e.f)
+        else
```
mbudiu-vmw: else error

p4-16/spec/P4-16-draft-spec.mdk
mbudiu-vmw: Yes, the control-plane can change the default_action if it is not const.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Fixed.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Okay. I thought it might be nice to reproduce it here, but I'll delete (and same for parserTypeDeclaration).

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Removed.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: There was another discussion that suggested otherwise, but you're right (#161). I'll fix this now.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: +1

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: What about structs: they need not be byte-aligned but they are allowed?

mbudiu-vmw: In the end all leaves are headers, to it does not matter.

jnfoster: Correct, under the assumptions about T. I've updated the prose and pseudo-code to match.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Restored to

> These identifiers correspond to the P4~14~ match kinds with the same
> names. The semantics of these annotations is actually not needed to
> describe the behavior of the P4 abstract machine; how they are used
> influences only the control-plane API and the implementation of
> the look-up table. From the point of view of the P4 program, a look-up
> table is an abstract finite map that is given a key and produces as a
> result either an action or a "miss" indication, as described in
> Section [#sec-mau-semantics].

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Added a comment that a compiler may add a default_action and that we assume this transformation has been done. 

Also restored the original text here.

p4-16/spec/P4-16-draft-spec.mdk
jnfoster: Restored the old one.

jnfoster:
@mbudiu-vmw passing this over to you for review again.

See d454490 for how I handled the discussion of `default_action`?

Thanks.

jnfoster:
Good point. Fixed.

jnfoster:
@jafingerhut do you have further comments or should I merge?

jafingerhut:
@jnfoster these changes are good for me

PR #244, opened 2017-05-15
reduce formality

jnfoster:
I find the mention of ISO/IEC 2382 somewhat odd. It sets up expectations for a highly pedantic document, which this specification is not.

How do people feel about deleting it and replacing it with a more
direct statement that captures what we are trying to convey?


vgurevich:
As the person who inserted that long time ago with the hopes of writing more formal stuff :) , but ultimately never had a chance to, I do not mind.

jnfoster:
Thanks. Glad you are not offended :-)

PR #245, opened 2017-05-15
Make claims about complexity truthier

### LINKED ISSUE #217: Complexity
This claim is wrong.

> P4 is not a Turing-complete language. 

Given a sufficiently powerful extern (e.g., arbitrary precision integers) it's trivial to encode a Turing-complete formalism into P4 parsers.

This claim is also potentially misleading:

> In fact, assuming a fixed cost for a table lookup operations and interactions with extern objects,
> all P4 programs (i.e., parsers and controls) execute a constant number of operations for each byte > of an input packet received and analyzed.

It requires reading "P4 program" as a single parser or control, and not an instantiation of an architecture. Note that in the presence of recirculation, one can write programs with arbitrary complexity, or even non-terminating programs.

jnfoster:
Fixes #217

### DIFF
--- a/p4-16/spec/P4-16-draft-spec.mdk
+++ b/p4-16/spec/P4-16-draft-spec.mdk
@@ -263,7 +263,7 @@ cards, FPGAs, software switches, and hardware ASICs. As such, the
 language is restricted to constructs that can be efficiently
 implemented on all of these platforms.
 
-Assuming a fixed cost for a table lookup operations and interactions
+Assuming a fixed cost for table lookup operations and interactions
 with extern objects, all P4 programs (i.e., parsers and controls)
 execute a constant number of operations for each byte of an input
 packet received and analyzed. Although parsers may contain loops,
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-draft-spec.mdk
```diff
@@ -263,11 +263,13 @@ cards, FPGAs, software switches, and hardware ASICs. As such, the
 language is restricted to constructs that can be efficiently
 implemented on all of these platforms.
 
-P4 is not a Turing-complete language. In fact, assuming a fixed cost
-for a table lookup operations and interactions with extern objects,
-all P4 programs (i.e., parsers and controls) execute a constant number
-of operations for each byte of an input packet received and
-analyzed. In other words, the computational complexity of a P4 program
+Assuming a fixed cost for a table lookup operations and interactions
```
jafingerhut: "for a table" -> "for table"

jnfoster: Fixed.

PR #251, opened 2017-05-15
v1.0.0 release candidate

jnfoster:


mbudiu-vmw:
I don't see an approve button. I approve anyway.

jafingerhut:
Do other languages with years for releases like C++11, C++14, C++17, etc. do versioning of language specs within one of those 'major versions'?  If so, seems like we would be in good company calling this P4-16 1.0.0

jnfoster:
This is a valid point, and I'm personally quite sympathetic to it. However, there's also a concern that version numbers like v1.2 might cause a different kind of confusion. I had a quick chat with the P4 board and there's a preference to identify language revisions with years, similar to what C++ does.

So P4_16 v1.0.0 it is.

PR #252, opened 2017-05-15
Move error to type

jnfoster:


jnfoster:
Let's do a picky pass to improve the formatting later. For now, I just
wanted to get one thing in the TOC colored correctly :-)

mbudiu-vmw:
Since you are making this change, why not make tuple and match_kind types too?
Without this change the highlighting will be wrong.

jnfoster:
Done.

PR #253, opened 2017-05-15
Additional nits from fifth round of polishing.

jnfoster:


jnfoster:
Okay @mbudiu-vmw here are your orphaned round-five review requests.

PR #259, opened 2017-05-16
convert to simple quotes

cc10512:


cc10512:
Fully mechanic. Since all our code samples are using ~ Begin/~ End, the only triple quotes in the actual text were keywords. While I didn't check all of them, I did check the ones followed by a new line.

jnfoster:
We can more easily check the rendered PDF for errors.

PR #268, opened 2017-05-19
Missing '\n'

jnfoster:


jnfoster:
Sorry, I misunderstood what you said in our email discussion. This makes sense since we don't escape anything but quotes.

PR #275, opened 2017-05-22
v 1.0.4

jnfoster:
Draft proposal for v1.0.4 maintenance release.

* Copy all files from v1.0.3 to v1.0.4. 

* Change parallel to sequential semantics.

* Add field.valid as a key for matching in tables.

* Update descriptions for primitive operations to clarify which ones produce undefined values.

Because of the copy step above, it will be easiest to review this by examining the individual commits, excluding the first.

Note that this pull request was originally #126. That pull request got mangled due to making p4-spec public. There is some discussion there that may be of interest.


p4-14/v1.0.4/tex/p4.pt
```diff
@@ -646,11 +646,11 @@ valid bit for a metadata field, defining a separate metadata flag to
 represent that field's validity is a reasonable work around.
 \end{adjustwidth}
 
-Only valid packet header fields may result in a match (when a value is
-specified for exact or ternary matches against the field), although a
-match operation may explicitly check if a header instance (or field)
-is valid. Only valid packet headers are considered for deparsing (see
-Section~\ref{sec:deparsing}).  
+When an invalid packet header is matched in a table (i.e., when a value is
+specified for exact or ternary matches against the field), its value is
```
jafingerhut: I don't see how the phrase 'when a value is specified for exact or ternary matches against the field' is clarifying anything here.  In fact, it seems to be muddying the waters by leaving out 'lpm' match_kind's.  Adding 'lpm' wouldn't really clarify anything, because then what about 'range' and user-defined match_kinds?  It seems leaving out the parenthesized phrase entirely is better.

p4-14/v1.0.4/tex/p4.pt
```diff
@@ -1977,6 +1976,10 @@ The hash value is used to generate a value between \texttt{base} and
 The result can be used as a hash digest of a long key (such as 5-tuple) or as 
 an offset/index to reference a register array. Normal value conversion 
 takes place when setting \texttt{dest} to the result.
+
+If any of the fields in the \texttt{field_list_calc} have a parent
+header that is not valid, then the operation leaves \texttt{dest} with
+an undefined value.
```
jafingerhut: I will need to re-read a comment from antoninbas, but this change perhaps needs correction.  I believe he said that if any fields have a parent header that is not valid, bmv2 leaves those fields out of the hash calculation.  Thus the set of fields included in the hash calculation can change from one packet to the next based upon which set of headers are valid for those packets.

p4-14/v1.0.4/tex/p4.pt
jnfoster: I agree. I've removed the parenthetical and rephrased.

p4-14/v1.0.4/tex/p4.pt
jnfoster: I'll wait for your opinion.

jafingerhut: Here is a quote on this topic from @antoninbas from this PR: https://github.com/p4lang/p4-spec/pull/174

"bmv2 skips invalid fields / headers in calculations, as this has proved very convenient in P4 programs which hash on L4 fields
this is also consistent with deparsing. IMO a field list calculation is just: 1) serializing the fields to a buffer (just like deparsing), and 2) applying a calculation over the buffer."

If people are OK with making this the official defined behavior of the 'modify_field_with_hash_based_offset' primitive, then I propose changing the phrase above:

    "then the operation leaves \texttt{dest} with an undeifned value"

to something like this:

    "then those header fields are left out of the field list for the purposes of calculating the hash value."

jnfoster: Speculatively making the change. @antoninbas can you confirm that you agree with it?

antoninbas: yes I stand with what I wrote and I agree with @jafingerhut's phrasing

jafingerhut:
My only remaining comment is that we should update the date from "March XX" when we are ready to pull the trigger on releasing.  Thanks!

p4-14/v1.0.4/tex/p4.pt
```diff
@@ -36,16 +36,16 @@
...
 
-%\title{\sffamily\bfseries\huge The P4 Language Specification\\ \vspace{3mm} \Large Version 1.0.3}
+%\title{\sffamily\bfseries\huge The P4 Language Specification\\ \vspace{3mm} \Large Version 1.0.4}
 \centerline{\sffamily\bfseries\huge The P4 Language Specification}
 \vspace{3mm}
-\centerline{\sffamily\Large Version 1.0.3}
+\centerline{\sffamily\Large Version 1.0.4}
 \vspace{3mm}
-\centerline{\sffamily\large November 2, 2016}
+\centerline{\sffamily\large March XX, 2017}
```
cc10512: March XX should we replace by an actual date. I would say use today's date.

p4-14/v1.0.4/tex/p4.pt
```diff
@@ -646,11 +646,11 @@ valid bit for a metadata field, defining a separate metadata flag to
 represent that field's validity is a reasonable work around.
 \end{adjustwidth}
 
-Only valid packet header fields may result in a match (when a value is
-specified for exact or ternary matches against the field), although a
-match operation may explicitly check if a header instance (or field)
-is valid. Only valid packet headers are considered for deparsing (see
-Section~\ref{sec:deparsing}).  
+When an invalid packet header is used as a part of the match key for a table,
+the value of the header is undefined.
```
cc10512: value of the key?

p4-14/v1.0.4/tex/p4.pt
```diff
@@ -3297,13 +3306,27 @@ Control flow begins with the control function \texttt{ingress}.
 1.0.1 & 2015-01-28 & Added action profiles and action selectors. Added attribute \texttt{support_timeout} to tables. \\  \hline
 1.0.2 & 2015-03-03 & Added \texttt{push} and \texttt{pop} primitive actions. \\  \hline
 1.0.3 & 2016-11-3 & See \ref{sec:change-summary-1.0.3} for details. \\ \hline
+1.0.4 & 2017-MM-DD & See \ref{sec:change-summary-1.0.4} for details. \\ \hline
```
cc10512: Fix date.

p4-14/v1.0.4/tex/p4.pt
jafingerhut: @cc10512 It seems reasonable to me to restrict the undefined-ness to only those fields of the search key that are parts of invalid headers.  Why force the undefined-ness to spread to the whole key?  For example, if the undefined-ness is restricted to one field, and if the match_kind was ternary or lpm for that field, and every current table entry installed was "don't care" for all bits of that field, the undefined-ness doesn't make the forwarding behavior become undefined.

cc10512: @jafingerhut thank you for the explanation. I would not look forward to debugging such a program, but I can see how this would be (ab)used.

jnfoster:
I've fixed the dates, added a changelog, and built the PDF. 

@cc10512 please review and then we'll merge and post?

PR #277, opened 2017-05-24
Allow empty actions list

mbudiu-vmw:
Because we don't require a default_action, and the compiler will (logically) insert a NoAction call in this case, it makes sense to allow tables to have an empty actions list.


mbudiu-vmw:
Can someone review this so we can merge it?

Issue #276, opened 2017-05-24
Meaning of 2nd arg in 2-argument extract?


jafingerhut:
The P4-16 spec says that the second argument is in units of bits, and it is the length of the varbit field of the header, not the complete header.

That is consistent with this sample extract call for IPv4 options in a p4c test program: https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/match.p4#L64-L65

If both of the above are correct, then one of the sample programs in the spec uses a wrong 2nd argument in an extract call, the one showing how to parse TCP options.  The call to extract in this code fragment:
```
    state parse_tcp_option_sack {
         bit<32> n = (bit<32>)b.lookahead<Tcp_option_sack_top>().length;
         b.extract(vec.next.sack, n);
         transition start;
    }
```
should be something equivalent to this, where the -16 part is to subtract out the length of the fixed length fields of the Tcp_option_sack_h header:
```
        b.extract(vec.next.sack, 8*n-16);
```
and the corresponding change should be made in this p4c test program: https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/spec-ex19.p4#L55-L56

This is based upon my understanding that the 8-bit length field in the SACK option header is equal to the total length of the SACK option header in bytes, including the 2 bytes of type and length field, plus all bytes of variable length part after that.

I am happy to make PRs for the spec and for spec-ex19.p4, but wanted to verify that it was those that should be corrected, and not the first 2 things above.

mbudiu-vmw:
It looks like you are right.

PR #279, opened 2017-05-25
Correct examples of extracting headers with varbit field

jafingerhut:


jafingerhut:
Proposed fix for issue https://github.com/p4lang/p4-spec/issues/276

jafingerhut:
Note: I have looked at quite a few packets recorded from TCP sessions that included TCP options, and there is no requirement that there be an "End" TCP option after the last non-End TCP option.  It is possible for the last non-End option to go all the way up to the end of the TCP header, with the TCP payload data beginning on the very next byte, if there is any, or the packet ends if there is no data in the TCP payload (e.g. the packet is sent primarily as an ACK packet, with no new data to send).

Thus this example TCP option parsing code is likely to either (a) get a PacketTooShort parse error, or (b) misinterpret TCP payload data as additional TCP options.  I realize that sample code need not be correct in all ways, and it can still serve its purpose, but this example seems like a reasonable motivation for an extension to P4-16 that I discussed with @jnfoster verbally once.

That extension would be that one could invoke a sub-parser, but give it an offset into the packet which that sub-parser should 'pretend that the packet ends there'.  For TCP option parsing, for example, the top level parser would use the calculated TCP header length to determine the location in the packet (or the end of the packet) where TCP options must end there, or earlier, and the TCP option sub-parser should never go past that point.  I can open a separate enhancement issue for that idea.

mbudiu-vmw:
You can just keep a counter of remaining bytes which you decrement explicitly.

jafingerhut:
@mbudiu-vmw And do what when it gets to 0?  I don't think P4-16 allows if statements in parsers.

mbudiu-vmw:
You can match in select on zero or something with a set sign bit

jafingerhut:
@mbudiu-vmw OK, I can see it as possible with select.  Convenient, no, but barely possible, yes.

jafingerhut:
Attached is a P4-16 parser adapted from the one in testdata/p4_16_samples/spec-ex19.p4 in the p4lang/p4c repository, but enhanced to try to detect when the end of the TCP header has been reached, and detect several error cases, e.g. a multi-byte option that would run past the end of the TCP header.  I haven't tested it yet.  Definitely possible, if it is correct, and definitely not convenient to write.
[tcp-option-sub-parser.p4.txt](https://github.com/p4lang/p4-spec/files/1028228/tcp-option-sub-parser.p4.txt)

vgurevich:
One other comment: Since we removed the explicit prohibition on extracting
the same header more than once, this code might need to have explicit
checks for that or, alternatively you can use header stacks and take
advantage of a built-in error.

But, if anything, it clearly illustrates why TLVs are not good for
high-speed HW processing...

jafingerhut:
@vgurevich This code does use header stack, yes?

vgurevich:
Whoops! I missed line 66. My bad!

jafingerhut:
I know there are a lot of comments above, but I believe the small proposed changes in this pull request are good changes to make to the spec to clarify the units of the second argument of the 2-argument extract() call.  Does anyone see anything else that should be addressed before accepting this PR?

jnfoster:
Looks good to me.

Issue #286, opened 2017-06-09
Question: P4-16 headerUnionDeclaration: definition / appearance


SmartHoneybee:
The definition of `derivedTypeDeclaration` in `p4-16/spec/P4-16-spec.mdk` mentions `headerUnionDeclaration` twice, the definition in `p4-16/spec/grammar.mdk` only once. Why?

In `p4-16/spec/P4-16-spec.mdk`  `headerUnionDeclaration` is defined as:
```
headerUnionDeclaration
    : optAnnotations HEADER_UNION name { structure.declareType(*$3); }
      '{' structFieldList '}'          { $$ = new IR::Type_Union(@3, *$3, $1, *$6); }
    ;
```
but in `p4-16/spec/grammar.mdk` as:
```
headerUnionDeclaration
    : optAnnotations HEADER_UNION name
      '{' structFieldList '}'
    ;
```
I'm guessing that the latter is the "correct definition". Is that correct?

Thanks.

jnfoster:
Indeed. That looks like a typo (the actual Bison code got copied into the Markdown).

SmartHoneybee:
Thank you for responding.

I found some other potential typos while studying the grammar and would appreciate feedback if my interpretation is right.

* `typedefDeclaration` in section 7.3. lists a `name` in `TYPEDEF derivedTypeDeclaration name ';'`, but the P4 grammar (in appendix E) does not (it is just `TYPEDEF derivedTypeDeclaration ';'` there).
 (*Interpretation*: `name` is missing in the P4 grammar.)
* `typeRef` in section 7.2. lists `tupleType` as option, but `typeRef` in the P4 grammar does not.
(*Interpretation*: `tupleType` is missing in the P4 grammar.)
* `instantiation` in the P4 grammar uses a `:` instead of a `|` to separate the options
* `specializedType` in the P4 grammar references `prefixedTYpe` instead of `prefixedType`

mbudiu-vmw:
We should fix these bugs. I will submit a PR to correct them.
Thank you.

jafingerhut:
Thanks, @mbudiu-vmw for the fixes.  There is one remaining occurrence of "derivedTypeDeclarationame" in file grammark.mdk that should be replaced with "derivedTypeDeclaration".

SmartHoneybee:
Awesome, thank you for the timely responses.

When is the next P4<sub>16</sub> specification release scheduled? (I.e. when will those changes become part of a published specification version?)

[...]

PR #292, opened 2017-06-19
fixes inconsistency for optAnnotations in table entries

### LINKED ISSUE #289: Possible inconsistency with annotation position
Annotations on match key elements come after the key element, while in all other cases, it seems that the annotations come before the object they annotate (e.g. struct fields).
Is there a reason for this?

cc10512:
Fixes #289 and addresses one of the requests in #286


cc10512:
this was the solution with bison -- moving the annotation to the end. I haven't tried to move it back to front.

Issue #332, opened 2017-06-27
reserved keyword used as variable name in Appendix C


tomrodeheffer:
In P4_16 Language Specification 2017-05-22 Appendix C there are are several P4 code fragments that use the reserved keyword `header` as if it could be a variable name:

```P4
ck16.clear();           // prepare checksum unit
ck16.update(header.ipv4); // write header
verify(ck16.get() == 16w0, error.IPv4ChecksumError); // check for 0 checksum
```

```P4
header.ipv4.hdrChecksum = 16w0;
ck16.clear();
ck16.update(header.ipv4);
header.ipv4.hdrChecksum = ck16.get();
```

```P4
ck16.clear();
ck16.update(header.ipv4.hdrChecksum);  // original checksum
ck16.remove( { header.ipv4.ttl, header.ipv4.proto } );
header.ipv4.ttl = header.ipv4.ttl - 1;
ck16.update( { header.ipv4.ttl, header.ipv4.proto } );
header.ipv4.hdrChecksum = ck16.get();
```

This is completely bogus.  Just about any other identifier would do in these fragments, for example `h`, or `p` as is used in Section 5.

jnfoster:
Indeed. This is not legal P4_16. I've made a pull request (#333).

PR #333, opened 2017-06-27
Replace `header` with `h` in Checksum example in Appendix

jnfoster:
Note that `header` is a reserved keyword!


jnfoster:
Fixes #332.

PR #345, opened 2017-07-03
fixes for instantiation, non-primitive implementations, random hashes.

cc10512:


p4-16/psa/PSA.mdk
```diff
@@ -457,7 +457,7 @@ enum HashAlgorithm {
   crc32_custom,
   crc16,
   crc16_custom,
-  random,           /// are random hash algorithms useful?
+  target_default,    /// target implementation defined
```
vgurevich: target_default is fine.

I would move identity to be always the first. 

The meaning of *_custom is not quite clear...

p4-16/psa/PSA.mdk
```diff
@@ -172,18 +172,21 @@ backend is responsible for mapping the PRE extern APIs to the
...
 
-Note: some of these operations may not be implemented as primitive
-operations on a certain target. However, All of the operations can be
-implemented as a combination of other operations. Applications that
-rely on non-primitive operations may incur significant performance
-penalties, however, they should be functionally correct.
+**Note to implementers**: some of these operations may not be
+implemented as primitive operations on a certain target. However, the
+goal is that all operations should be implementable as a combination
+of other operations. Target documentation should highlight such
```
jafingerhut: "may not be implemented as primitive operations" - What is a primitive operation?  Such things were defined in the P4_14 spec, but the word "primitive" does not occur in the P4_16 v1.0.0 spec, not even once.

"should be implementable as a combination of other operations."  Of what other operations?  Of other operations that could have been written as P4_16 assignment statements?  Of other PRE extern method calls?

The second one doesn't make sense, because the implementation must "bottom out" somewhere.

If the intended meaning is the first, "P4_16 assignment statements", what is the importance of stating this?  The implementer could implement it this way if their chosen implementation is such that P4_16 assignment statements can achieve the desired internal state modification, but it is really up to them.  And the cost of some P4_16 assignment statements can be way more expensive than others, in a target-dependent way, so not sure you are really giving users of the PSA any kind of reassurance with a statement like this.

cc10512: @jafingerhut the architecture is the time to talk about hardware mapping. These are not the operations in the p4 program. Primitive operations and other operations are referring to what the hardware target supports natively or not. I will add clarifications to that effect. 

I think we should give users a hint that performance is target dependent -- it will be better to be warned than be surprised by performance variability. If you have any suggestions on how to express this in a more marketable way, by all means!

p4-16/psa/PSA.mdk
```diff
@@ -171,18 +171,21 @@ backend is responsible for mapping the PRE extern APIs to the
...
 
-Note: some of these operations may not be implemented as primitive
-operations on a certain target. However, All of the operations can be
-implemented as a combination of other operations. Applications that
-rely on non-primitive operations may incur significant performance
-penalties, however, they should be functionally correct.
+**Note to implementers**: some of these operations may not be
+implemented as native operations on a certain target. However, the
+goal is that all operations should be implementable as a combination
+of native and non-native operations. Target documentation should highlight such
```
jafingerhut: Can you give an example of a native operation?  Of a non-native operation?  I don't necessarily mean 'in the spec' yet, until I know what you mean by these terms, although any reader of the PSA spec might ask the same questions.

jafingerhut: My main comment here is: If we left out these two sentences

"some of these operations may not be implemented as native operations on a certain target. However, the goal is that all operations should be implementable as a combination of native and non-native operations."

then would you be content with the result?  I don't see what an implementer can get out of those sentences that means anything to them, i.e. that they can use to guide their actions when designing a system.

PR #346, opened 2017-07-07
Suggested changes to PSA Hash extern

jafingerhut:


p4-16/psa/PSA.mdk
```diff
@@ -451,6 +451,8 @@ enum HashAlgorithm {
   crc32_custom,
   crc16,
   crc16_custom,
+  ones_complement16,  /// One's complement 16-bit sum used for IPv4 headers,
+                      /// TCP, and UDP.
   random,           /// are random hash algorithms useful?
   identity
```
cc10512: can you also please move identity to be the first item on the list of algorithms (per @vgurevich's request)?

p4-16/psa/PSA.mdk
```diff
@@ -462,26 +464,41 @@ Example usage:
...
+  /// @param data The data over which to calculate the hash.
+  /// @param max The hash value is divided by max to get modulo.
+  ///        An implementation may limit the largest value supported,
+  ///        e.g. to a value like 32, or 256.
+  /// @return (base + (h % max)) where h is the hash value.
   O getHash<T, D, M>(in T base, in D data, in M max);
 }
 ```
 
+TBD: Should there be a `const` defined that specifies the maximum
```
cc10512: I think that maximum should be a target specific value. Of course there is a max, 2^sizeof(O)-1 ...
But, do we need separate types for base and max (i.e., T and M)? They are likely to be the same size ...

Issue #315, opened 2017-06-25
[PSA] consistent naming convention


hanw:
It would be really nice if the authors could standardize on the naming conventions. Why there types, named PortId_t, entry_key (without _t) and psa_parser_input_metadata_t? if everything is PSA-specific, then let's use psa_ everywhere. If we want to use "_t" for types, let's be consistent. Also, it would be nice if we can establish some conventions with regards to case CamelCase, underscores, etc.

PR #348, opened 2017-07-12
New PSA.mdk Counter section

jafingerhut:
There isn't any fleshing out of the counter control plane API with
these changes.

There is significant fleshing out of the Counter and DirectCounter
extern definitions from the P4 program perspective, including a
reference implementation defined as pseudocode, and an example P4
program fragment demonstrating the use of both Counter and
DirectCounter.

Perhaps controversial changes:

Eliminate CounterType_t.  If you want a packet counter, just
instantiate a counter and only use the count() method that defaults to
adding 1 to the counter.  If you want a byte counter, use the count()
method that takes an 'add_value' parameter, and pass whatever you want
as the number of bytes in the packet.  If you want both, instantiate 2
counter externs, and update them both.

Enable more than one DirectCounter to be associated with a single
table.

Specify that the counter update operation should be saturating, not
wrap-around, including the rationale for why.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -697,7 +697,7 @@ precise?
 
 TBD: The rules above seem to imply that an action that calls `count`
 on a `DirectCounter` instance may only be an action of that instance's
-.ne owner table.  Normally an action definition can be shared across
+one owner table.  Normally an action definition can be shared across
 multiple tables (e.g. NoAction, but also the definition of more
 complex actions can be shared across tables, even ones inside of
 different control blocks if the action is defined at the top level).
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -529,14 +529,16 @@ only update them.  If you wish to implement a feature involving
 sequence numbers, for example, use Registers instead (Section
 [#sec-registers]).
 
-Direct counters are counters associated with a particular P4 table.
+Direct counters are counters associated with a particular P4 table,
+and are implemented by the extern `DirectCounter`.  There are also
+'indirect' counters, which are implemented by the extern `Counter`.
 The primary differences between direct counters and indirect counters
 are:
 
 - Number of independently updatable counter values:
   - A single instantiation of a direct counter always contains as many
-    independent counter values as the table with which it is
-    associated.
+    independent counter values as the number of entries in the table
+    with which it is associated.
   - You must specify the number of independent counter values for an
     indirect counter when instantiating it.  This number of counters
     need not be the same as the size of any P4 table.
@@ -556,10 +558,10 @@ that adds 1 to the counter value.
 
 One may implement byte counters by using the `count` method that adds
 a caller-specified `add_value` to the counter, containing a value that
-the P4 program wishes to use for the length of the packet.
+you wish to use for the length of the packet.
 
-We mention these as common uses cases for counters.  You may of course
-use them to count things other than packets and bytes.
+We mention these as common use cases for counters.  You may of course
+use them to count things other than packets or bytes.
 
 
 ### Counter
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -578,8 +578,8 @@ use them to count things other than packets or bytes.
 /// bits possible in type S for the index.  Type S must contain enough
 /// bits to specify all values in the range [0, n_counters - 1].
 
-extern Counter<W, V, N, S> {
-  Counter(N n_counters);
+extern Counter<W, V, S> {
+  Counter(bit<32> n_counters);
   void count(in S index, in V add_value);
   void count(in S index);
 
@@ -744,7 +744,7 @@ control ingress(inout headers hdr,
                 in  psa_ingress_input_metadata_t  istd,
                 out psa_ingress_output_metadata_t ostd)
 {
-    Counter<ByteCounter_t, PacketLength_t, PortId_t, PortId_t>(MAX_PORTS)
+    Counter<ByteCounter_t, PacketLength_t, PortId_t>((bit<32>) MAX_PORTS)
         port_bytes_in;
     DirectCounter<PacketCounter_t, bit<1>>() per_prefix_pkt_count;
     DirectCounter<ByteCounter_t, PacketLength_t>() per_prefix_byte_count;
@@ -788,7 +788,7 @@ control egress(inout headers hdr
                // ...
                )
 {
-    Counter<ByteCounter_t, PacketLength_t, PortId_t, PortId_t>(MAX_PORTS)
+    Counter<ByteCounter_t, PacketLength_t, PortId_t>((bit<32>) MAX_PORTS)
         port_bytes_out;
     apply {
         // Note that hdr.ipv4.totalLen for a packet may have been
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -523,34 +523,69 @@ extern Checksum<W> {
 
 ## Counters
 
-Counters are a simple mechanism for keeping statistics about the
-packets that trigger a table in a Match Action unit.
+Counters are a mechanism for keeping statistics.  The control plane
+can read counter values.  A P4 program cannot read counter values,
+only update them.  If you wish to implement a feature involving
+sequence numbers, for example, use Registers instead (Section
+[#sec-registers]).
 
-Direct counters fire when the count method is invoked in an action, and
-have an instance for each entry in the table.
+Direct counters are counters associated with a particular P4 table.
+The primary differences between direct counters and indirect counters
+are:
 
+- Number of independently updatable counter values:
+  - A single instantiation of a direct counter always contains as many
+    independent counter values as the table with which it is
+    associated.
+  - You must specify the number of independent counter values for an
+    indirect counter when instantiating it.  This number of counters
+    need not be the same as the size of any P4 table.
+- Where counter updates are allowed in the P4 program:
+  - For a direct counter, you may only invoke its `count` method from
+    inside the actions of the table with which it is associated, and
+    this always updates the counter value associated with the matching
+    table entry.
+  - For an indirect counter, you may invoke its `count` method
+    anywhere in the P4 program where extern object method invocations
+    are permitted (e.g. inside actions, or directly inside a control's
+    `apply` block), and every such invocation must specify the index
+    of the counter value to be updated.
 
-### Counter types
+One may implement packet counters by always using the `count` method
+that adds 1 to the counter value.
 
-```
-enum CounterType_t {
-    packets,
-    bytes,
-    packets_and_bytes
-}
-```
+One may implement byte counters by using the `count` method that adds
+a caller-specified `add_value` to the counter, containing a value that
+the P4 program wishes to use for the length of the packet.
+
+We mention these as common uses cases for counters.  You may of course
+use them to count things other than packets and bytes.
+
 
 ### Counter
 
 ```
-extern Counter<W, S> {
-  Counter(S n_counters, W size_in_bits, CounterType_t counter_type);
-  void count(in S index, in W increment);
+/// Indirect counter with n_counters independent counter values, where
+/// every counter value has a size specified by type W.  Each counter
+/// can be updated either by adding 1 via the `count(index)` method,
+/// or by adding a value add_value with size specified by type V.
+/// Type W must contain at least as many bits as type V.
+///
+/// Type N is separate from type S to handle the case when the number
+/// of counter values is a power of 2, but you wish to use the fewest
+/// bits possible in type S for the index.  Type S must contain enough
+/// bits to specify all values in the range [0, n_counters - 1].
+
+extern Counter<W, V, N, S> {
+  Counter(N n_counters);
+  void count(in S index, in V add_value);
+  void count(in S index);
 
   /*
   @ControlPlaneAPI
   {
     W    read<W>      (in S index);
+    W    read_and_clear<W> (in S index);
     W    sync_read<W> (in S index);
     void set          (in S index, in W seed);
     void reset        (in S index);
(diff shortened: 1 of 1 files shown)


p4-16/psa/PSA.mdk
```diff
@@ -582,6 +673,138 @@ extern DirectCounter<W> {
...
+An action of an owner table need not have `count` method calls for all
+of the `DirectCounter` instances that the table owns.
+
+The reference implementation for the `DirectCounter` extern is
+essentially the same as that for `Counter`, including the saturating
+behavior.  Since there is no `index` parameter to the `count` method,
+there is no need to check for whether it is in range.
+
+TBD: What to say about `entry_key` type?  Does it need to be made more
+precise?
```
cc10512: I believe this is the issue that @jnfoster brought up yesterday: the entry_key must match the table to which the counter is associated, and therefore we can't have a single entry key defined for the whole PSA. He has proposed wording yesterday, something to this effect: PSA should specify that the counters/meters will use the P4 runtime mode of addressing

p4-16/psa/PSA.mdk
```diff
@@ -582,6 +673,138 @@ extern DirectCounter<W> {
...
+
+TBD: The rules above seem to imply that an action that calls `count`
+on a `DirectCounter` instance may only be an action of that instance's
+.ne owner table.  Normally an action definition can be shared across
```
cc10512: typo: .ne

p4-16/psa/PSA.mdk
```diff
@@ -523,34 +523,69 @@ extern Checksum<W> {
...
+/// Indirect counter with n_counters independent counter values, where
+/// every counter value has a size specified by type W.  Each counter
+/// can be updated either by adding 1 via the `count(index)` method,
+/// or by adding a value add_value with size specified by type V.
+/// Type W must contain at least as many bits as type V.
+///
+/// Type N is separate from type S to handle the case when the number
+/// of counter values is a power of 2, but you wish to use the fewest
+/// bits possible in type S for the index.  Type S must contain enough
+/// bits to specify all values in the range [0, n_counters - 1].
```
jafingerhut: Please suggest a different way to specify it in legal P4_16.  I'm definitely open to something better.

p4-16/psa/PSA.mdk
```diff
@@ -561,11 +596,67 @@ extern Counter<W, S> {
...
+        // See below for optional debug information to record.
+    }
+}
+
+void count(in S index) {
+    count(index, 1);
+}
+```
+
+Rationale for saturating behavior: With saturating behavior, the
```
jafingerhut: I realize that is potentially controversial, as mentioned in the commit message, and that is why this is a PR open to comments, not something already committed and published.  There is a proposed rationale in the text explaining why I personally think it is better, but if you have arguments why you think people would prefer some other behavior, please explain.

p4-16/psa/PSA.mdk
jafingerhut: I could fairly easily get rid of the type parameter N by making the type of n_counters bit<32> instead, which with the current p4c then requires casting the n_counters parameter value to bit<32> when you call the constructor.  Not sure if you consider that an improvement, but I've committed such a change to this PR for you to examine.  If you have more ideas for removing more of them, please give specific suggestions, preferably ones that compile with today's compiler, or a version of it planned to be committed in the next couple of months.

p4-16/psa/PSA.mdk
```diff
@@ -523,34 +523,71 @@ extern Checksum<W> {
...
+[#sec-registers]).
+
+Direct counters are counters associated with a particular P4 table,
+and are implemented by the extern `DirectCounter`.  There are also
+'indirect' counters, which are implemented by the extern `Counter`.
+The primary differences between direct counters and indirect counters
+are:
+
+- Number of independently updatable counter values:
+  - A single instantiation of a direct counter always contains as many
```
vgurevich: As mentioned in the comment to another commit, the semantics of DirectCounter used together with ActionProfile needs to be clarified. Here are a couple of options:
1. DirectCounter cannot be used on tables that use `implementation` property
2. Specifically for tables with ActionProfiles, the number of direct counters is equal to the number of entries in the action profile
3. Regardless of the implementation, the number of counters is equal to the number of match entries

The last option is the cleanest on the surface, but since counting is an action, it might be difficult to implement on all targets

p4-16/psa/PSA.mdk
```diff
@@ -523,34 +523,71 @@ extern Checksum<W> {
...
+    independent counter values as the number of entries in the table
+    with which it is associated.
+  - You must specify the number of independent counter values for an
+    indirect counter when instantiating it.  This number of counters
+    need not be the same as the size of any P4 table.
+- Where counter updates are allowed in the P4 program:
+  - For a direct counter, you may only invoke its `count` method from
+    inside the actions of the table with which it is associated, and
+    this always updates the counter value associated with the matching
+    table entry.
```
vgurevich: Same comment. In general, there is no "back link" in P4 from action back to match

p4-16/psa/PSA.mdk
```diff
@@ -523,34 +523,71 @@ extern Checksum<W> {
...
+  - For an indirect counter, you may invoke its `count` method
+    anywhere in the P4 program where extern object method invocations
+    are permitted (e.g. inside actions, or directly inside a control's
+    `apply` block), and every such invocation must specify the index
+    of the counter value to be updated.
```
vgurevich: In most real-world hardware implementations, one can't increment the same counter instance (and more often than not even the same counter, regardless of the instance) the more than once.

p4-16/psa/PSA.mdk
jafingerhut: I don't understand this comment.

+ Are you saying that for a DirectCounter, you want to allow a `count` update method call from somewhere outside of an action for the table that the DirectCounter is associated with?

+ Are you saying that the `count` method call should sometimes update a counter value associated with a table entry different than the one that was matched?

If you are not saying either of those things, then how do you want the statement you are commenting to change?

p4-16/psa/PSA.mdk
```diff
@@ -561,11 +598,67 @@ extern Counter<W, S> {
...
+}
+
+void count(in S index) {
+    count(index, 1);
+}
+```
+
+Rationale for saturating behavior: With saturating behavior, the
+control plane can know that information was lost, if it reads the
+maximum possible value from a counter.  In this case, it knows that
```
vgurevich: Two thoughts:
1. The number of bits in the HW usually should have little to do with how it is exposed to the control plane. Typically control plane exposes all counters as 64-bit values regardless of the underlying width
2. Given that the data plane has no access to the counter value anyway, it should not matter much from P4 perspective, but certainly still matters on the control plane side
3. Given the first point, wrap-around counters are preferred, since there is no information loss (provided that the control plane polls them relatively often or uses counter ejection mechanisms) and no requirement to zero them out. But, again, I'd rather not mandate the implementation details, but a simple control plane requiremen af anything.

p4-16/psa/PSA.mdk
jafingerhut: Understood.  That is a platform-specific limitation, one that we can add a note about in the PSA spec if you wish.  Not all platforms will have the same limitations on the number of updates possible per packet.

p4-16/psa/PSA.mdk
jafingerhut: So is there any point in the P4 program specifying the width of counters _at all_?  If the control plane will expose them as 64-bit value regardless of underlying width, do you want to just let the compiler pick the hardware width and hide that detail from the programmer?

If you want to require that the P4 program specify the hardware counter width, why?

vgurevich: The reason to specify width is to allow the compiler to properly allocate necessary resources, because practically speaking no compiler can make a good decision without a hint.

p4-16/psa/PSA.mdk
```diff
@@ -523,34 +523,71 @@ extern Checksum<W> {
...
-    packets,
-    bytes,
-    packets_and_bytes
-}
-```
 
 ### Counter
 
 ```
-extern Counter<W, S> {
```
vgurevich: Can we also decide on the naming conventions? For example, is it Counter or something like {Psa, PSA_, }Counter[_t] ?

Alternatively, I can open a separate ticket and we'll do a wholesale renaming of everything.

p4-16/psa/PSA.mdk
jafingerhut: But a programmer can make a bad decision that will guarantee information loss, i.e. a control plane that cannot read the hardware counters often enough to avoid missing many updates.  e.g. Nothing is stopping me (except the compiler) from defining a 2-bit wide packet count, and if you want it to wrap, the control plane has 0 chance of reading it as fast as it can wrap.  If the compiler knows that is too small, then it must know what is big enough?

p4-16/psa/PSA.mdk
jafingerhut: I am happy to follow whatever naming convention is decided upon.  I'd recommend a separate issue for it, since it will have a global effect on the document.

p4-16/psa/PSA.mdk
vgurevich: @jafingerhut ,

Programmer can, indeed, make a bad decision, but I thought we agreed that we can't control that :) Also, if you look at the equivalent attribute in P4_14 spec, you will see that it means "minimum width", meaning that the compiler is free to allocate more bits, just no less than what was specified.

Generally speaking, the compiler simply does not have a lot of meaningful info to use for this decision. It only knows what HW allows and, again, typically it will offer some options to choose from, but rarely it will support all widths from 1 to N and most typically the HW will not support widths that are too small.

p4-16/psa/PSA.mdk
vgurevich: @jafingerhut 

The point is that the counter value is associated with the action, and more specifically the action data, not the match entry. While in the most common case, there is 1:1 correspondence, it might not always be true.

p4-16/psa/PSA.mdk
vgurevich: True, but then again, how can we define PSA compliance?

p4-16/psa/PSA.mdk
vgurevich: Actually, @hanw already opened an issue #315, based on my previous comments, so we are good.

p4-16/psa/PSA.mdk
jafingerhut: On other issues, you seem quite willing to give a wide range of options to an implementation (e.g. updating a byte counter adds some value unknown to the programmer to the byte count value, unless they read the documentation for that implementation, and it happens to explain it).  Why should this issue be different if some implementations allow multiple updates per packet, and others limit it to 1?

jafingerhut:
For those interested in this one, I'd recommend reading the new separate PR here: https://github.com/p4lang/p4-spec/pull/350

If that newer one is merged in, I expect this one to be closed without merging.

p4-16/psa/PSA.mdk
vgurevich: @jafingerhut 

I think I am still struggling to internalize the scope and the goals of the PSA project and the comment above is the indication of this struggle. We had some discussion, but I do not believe we codified that. 

We are working on *Portable* Switch Architecture, so I guess that portability is one of the the goals (if not the main one) :) The question is how do we define portability and, subsequently PSA-compliance?

Usually, we can define portability as the ability of the program to produce the same results regardless of the target. As a consequence, we can check the compliance by ensuring that all required objects/primitives/functions have been defined and pass the functional tests. The ability of the program to be compiled is not questioned or it is easily quantifiable (e.g. C specification states that the compiler should accept identifiers that are at least 32 characters long, that it should accept programs that have at least 50 levels of statement nesting, etc.) 

In P4 world, the simple ability of the program to compile is actually not given, at least not for the ASIC targets (as opposed to SW models, NPUs, FPGAs, etc.). Therefore, the PSA implementation for a given target can have all the objects correctly defined and yet a program, utilizing PSA architecture and developed on a different target might not at all be compilable and not only because of the lack of resources, but because it violates other kinds of restrictions (can't increment the counter more than once, can increment, but only by 1, can increment, but only if it is action data and not coming from metadata, etc.). 

I am afraid that if we can't define PSA in a way to ensure reasonable compliance, it will become useless, because people will have no confidence that their programs could still work anywhere outside the original target and then they won't bother and claims that certain targets support PSA will also become hollow at best and misleading at worst. 

I am not sure we have a lot of good options here, given what I know about ASIC targets at least, but one way I am trying to explore is by limiting functionality to such a bare minimum that it can be implemented everywhere and fairly easily. That was the basis for my suggestion to get rid of byte counters (or accepting not-very-well-defined notion of packet length) and this is the basis for the comment above. 

The other important part of this whole endeavor is also the desire to ensure that a control plane, written to control a P4 program running on target A under PSA can control that same program on target B, meaning that the control plane interface to both standard P4 objects (tables) and PSA externs must be standardized. That's a very tall order, but without it PSA will probably not be very useful either. That's why I am a little worried that we started the effort without a good way to formalize the APIs -- comments and annotations do not seem to be enough, at least not to me.

Or, we can go back and scale down the goals. 

What do you think?

cc10512: @vgurevich 

To address your concerns on portability: I agree with both your points -- PSA should define sufficient functionality to be useful and that can be ported to different targets. And that we should strive to hide the target from the control plane, such that not only P4 programs are portable, but as much of the control plane is portable as well.

As you know, I have been pushing for getting the control plane API in the PSA spec so that one can look at the PSA spec and know how to use these externs without leafing through several target reference manuals. 
However, I do not think we are ready to change the language to support the expression of control plane APIs in P4, and certainly not in the timeframe we are discussing to push out a first draft of the PSA. Given these constraints, I think we should have at least a textual definition of these APIs. I'm sure it is not fully satisfactory, but it is a reasonable compromise toward getting people's mindshare.

p4-16/psa/PSA.mdk
cc10512: I was thinking we can push the W as a method parameter type, as it is used only in the control-plane. So we are left with S and V, and on the control-plane APIs we'll have something along the following lines:
```
W read<W> (in S index);
```
Moreover, such syntax is currently supported in the compiler.

p4-16/psa/PSA.mdk
cc10512: I was just asking, I do not know enough of networking apps to have a strong opinion one way or another. Seems reasonable to me, except, of course if hardware limitations on some targets decide that saturating operations are something they can live without.

p4-16/psa/PSA.mdk
cc10512: @vgurevich actually we can control the programmer's behavior to a certain extent by making it harder to code in a bad way. Giving them safe default behavior and the threads to build the rope to hang themselves will always push them toward the safe behavior because it's too much work to twist enough rope.

p4-16/psa/PSA.mdk
vgurevich: @cc10512 

In that case, I would suggest we try to stay with minimal useful functionality (AKA Lowest Common Denominator), both on the data plane and control plane side. That's the only way to achieve these goals.

p4-16/psa/PSA.mdk
vgurevich: @cc10512 

Totally agree on the defaults. 

Applying this, plus the principle of minimal functionality to this particular case, I'd suggest removing the counter width altogether as @jafingerhut suggested earlier. Most probably the compiler will be forced to use the widest possible counters supported by the HW, but we already agreed that PSA is not going to produce the most efficient programs, so that's OK.

In the worst case, implementors might decide to provide annotations instead, but these won't be a part of PSA.

p4-16/psa/PSA.mdk
jafingerhut: Thanks for the thoughtful comments here.  My apologies if I'm a little too argumentative on some of these points at times.

I do like the goal of a PSA that defines externs, rules for using those externs in P4 programs such that those P4 programs should be portable across PSA implementations, as well as control plane APIs that are portable across PSA implementations.  I'd be happy if that were some kind of 'mission statement' for the PSA.

While I consider it slightly unfortunate if we have small differences between PSA implementations, such as using different values for packet lengths on counter and meter updates, I also think most people could be content with such differences, more often than not.

Issue #308, opened 2017-06-25
[PSA] count() operation on counters


hanw:
For a 'bytes' or 'packets_and_bytes' counter type, count() having a
second parameter 'increment' that specifies how much to add to the
byte counter is a very good thing.  It is best if the P4 program has
the flexibility to choose the length in bytes it wants to use for the
packet, e.g. in case it is increasing or decreasing the received
packet's size before transmitting it.

For counters that have type 'packets', presumably every call to
'count' will add exactly 1, yes?  In that case, having a separate
method call that only takes an 'index' parameter would be best, since
the 'increment' parameter value would be ignored, anyway.

jafingerhut:
Proposal:

+ For direct counters, allow the counter_type to be 'packets' only -- do not support counter_type 'bytes', nor 'packets_and_bytes'.  If you want those counter types, use a regular counter instead of a direct counter.  (If someone wants to allow those counter types for direct counters, then they must define what value will be added to the 'bytes' counter when a packet searches the table.)  Remove method 'count()' from the extern object definition, if there is no reason for it, e.g. if it is automatically called whenever the corresponding table is searched, and should not be called from the user's P4 program directly, then having such a method only leads to confusion for why it is there.

+ For 'regular' counters (i.e. not direct counters), only support the existing 'void count(in S index, in W increment);' method if the counter_type is 'bytes' or 'packets_and_bytes'.  The value 'increment' is the value added to the bytes counter updated, and 1 is always added to the packets counter.  If the counter_type is 'packets', add a new method 'void count(in S index);' that always adds 1 to the counter specified by 'index'.

+ Do we want to specify that hardware counters are all saturating, i.e. 'stick' at their maximum values rather than wrap around?  If we want the option to have either saturating or wrapping options, it would be good to put that in the PSA spec.

vgurevich:
I would make even a stronger statement -- we should probably start defining packet counters only.

Byte counters (or packet-and-byte) should be defined separately and they will probably be quite restricted, based on what we know about high-speed pipelines and their implementations. The reason I am saying that is because more often than not, packet length is not even available during packet processing (especially at the ingress). Also, it can change when headers are added or removed. Usually, HW uses special mechanisms to account for that that are not necessarily flexible or compatible across multiple different implementations.

jafingerhut:
@jnfoster @vgurevich @cc10512  (others welcome to answer -- I just want to call attention to the questions below to these people, who expressed interest in this topic)

[...]

PR #350, opened 2017-07-13
Rewrite of new PSA.mdk Counter section

jafingerhut:
This is much less of a change to the original than my first draft, so
probably less controversial.

### DIFF
--- a/p4-16/psa/Makefile
+++ b/p4-16/psa/Makefile
@@ -7,6 +7,12 @@ build/${SPEC}.pdf: ${SPEC}.mdk
 	madoko --pdf -vv --odir=build $<
 
 build/${SPEC}.pdf: p4.json
+build/${SPEC}.pdf: psa.p4
 
 clean:
 	${RM} -rf build
+
+P4C=p4test
+
+check:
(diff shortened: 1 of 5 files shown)


jafingerhut:
This proposed change is significantly different than the changes proposed in my earlier PR here: https://github.com/p4lang/p4-spec/pull/348

I therefore decided to open a separate PR for this one, in hopes it would be less confusing.  I think this addresses most, if not all, of the comments on that earlier PR, but read it over and see what you think.

jafingerhut:
Antonin Bas wrote a comment on a related issue here: https://github.com/p4lang/p4-spec/issues/308#issuecomment-314860638

Quoting him here: "in bmv2 direct counters do not require calling count explicitly in the action, because this is the way P4_14 works. The simple_switch count action primitive (https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/primitives.cpp#L291) always takes an index because it can only be used for indirect counters."

jafingerhut:
Closing this mess of a rebase attempt in favor of this one: https://github.com/p4lang/p4-spec/pull/352

PR #352, opened 2017-07-18
Proposed updates for Counter and DirectCounter section

jafingerhut:


jafingerhut:
I had the recipe for rebasing a PR, but seem to have misapplied it this time.

This proposed change is significantly different than the changes proposed in my earlier PR here: https://github.com/p4lang/p4-spec/pull/348

I therefore decided to open a separate PR for this one, in hopes it would be less confusing. I think this addresses most, if not all, of the comments on that earlier PR, but read it over and see what you think.

p4-16/psa/PSA.mdk
```diff
@@ -469,12 +510,156 @@ have an instance for each entry in the table.
...
+one associated with the table entry that matched.
+
+TBD: How to describe which counter value is updated for tables with
+action profiles and direct counters?  Or should this combination even
+be allowed?
+
+An action of an owner table need not have `count` method calls for all
+of the `DirectCounter` instances that the table owns.
+
+The reference implementation for the `DirectCounter` extern is
```
vgurevich: Typically, the implementation of a `DrectCounter` uses the same techniques as the corresponding implementation of a `Counter` for a given target, although this is not necessarily required

p4-16/psa/PSA.mdk
```diff
@@ -469,12 +510,156 @@ have an instance for each entry in the table.
...
+essentially the same as that for `Counter`, including the saturating
+behavior.  Since there is no `index` parameter to the `count` method,
+there is no need to check for whether it is in range.
+
+TBD: What to say about `entry_key` type?  Does it need to be made more
```
vgurevich: Direct counters should be addressed using the same method as used by the control plane to access the action data portion of the corresponding table.

For tables without special "implementation" property, it is whatever is used to access table entries (typically this is either the whole key or a special opaque entry handle)
For tables with "implementation" property (e.g. action profile) it is whatever mechanism used to access individual members of that action profile (typically a member handle or a member index)

p4-16/psa/PSA.mdk
jafingerhut: Would your concerns be addressed if the spec called it an 'example implementation' instead of 'reference implementation'?

p4-16/psa/PSA.mdk
jafingerhut: Have you looked at the control plane APIs specified in Google Protocol Buffers syntax at the link below, to see if they match your expectations regarding action profiles?

https://github.com/p4lang/PI/blob/master/proto/p4/p4runtime.proto

The p4-api working group recommends trying to keep the P4_14 and P4_16 PSA control plane APIs the same for things that are common between them, e.g. counters, meters, registers, unless there is a good reason they cannot be made the same.

p4-16/psa/PSA.mdk
vgurevich: Sure. "example" or "sample" implementation sounds less prescriptive than "reference" one. For that same reason I'd also try to use the words "may"/"might" rather than "is"

p4-16/psa/PSA.mdk
vgurevich: Generally, for the specs I try to look at the document and not at the IDL code, even when it has comments :) and that's precisely my concern with the control plane API process. 

Looking at the file I can see that the idea is to use the match fields of the tables as the key, although I do not see this being described, e.g. in the comments to the ReadRequest or WriteRequest. This is workable, although there are a number of cases, where this is quite limiting. In many cases, this is not very efficient either, especially when the key is wide. There are advantages too, no doubt about that, for example control plane can be more-or-less stateless if it wishes to (as opposed to the situation where an opaque, fiex-length ID is returned for every entry added to the table). How we are going to converge and how acceptable this will be to a wide range of potential users is another challenge.

Nevertheless, this answers our question about the width of the entry_key -- it's whatever is specified in FieldMatch of the corresponding table plus priority for some tables (again, something not articulated in the code). 

This, and other "trivial" things like the lack of any error code in the WriteResponse or ReadResponse  are precisely the reasons we either need to work with the API group really closely or leave API out for now (but I'd prefer the former option).

jafingerhut: I would also prefer that we work with the p4-api group, and include the best answers we can think of for the PSA control plane APIs in the PSA spec, even including excerpts in the PSA document of the Protobufs .proto files instead of the current control plane API excerpts now in the draft PSA document.

p4-16/psa/PSA.mdk
jafingerhut: I made this update.

p4-16/psa/PSA.mdk
```diff
@@ -450,11 +450,52 @@ Parameters:
...
+    inside the actions of the table with which it is associated, and
+    this always updates the counter value associated with the matching
+    table entry.
+  - For an indirect counter, you may invoke its `count` method
+    anywhere in the P4 program where extern object method invocations
+    are permitted (e.g. inside actions, or directly inside a control's
+    `apply` block), and every such invocation must specify the index
+    of the counter value to be updated.
+
+Counters are only intended to support packet counters and byte
```
samar-abdi: Are we mandating the bitwidth of a counter? If not, should it be a parameter of the extern constructor?

p4-16/psa/PSA.mdk
```diff
@@ -469,12 +510,164 @@ have an instance for each entry in the table.
...
+this table the `DirectCounter` instance's "owner".  It is illegal to
+call the `count` method for a `DirectCounter` instance anywhere except
+inside an action of its owner table.
+
+The counter value updated by an invocation of `count` is always the
+one associated with the table entry that matched.
+
+TBD: How to describe which counter value is updated for tables with
+action profiles and direct counters?  Or should this combination even
+be allowed?
```
samar-abdi: The direct counter values can be incremented explicitly in actions that are part of the action profile. I suppose I don't follow why this is a concern.

p4-16/psa/PSA.mdk
samar-abdi: Taking this back, as it is already covered in psa.p4.

p4-16/psa/PSA.mdk
jafingerhut: If you (or anyone) could describe a brief example of a table with an action profile and a direct counter, and how to determine exactly how many counter states exist at any given time for that direct counter (i.e. is it the number of search keys that have been added to the table, but not yet removed?  Something else?) and which one is updated when the `DirectCounter`'s count() method is called, I would be grateful.  My current problem is that I don't have a clear mental model of how action profiles and action selectors work.  Frankly, anyone who could write up some examples of those and how they work, with example sequences of control plane APIs and what effect they have on the state of the system, that would be great.

samar-abdi: Hi Andy, 

Below is a sample v1 model program that uses action selector and direct counter. Direct counters and meters are resources tied to the table entry, and can only be addressed by the match key for that table. They are not tied in any way to action profile members or groups. 

For the foo_table (below), the table entries need to be created after an action profile member/group has been created. P4Runtime provides a TableEntry object that refers to the member or group by its controller assigned id (which should not be confused with the P4 compiler assigned ID for the action_selector).

A controller api would add group table entries as follows (canned syntax):
add_action_profile_member <action_profile_id>, <controller_assigned_member_id>, <action_id>, <action params ...>
add_action_profile_group <action_profile_id>, <controller_assigned_group_id>, <controller_assigned_member_ids....>
add_table_entry <table_id>, <match_key...>, <controller_assigned_group_id>

If a packet matches on the table_entry added above, it would execute the action_selector algorithm on the group, which would pick the action profile member from the group's members, and execute the corresponding action, thereby increment the direct counter associated with the table_entry.
 
-----------------------------------------------------------------
control foo_ctrl(inout parsed_packet_t hdr) {
  direct_counter(CounterType.packets) foo_counter;
  action foo() {
    foo_counter.count();
  }
  table foo_table {
    key = {
      hdr.ipv4_base.dst_addr : lpm;
      hdr.ipv4_base.src_addr : selector;
    }
    actions = {
      foo;
    }
    counters = foo_counter;
    implementation = action_selector(HashAlgorithm.crc16, 32w1024, 32w14);
  }
  apply {
    foo_table.apply();
  }
}

p4-16/psa/PSA.mdk
```diff
@@ -471,12 +510,164 @@ addressed using P4Runtime table entry as key.
...
+table has a `default_action` table property with an action assigned to
+it in the P4 program, or the control plane has made an explicit call
+to assign the table a default action.  If neither of these is true,
+then there is no default action assigned to the table.
+
+TBD: Verify that the method of reading this default action counter
+state is documented for the control plane API.  I believe that Antonin
+Bas said that it can be accessed using the same API call used to read
+a `DirectCounter` value associated with a table entry, except that the
+key in the API call should be empty.
```
antoninbas: correct

PR #354, opened 2017-07-20
Proposed text to define a ValueSet extern in PSA

jafingerhut:


p4-16/psa/PSA.mdk
```diff
@@ -590,6 +590,92 @@ stream (copying or redirecting).
...
+`select` statement that is matching on a variable returned from an
+`is_member` method call?
+
+TBD: Should it be allowed to perform more complex calculations on the
+return value of `is_member` before using it in a `select` expression?
+Eventually I am sure some PSA implementations would (reasonably)
+reject such things as too complex.  Here is an example that might not
+be too complex, but you can easily imagine similar code that is more
+complex:
+```
```
vgurevich: Although the language allows that, PSA-compliant programs should not attempt to use the value, returned by the is_member() method in any contexts, other than the transition statement. Doing otherwise, can lead them to be non-portable across different architectures, especially those that do not/cannot explicitly generate such a result.

jafingerhut: This question is not directly related to ValueSet, but is related to your comment above in that it is a question about whether you think all PSA compliant devices should be able to do 'select' statements on local variables in a parser.  In particular, do you think the P4_16 program linked below should be portable across PSA devices?  The only potentially tricky part I can see is the handling of the 'tcp_hdr_bytes_left' local variable in the parser.

https://github.com/jafingerhut/p4-guide/blob/master/tcp-options-parser/tcp-options-parser.p4#L151-L169

If you _do_ think that such a program should be portable across PSA-compliant devices, then what do you see as the fundamental difference between the kind of code in your comment above, and the one for parsing TCP options?

vgurevich: @jafingerhut 

The fragment you are referring to is totally OK (although I am sure that not all otherwise PSA-compliant devices will be able to perform such arithmetic easily).

I was referring to the fact, that that the way ParserValue sets are **typically** implemented in modern parsers (and I've seen a few) is that the match result cannot be easily retrieved as a number/boolean and therefore stored in a local variable for subsequent manipulation. It can be done (if push comes to shove) by synthesizing extra states, but will be extremely expensive.

cc10512: @vgurevich, I seem to recall that we also discussed this in the language and that assignment was one of the push backs. So, indeed, we should only allow `is_member` in a `select` expression. We can go to the "extreme" of rewriting Example 2 and inline the calls, so that we make it clearer that is the only supported use.

vgurevich: @cc10512 Correct. Exactly for that reason.

p4-16/psa/PSA.mdk
```diff
@@ -590,6 +590,92 @@ stream (copying or redirecting).
...
+that matches on multiple fields, by making the type `D` a `struct`
+containing multiple bit vectors.
+
+```
+[INCLUDE=examples/psa-example-value-sets3.p4:ValueSet_Example_3]
+```
+
+TBD: Should PSA allow code in the style of the second example, where
+multiple boolean variables can be in the `select` expression?  I can't
+think of a strong reason why not.
```
jafingerhut: I will leave the example in, and remove the TBD, assuming for now that PSA code should allow that kind of code.

p4-16/psa/PSA.mdk
jafingerhut: I will remove the TBD, and add a note that PSA-compliant implementations are not required to support any use of ValueSet `is_member` return values except directly in `select` expressions (OK if that is through assignment to a named variable, though).

p4-16/psa/PSA.mdk
```diff
@@ -590,6 +590,92 @@ stream (copying or redirecting).
...
+
+TBD: Should PSA allow programs to use the value `false` in a case of a
+`select` statement that is matching on a variable returned from an
+`is_member` method call?
```
jafingerhut: OK, thanks for the answer.  Removing that TBD from the PR.

PR #355, opened 2017-07-20
Fix cross-references within the PSA to use Madoko markup

jafingerhut:


jafingerhut:
The @see <symbol> occurrences in PSA.mdk are not Madoko markup, as far as I can tell.  Please point me at Madoko docs describing them if I missed it.  At least right now, they turn into "@see <symbol>" text in the resulting PDF output.

PR #357, opened 2017-07-21
Proposed draft of a PSA Portability section for the PSA spec

jafingerhut:


jafingerhut:
A lot of the numbers in here are just made-up placeholders, although I believe some of them may be close to what we would want.

p4-16/psa/PSA.mdk
```diff
@@ -608,3 +608,209 @@ matches the in parameter of the subsequent block.
...
+
+- Applies the same table more than 1 time per packet.
+- Calls `getHash` on the same `Hash` instance more than 1 time per
+  packet.  (TBD: Is this a useful restriction?)
+- Calls `clear` or `get` on the same `Checksum` instance more than 1
+  time per packet, or calls `update` and `remove` more than a total of
+  6 times per packet.  (TBD: Is this a useful restriction?)
+- Updates the same `Counter`, `DirectCounter`, `Meter`, `DirectMeter`,
+  or `Register` instance more than 1 time per packet.
+- Uses the same `ActionProfile` for more than 1 table.
```
cc10512: No shared ActionProfiles? I thought that sharing an ActionProfile by multiple tables was a desired feature that should be supported by a compliant PSA architecture. I realize that you made the rejection optional for this, so I'm actually wondering whether this should be dropped altogether. Similar for ActionSelector.

p4-16/psa/PSA.mdk
```diff
@@ -608,3 +608,209 @@ matches the in parameter of the subsequent block.
...
+limitations should be portable across all PSA compliant
+implementations.
+
+Minimum requirements for packet parsing:
+
+- 32 bits in a parser `select` expression.
+- 50 parser states.
+- 20 parser state transitions per packet.
+- 128 bytes extracted per packet.
+- 16 bytes of metadata assigned in the parser.
```
jafingerhut: I suppose there are parsers that pick one of several IPv6 headers from which to assign the SA and DA to metadata variables, plus one of several L4 headers from which to assign source and dest port numbers?  That would total 16+16+2+2 = 36 bytes, and we could round up to 40 bytes for a bit more headroom above that.

But feel free to suggest a number other than 40 bytes if you'd be more comfortable proposing that as a PSA minimum.

p4-16/psa/PSA.mdk
jafingerhut: I am happy to defer to others on proposed PSA minimums for ActionProfile and ActionSelector until I get a better mental handle on how they can be implemented.

I would guess that if an ActionProfile is used by more than 1 table, there would effectively need to be 1 copy of its contents per table it is used by, in a 1-packet-per-clock-cycle hardware implementation?

p4-16/psa/PSA.mdk
jafingerhut: And in case it isn't clear from the draft text, I am expecting _many_ of these proposed numbers to change based upon review comments.  Most of the initial values I tossed in there are intended as starting points for discussion.

p4-16/psa/PSA.mdk
antoninbas: The ability to share action profiles between is a big reason for why they exist in the first place. @jafingerhut a target is free to implement this sharing as it wants, duplicating the contents of the action profile in memory is not always needed. Even when you are only allowed one memory lookup per clock cycle, _mutually exclusive tables_ can share an action profile without the need for duplication.

jafingerhut: @antoninbas This is a different question not directly related to action profiles, but somewhat related (the answer to this question might inform the answer to whether action profiles should be allowed in the PSA more than once per packet, is the reason I ask).

Are you (or is anyone) proposing that PSA should require a PSA compliant implementation to support multiple apply() operations for the _same table instance_, if one is in the 'then' branch of an 'if' statement, and the other is in the 'else' branch of the same 'if' statement?  For example, are you proposing that this program must be supported for all PSA compliant implementations, even if there are match dependencies between every pair of tables involved?
```
if (some_condition) {
    T1.apply();
    T2.apply();
    T3.apply();
    T4.apply();
    T5.apply();
} else {
    T5.apply();
    T4.apply();
    T3.apply();
    T2.apply();
    T1.apply();
}
```
If those are the only apply() statements for the tables T1 through T5 in the entire P4 program, then it is easily provable statically that they can all be done at most once per packet, but I would guess that some hardware platforms might have a hard time implementing it.  Certainly bmv2 today does not, although that may be for reasons unrelated to why some hardware platforms might have difficulty implementing it.

p4-16/psa/PSA.mdk
```diff
@@ -814,3 +814,154 @@ only occurs once in the source code for control block `C`.  Some
...
+- Tables are always instantiated directly in their enclosing control,
+  and cannot be instantiated at the top level.  There is no type by
+  which a table can be passed as either a constructor or run-time
+  parameter anywhere else.  They are only intended to be used from
+  within the control where they are defined.
+
+The first table lists restrictions on what types can be passed as
+constructor parameters to other types.
+
+|-------------|-------------------------------------------------|
```
mbudiu-vmw: I think that this table should be moved to the p4-16 spec instead. It is very good.
Maybe we can also consider listing non-constructor parameters.

jafingerhut: In case you hadn't read that far, a later table proposes restrictions on run-time parameter types, at least for PSA if not also P4_16 in general.

mbudiu-vmw: Most of these tables should be moved to the spec, and the ^P4 note should not be necessary.

jafingerhut: Regarding moving these to the P4_14 spec instead, I have no objections to doing so.  It seems worth asking the P4_16 language working group members to review it first.

p4-16/psa/PSA.mdk
antoninbas: Conceptually, it is not very hard to realize this but it is expensive. You need to duplicate each table: stage0{T1, T5}, stage1{T2, T4},...
The control plane needs to be able to provide a unified view of these tables, just like for a control that calls the same table multiple times unconditionally.
Just like for action profile sharing, if some requirements are met (in this case no dependency between tables), the implementation becomes much easier.
I think that's a broad topic that would deserve to be discussed at the next meeting.
Thinking for the long term, I think we should allow both of these scenarios in PSA. I know for a fact that people are interested in both scenarios (multiple invocations, action profile sharing). In both cases, implementation can be expensive in the general case but is not impossible. It will be up to the vendor and the compiler software they provide to guide the P4 programmer if their program written for PSA does not fit on the hardware target.
Thinking for the short-term, I would expect compiler and control-plane software to not support these scenarios in the general case at first.

jafingerhut:
At 2017-Aug-07 PSA working group meeting, decision was made to move all of the minimal proposed scale numbers into a later "PSA profile" document.  One advantage of this is to get a PSA v1.0 out the door sooner.  Another is to decouple changes in these profiles from the definition of PSA standard metadata and externs.

Regarding the last part of this PR, regarding what kind of stateful objects (packages, control, parsers, and externs) are allowed as constructor parameters, run-time parameters, and return values, those should still be considered for inclusion in either the PSA, or if they are worth adding as restrictions for the P4_16 language itself, then in the P4_16 language spec.

jafingerhut:
Issues created to put this text in other documents:
- https://github.com/p4lang/p4-spec/issues/375
- https://github.com/p4lang/p4-spec/issues/376
- https://github.com/p4lang/p4-spec/issues/377

PR #366, opened 2017-08-01
Proposed addition of timestamps to PSA

jafingerhut:


jafingerhut:
Instead of the following proposed conversion extern method:
```
    void timestamp_to_nanoseconds(in timestamp_t t,
                                  out bit<64> seconds,
                                  out bit<30> nanoseconds);
```
it is likely easier to implement this one instead:
```
    void timestamp_to_nanoseconds(in timestamp_t t,
                                  out bit<80> nanoseconds);
```
where the output value 'nanoseconds' is the total number of nanoseconds elapsed since Jan 1, 1970.  Any implementation of the first method above would probably require first going through the calculation of required for the second method, then separating the total nanoseconds into whole seconds and nanoseconds in the range [0, 10^9-1], which requires additional arithmetic.

vgurevich:
@jafingerhut ,

I would suggest replacing "Jan 1, 1970" with "the time when the timestamp was equal to 0". This will be mush more reasonable to implement (although still quite difficult). 

In reality, though, I suspect that aside from the application layer protocols that really need time-of-day based values (but usually are not the ones handled by P4 programs) the scaling will be done at the control plane level. For example, to calculate the timeout, a difference timestamp2-timestamp1 will be compared with the number calculated as timeout_in_ns * NUMBER_OF_TICKS_IN_NS

jafingerhut:
@vgurevich Suggestion noted.  Doing only part of a conversion to some common external reference, but not all of it, doesn't seem like very much of a benefit for external data analysis software to me, but I'm open to suggestions.  All of the conversions are described as optional in the draft version of the section.  Implementing an absolute reference time like Jan 1, 1970 would require driver and/or control plane involvement to update conversion parameters 2 or 3 times per timestamp wrap time.  The multiplication or division would be the most difficult part, I would think, and once you have done that addition by a control-plane provided constant is much lower cost than that.

DGeorgeNetro:
I agree with @vgurevich about having a non-unix epoch start time. For this to be really useful one would have to do fairly onerous time synchronization.

In the case of in-band telemetry, the usefulness of nanoseconds vs seconds and nanoseconds depends on the protocol, for example IOAM uses 32-bit seconds and 32-bit nanoseconds
https://tools.ietf.org/html/draft-brockners-inband-oam-data-07
How does one decide which is better? I would say seconds and nanoseconds is a more comfortable format.

I disagree with:
"Any implementation of the first method above would probably require first going through the calculation of required for the second method"
It is entirely possible that hardware implement a 32-bit nanosecond and 32-bit second counter. The divide to go from nanoseconds to seconds is very more expensive than the multiply to go to all nanoseconds.

jafingerhut:
@DGeorgeNetro Do you know any existing examples of hardware providing a separate nanosecond and second counter?  I'm just curious to broaden my horizons here.  It is trivial for an ASIC or FPGA to keep a count of clock cycles, but those will only provide a nanosecond counter if they happen to run at 1.000 GHz clock rate.

p4-16/psa/psa.p4
```diff
@@ -373,6 +377,70 @@ extern ValueSet<D> {
...
+    void timestamp_to_microseconds(in timestamp_t t,
+                                   out bit<64> microseconds);
+    void timestamp_to_milliseconds(in timestamp_t t,
+                                   out bit<64> milliseconds);
+    void nanoseconds_to_timestamp(in bit<64> nanoseconds,
+                                  out timestampt_t t);
+    void microseconds_to_timestamp(in bit<64> microseconds,
+                                   out timestampt_t t);
+    void milliseconds_to_timestamp(in bit<64> milliseconds,
+                                   out timestampt_t t);
```
cc10512: Are these APIs really needed in the data plane? If you had just the timestamps you can get the relative delta between events, is it important in the data plane to relate them back to [m,u,n]sec?

jafingerhut: It depends upon what time-based features you want to enable in the data plane.

The only features _required_ by this draft text are platform-dependent timestamp_t values, and yes those should be enough to calculate elapsed time between two events in your P4 program, so definitely enough for doing timeouts, measuring one-way latency in a single device, etc.  It also enables you to export device-dependent units of queue latencies in INT headers.

If that is all the features anyone wants to make "PSA portable", then great.

These externs are written in this draft as _optional_, not required.  My only reasons for even mentioning them is that they could enable additional features to be implemented, on those platforms that implement these externs.  If we don't say anything in PSA about them, they are more likely to be completely different APIs and behavior across implementations.

If that is too much to hope for in PSA, then we can just drop them.

jafingerhut:
I have removed what appears to be the most controversial part of these changes for now: the conversion functions between timestamps and other time units.

If someone has ideas for a more generally agreed-upon way to do this, they can open a separate issue or pull request for it.

DGeorgeNetro:
@jafingerhut Netronome's hardware use seconds and nanoseconds with the idea of supporting IEEE1588v2/PTP which uses a seconds/nanoseconds format. I couldn't say exactly what other hardware out there supports it.

You are quite right that this implies supporting a 1 GHz internal clock, FPGAs can achieve this using their configurable PLLs. An ASIC would have have to have the feature burned in.

I would suggest a generic timestamp_t time with target-defined width (as is already in place) and both conversion functions, timestamp_to_nanoseconds() and timestamp_to_timespec(). (where timespec is a type with seconds and nanoseconds, like Linux timespec).

Having both means the most logical for the protocol can be used and an unnecessary conversion from what the target provides, to an intermediate format and back wouldn't happen.

I would also leave the details of conversion out of the spec as this is target specific. A target supporting timestamps would have to make sure the above works and should not require the user to do anything.

jafingerhut:
@DGeorgeNetro As mentioned above, if you want this to be part of the PSA spec, I would recommend opening a new issue or pull request with the suggested changes.  It doesn't require a 1 GHz internal clock if one is willing to dedicate gates for a large enough integer multiply or divide by the appropriate constant, or some other mechanisms.  I suspect one issue might be that _without_ some notes on implementation details of how to do it written down somewhere (whether in the PSA, or somewhere else), everyone will have their own favorite implementation in mind that works easily on the implementation they think about most often, and disregard difficulties it might cause on others.

PR #374, opened 2017-08-07
Fixed documentation of controller API for direct resources to use P4R…

samar-abdi:
…untime keys.

### DIFF
--- a/p4-16/psa/psa.p4
+++ b/p4-16/psa/psa.p4
@@ -33,7 +33,6 @@ typedef bit<14> PacketLength_t;
 typedef bit<16> EgressInstance_t;
 typedef bit<8> ParserStatus_t;
 typedef bit<16> ParserErrorLocation_t;
-typedef bit<32> entry_key;           /// for DirectCounters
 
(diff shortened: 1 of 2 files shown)


p4-16/psa/psa.p4
```diff
@@ -236,10 +234,8 @@ extern Meter<S> {
   @ControlPlaneAPI
   {
     reset(in MeterColor_t color);
-    setParams(in S committedRate, in S committedBurstSize
-              in S peakRate, in S peakBurstSize);
-    getParams(out S committedRate, out S committedBurstSize
-              out S peakRate, out S peakBurstSize);
+    setParams(in MeterConfig);
+    getParams(out MeterConfig);
```
jafingerhut: The DirectMeter has the ability to set parameters independently for each entry, via the 'in TableEntry entry' parameter to setConfig.

I would have guessed that this same level of configurability would be possible for the Meter extern, by having say an 'index' parameter, maybe type uint32 if some other type is not more appropriate.

samar-abdi: Thanks. Fixed.

PR #379, opened 2017-08-13
Some additions to description of meters.

jafingerhut:
This change also includes a proposed change of naming things "direct counters" and "indexed counters" (the latter name "indexed" in place of what has previously been called "indirect").  It isn't a big deal, and I can back out that part to go back to indirect if that is what people want.  I just wanted to propose the name "indexed", since it doesn't have any connotations of adding a level of indirection that the name "indirect" can.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -716,7 +716,8 @@ must not use the undefined value in a way that affects the output
 packet or other side effects.  The example code below shows one way to
 achieve predictable behavior.  Note that this undefined behavior
 cannot occur if the value of `n_meters` of an indexed meter is $2^W$,
-and the type `S` used to construct the meter is `bit<W>`.
+and the type `S` used to construct the meter is `bit<W>`, since the
+index value could never be out of range.
 
 ```
 const int METER1_SIZE = 100;
@@ -749,8 +750,8 @@ Implementations will also have finite ranges and precisions that they
 support for the Peak Information Rate and Committed Information Rate.
 An implementation should document the maximum rates it support, as
 well as the precision it supports for implementing requested rates.
-It is recommended that the maximum rate supported be equal to the
-implementation' maximum speed port, and that the actual implemented
+It is recommended that the maximum rate supported be at least the rate
+of the implementation's fastest port, and that the actual implemented
 rate should always be within plus or minus 0.1% of the requested rate.
 
 ### Meter types
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -662,9 +662,9 @@ updating of `Counter` and `DirectCounter` externs.
 ## Meters
 
 Meters (RFC 2698) are a more complex mechanism for keeping statistics
-about the packets that trigger a table, most often used for dropping
+about packets, most often used for dropping
 or "marking" packets that exceed an average packet or bit rate.  To
-"mark" a packet means to change one or more of its quality of service
+mark a packet means to change one or more of its quality of service
 values in packet headers such as the 802.1Q PCP (priority code point)
 or DSCP (differentiated service code point) bits within the IPv4 or
 IPv6 type of service byte.  The meters specified in the PSA are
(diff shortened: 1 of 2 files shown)


p4-16/psa/PSA.mdk
```diff
@@ -451,25 +451,23 @@ sequence numbers in packets, for example, use Registers instead
 
 Direct counters are counters associated with a particular P4 table,
 and are implemented by the extern `DirectCounter`.  There are also
-indirect counters, which are implemented by the extern `Counter` (the
-name 'indirect' does not imply that there is any indirection involved
-in the implementation, only that they are not direct counters).  The
-primary differences between direct counters and indirect counters are:
+indexed counters, which are implemented by the extern `Counter`.  The
+primary differences between direct counters and indexed counters are:
```
cc10512: Shall we consider renaming `Counter` to `IndexedCounter`? It is clear from your last sentence that counters need an attribute (direct or indexed) so it would be nice to make those part of the name. I realize that this might be too much of a change, however, I wanted to throw it out there for discussion.

p4-16/psa/PSA.mdk
```diff
@@ -663,13 +661,98 @@ updating of `Counter` and `DirectCounter` externs.
...
+values in packet headers such as the 802.1Q PCP (priority code point)
+or DSCP (differentiated service code point) bits within the IPv4 or
+IPv6 type of service byte.  The meters specified in the PSA are
+3-color meters.
+
+PSA meters do not require any particular drop or marking actions, nor
+do they automatically implement those behaviors for you.  Meters keep
+enough state, and update their state during `execute()` method calls,
+in such a way that they return a green (also known as conform), yellow
+(exceed), or red (violate) result.  See RFC 2698 for details on the
```
cc10512: Capitalize GREEN, YELLOW, RED? They are specified as such in the enum.

p4-16/psa/PSA.mdk
```diff
@@ -663,13 +661,98 @@ updating of `Counter` and `DirectCounter` externs.
...
+must not use the undefined value in a way that affects the output
+packet or other side effects.  The example code below shows one way to
+achieve predictable behavior.  Note that this undefined behavior
+cannot occur if the value of `n_meters` of an indexed meter is $2^W$,
+and the type `S` used to construct the meter is `bit<W>`, since the
+index value could never be out of range.
+
+```
+const int METER1_SIZE = 100;
+Meter<bit<7>>((bit<32>) METER1_SIZE, MeterType_t.bytes) meter1;
```
cc10512: here is a case where we should allow either implicit casts or passing `int` as a constructor parameter. There is no good reason for forcing the `bit<32>` cast. @jnfoster, @mbudiu-vmw can we agree to have this "extension" in the language?

jafingerhut: Actually, I neglected to to test compiles with latest p4c before committing this.  As of now, "const int METER1_SIZE = 100;" gives an error.

The line "Meter<bit<7>>(100, MeterType_t.bytes) meter1;" gives no error, so neither should replacing the 100 with a METER1_SIZE defined with "#define METER1_SIZE 100".

For now, I will change the example to the version with #define METER1_SIZE 100, and no (bit<32>) cast, but it would be nice here if an extern constructor could take a compile-time constant as an 'int' type parameter, without having to specify a bit width.

p4-16/psa/PSA.mdk
jafingerhut: Done.

p4-16/psa/PSA.mdk
jafingerhut: Made the updates that avoid the need for (bit<32>) cast in my example, but if you use a bit<W> type as the first argument to the Meter constructor, it still needs the cast.

I'm happy to change the PSA if we decide that P4_16 externs can have arbitrary precision 'int' type parameters.

p4-16/psa/PSA.mdk
jafingerhut: I don't mind renaming Counter -> IndexedCounter and Meter -> IndexedMeter if there are no objections.

PR #380, opened 2017-08-14
Proposed updates to Register section of PSA

jafingerhut:


p4-16/psa/examples/psa-example-register2.p4
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// BEGIN:Register_Example2_Part1
+const PortId_t NUM_PORTS = 512;
+
+// It would be more convenient to use a struct type to represent the
+// state of a combined packet and byte count, and many other compound
+// values one might wish to store in a Register instance.  However,
+// the latest p4test as of 2017-Aug-13 does not allow a struct type to
+// be returned from a method call like Register.read().
```
cc10512: I agree that from a language perspective you may want to track a set of values with a single register, but a hardware implementation would be more likely to break it down into the set of values. In addition, atomic updates over a set of values also add a hardware implementation burden. IIRC, we discussed having two counters for bytes and packets, which are seen from the control plane as one. Without a clear use case scenario for larger structs, I suggest we stick with an extern representation that favors a "native" hardware size.

jafingerhut: I'm not sure I understand your comment.

A Register instantiation can specify whatever number of bits it wants per entry in the Register.  Whether that looks like a bit<40> or a struct with a bit<30> and a bit<10> in the P4_16 source code shouldn't matter to the hardware.  It is a nice convenience for a P4 program writer to have the second option available, though, yes?

Issue #375, opened 2017-08-08
P4_16 spec: Restrictions on use of stateful objects


jafingerhut:
Discussion on several recent issues has raised questions about whether P4_16 the language has certain restrictions on the use of stateful objects, i.e. packages, parsers, controls, tables, and extern objects.

The text below is intended to be one possible addition to the P4_16 spec that summarizes these restrictions.  It seems worth bringing this up in a future P4_16 language work group meeting, to verify that these restrictions are accurate, and to decide upon changes to the spec to make these restrictions explicit.

Any of this that is judged too restrictive for the P4_16 language in general, should still be considered as a restriction for PSA, over and above P4_16 language restrictions.

The following issues probably address only a subset of topics that this one does, and if so, they can be closed if this larger issue is addressed:

- https://github.com/p4lang/p4-spec/issues/361
- https://github.com/p4lang/p4-spec/issues/364
- https://github.com/p4lang/p4-spec/issues/367
- https://github.com/p4lang/p4-spec/issues/368

```
## Restrictions on use of stateful types

The stateful types of objects in P4_16 are packages, parsers,
controls, externs, and tables.  All other types are referred to as
"value types" here.

Some guiding principles:

- Packages can be instantiated, but have no run-time behavior, so
  there is no use in passing them to anything else other than another
  package.
- Controls are not allowed to call parsers, and vice versa, so there
  is no use in passing one type to the other in constructor parameters
  or run-time parameters.
- Externs are not allowed to call parsers or controls, so there is no
  use in passing objects of those types to them.
- Tables are always instantiated directly in their enclosing control,
  and cannot be instantiated at the top level.  There is no type by
  which a table can be passed as either a constructor or run-time
  parameter anywhere else.  They are only intended to be used from
  within the control where they are defined.
- No recursion is allowed, neither at run time, nor in the instantiation
  process during compile time.

The first table lists restrictions on what types can be passed as
constructor parameters to other types.

|-------------|-------------------------------------------------|
|             | can be a constructor parameter for this type ||||
|             |---------|--------|---------|--------|
| This type   | package | parser | control | extern |
|-------------|:-------:|:------:|:-------:|:------:|
| package     | yes     |  no    |  no     |  no    |
| parser      | yes     | yes    |  no     |  no    |
| control     | yes     |  no    | yes     |  no    |
| extern      | yes     | yes    | yes     | yes    |
| table       |  no     |  no    |  no     |  no    |
[...]

Issue #364, opened 2017-07-31
P4_16 spec: Is it documented illegal to pass controls/parsers as run-time parameters?


jafingerhut:
In the comment linked below @jnfoster mentions that it is not allowed to pass controls to other controls as run-time parameters, and I would guess that also parsers cannot be passed to other parsers as run-time parameters.

https://github.com/p4lang/p4-spec/issues/361#issuecomment-318768783

The latest p4test compiler as of 2017-Jul-19 gives an error if you attempt to do this (the attempted definition of MyC2 in the example below causes an error):
```
#include <core.p4>

// Architecture
control C1();
control C2();
control C3();
package S(C3 c3);

// User Program
control MyC1() {
    apply { }
}
control MyC2(C1 c1) {
    apply {
        c1.apply();
    }
}
control MyC3() {
    MyC1() c1;
    apply {
        MyC2.apply(c1);
    }
}

S(MyC3()) main;
```
If this is indeed not allowed by the language, is it already documented so in the P4_16 v1.0.0 spec?  If so, where?  If not, should it be?

jafingerhut:
With the merging of this PR: https://github.com/p4lang/p4-spec/pull/427

It is now documented that a P4_16 implementation need not support passing a control to a control, nor a parser to a parser, as a run-time parameter, and that it should be supported to pass these things as constructor/compile-time parameters.

I am therefore closing this issue.

Issue #361, opened 2017-07-26
Question: Should it be possible to pass externs to controls as a parameter?


jafingerhut:
In v1model.p4 the extern packet_in is passed to the Parser: https://github.com/p4lang/p4c/blob/master/p4include/v1model.p4#L149 and similarly the extern packet_out is passed to the Deparser.

These things also occur in the current draft psa.p4, and in addition, an extern of type PacketReplication is passed to the Ingress control: https://github.com/p4lang/p4c/blob/master/p4include/psa.p4#L558

I do not necessarily _want_ to do this in my own P4_16 programs, but was curious to know whether it is legal to write my own controls that can take references to externs as parameters (and if that is the correct way to think about the examples above).

I tried to write a test program to see if the current p4test code compiles it, and wrote this: https://github.com/jafingerhut/p4-guide/blob/master/instantiation/extern-ref.p4#L64-L70

The 2017-Jul-19 version of p4test gives the error below when I tried to compile that program:
```
extern-ref.p4(70): error: Could not find declaration for rand
        my_headers.ethernet.srcAddr[47:38] = ~rand.read();
                                              ^^^^
```
Is this only allowed for some externs, but not others?

If so, how does the compiler determine which ones to allow?

If it should be possible to do, am I seeing a compiler bug, or am I writing the program incorrectly?

If it is possible to do, does this make program analysis of where externs can be accessed any more challenging for a compiler?

Is it intended to be allowed to pass controls as parameters to other controls?

jnfoster:
> Is this only allowed for some externs, but not others?

All externs should behave the same way wrt parameter passing

> If so, how does the compiler determine which ones to allow?

N/A

> If it should be possible to do, am I seeing a compiler bug, or am I writing the program incorrectly?

Compiler bug.

> If it is possible to do, does this make program analysis of where externs can be accessed any more challenging for a compiler?

Yes, but I don't see why a conservative analysis wouldn't be sufficient for most of the compiler problems I can imagine... 

> Is it intended to be allowed to pass controls as parameters to other controls?

As constructor parameters, yes.

jafingerhut:
Thanks for the answers.  I went back to re-read a few parts of the P4_16 spec to see if it is explicit whether this is allowed, and it appears that it is (if I read carefully enough).  In the "Calling convention" section, it says that parameters with no direction can be a compile-time known value (and a couple of other possibilities that are irrelevant to this example).

[...]

PR #381, opened 2017-08-14
Proposed additions to P4_16 spec: restrictions on use of stateful obj…

jafingerhut:
…ects

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6515,9 +6515,8 @@ parameters: parsers, controls, extern methods, and actions.
 | value types | yes       | yes     | yes     | yes     |
 |-------------|-----------|---------|---------|---------|
 
-The restrictions on return types of extern methods is the same as for
-parameters, with the addition that `void` can be a return type for
-methods that return no value.
+Extern method calls may return only return a value that is a value
+type, or no value at all (specified by a return type of `void`).
 
 The next table lists restrictions on what kinds of calls can be made
 from which places in a P4 program.  Calling a parser, control, or
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6479,9 +6479,10 @@ implementation.  Thus there is no place to instantiate objects within
 an extern.
 
 You may declare variables and constants of any of the value types
-within a parser or control.  Declaring a variable or constant is not
-the same as instantiation, hence the answer "no" in those table
-entries.  Variables may not be declared at the top level of your
+within a parser or control (see Section [#sec-variables] for more
+details).  Declaring a variable or constant is not the same as
+instantiation, hence the answer "N/A" (for not applicable) in those
+table entries.  Variables may not be declared at the top level of your
 program, but constants may.
 
 |-------------|-------------------------------------------------|
@@ -6494,7 +6495,7 @@ program, but constants may.
 | control     |  no       |  no     | no      | yes     |  no    |
 | extern      | yes       |  no     | yes     | yes     |  no    |
 | table       |  no       |  no     | no      | yes     |  no    |
-| value types |  no       |  no     | no      | no      |  no    |
+| value types | N/A       | N/A     |N/A      |N/A      | N/A    |
 |-------------|-----------|---------|---------|---------|--------|
 
 The next table lists restrictions on what types can be passed as
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6523,21 +6523,25 @@ from which places in a P4 program.  Calling a parser, control, or
 table means invoking its `apply()` method.  The row for `extern`
 describes where extern method calls can be made from.
 
+One way that an extern can be called from the top level of a parser or
+control is in an initializer expression for a declared variable,
+e.g. `bit<32> x = rand.get();`.
+
 |-------------|-------------------------------------------------|
-|             | can be called at run time from this place in a P4 program ||||
-|             |-----------|---------|---------|--------|
-|             |           | control |         |        |
-|             | parser    | apply   |         |        |
-| This type   | state     | block   | action  | extern |
-+-------------+:---------:+:-------:+:-------:+:------:+
-| package     | N/A       | N/A     | N/A     | N/A    |
-| parser      | yes       |  no     | no      | no     |
-| control     |  no       | yes     | no      | no     |
-| extern      | yes       | yes     | yes     | no     |
-| table       |  no       | yes     | no      | no     |
-| action      |  no       | yes     | yes     | no     |
-| value types | N/A       | N/A     | N/A     | N/A    |
-|-------------|-----------|---------|---------|--------|
+|             | can be called at run time from this place in a P4 program |||||
+|             |-----------|---------|-----------|---------|--------|
+|             |           | control | parser or |         |        |
+|             | parser    | apply   | control   |         |        |
+| This type   | state     | block   | top level | action  | extern |
++-------------+:---------:+:-------:+:---------:+:-------:+:------:+
+| package     | N/A       | N/A     | N/A       | N/A     | N/A    |
+| parser      | yes       |  no     | no        | no      | no     |
+| control     |  no       | yes     | no        | no      | no     |
+| extern      | yes       | yes     | yes       | yes     | no     |
+| table       |  no       | yes     | no        | no      | no     |
+| action      |  no       | yes     | no        | yes     | no     |
+| value types | N/A       | N/A     | N/A       | N/A     | N/A    |
+|-------------|-----------|---------|-----------|---------|--------|
 
 There may not be any recursion in calls, neither by a thing calling
 itself directly, nor mutual recursion.
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6525,19 +6525,19 @@ describes where extern method calls can be made from.
 
 |-------------|-------------------------------------------------|
 |             | can be called at run time from this place in a P4 program |||||
-|             |-----------|---------|---------|------------|--------|
-|             |           | control | action  | action not |        |
-|             | parser    | apply   | used by | used by    |        |
-| This type   | state     | block   | table   | any table  | extern |
-+-------------+:---------:+:-------:+:-------:+:----------:+:------:+
-| package     | N/A       | N/A     | N/A     | N/A        | N/A    |
-| parser      | yes       |  no     | no      | no         | no     |
-| control     |  no       | yes     | no      | yes        | no     |
-| extern      | yes       | yes     | yes     | yes        | no     |
-| table       |  no       | yes     | no      | yes        | no     |
-| action      |  no       | yes     | yes     | yes        | no     |
-| value types | N/A       | N/A     | N/A     | N/A        | N/A    |
-|-------------|-----------|---------|---------|------------|--------|
+|             |-----------|---------|---------|--------|
+|             |           | control |         |        |
+|             | parser    | apply   |         |        |
+| This type   | state     | block   | action  | extern |
++-------------+:---------:+:-------:+:-------:+:------:+
+| package     | N/A       | N/A     | N/A     | N/A    |
+| parser      | yes       |  no     | no      | no     |
+| control     |  no       | yes     | no      | no     |
+| extern      | yes       | yes     | yes     | no     |
+| table       |  no       | yes     | no      | no     |
+| action      |  no       | yes     | yes     | no     |
+| value types | N/A       | N/A     | N/A     | N/A    |
+|-------------|-----------|---------|---------|--------|
 
 There may not be any recursion in calls, neither by a thing calling
 itself directly, nor mutual recursion.
@@ -6545,20 +6545,7 @@ itself directly, nor mutual recursion.
 An extern can never cause any other type of P4 program object to be
 called.  See Section [#sec-calling-convention-justification].
 
-An "action used by a table" means an action that is either in the
-`actions` list of a table, or it is possible for the action to be
-called from such an action, perhaps through one or more intermediate
-action calls.
-
-Any action can be called from a control `apply` block.
-
-It is allowed to write actions that are not used by any table.  Such
-"actions not used by any table" have exactly the same restrictions on
-what they can call as a control `apply` block does.  A potential
-convenience of creating such an action is that it can be defined
-within the control that calls it, giving the action access to all
-parameter values and local variables defined earlier within its
-enclosing control.  By contrast, control definitions cannot be nested.
+Any action can be called directly from a control `apply` block.
 
 Note that while the extern row shows that extern methods can be called
 from many places, particular externs may have additional restrictions
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1558,7 +1558,7 @@ When used as arguments, `extern` objects can only be passed as
 directionless parameters---e.g., see the packet argument in the
 very simple switch example.
 
-### Justification
+### Justification { #sec-calling-convention-justification }
 
 The main reason for using copy-in/copy-out semantics (instead of the more common
 call-by-reference semantics) is for controlling the side-effects of `extern`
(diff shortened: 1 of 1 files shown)


jafingerhut:
Created a PR at the suggestion of @mbudiu-vmw to make it easier for others to comment on particular lines, as compared to all of the text in a single comment on this issue: https://github.com/p4lang/p4-spec/issues/375

I have made changes due to some comments from @jnfoster to make it more explicit that parser loops are allowed, and we expect architectures to define mechanism for recirculating/cloning packets, but while a control or parser is executing, there can be no recursion "beneath" that on the stack.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+| package     | yes     |  no    |  no     |  no    |
+| parser      | yes     | yes    |  no     |  no    |
+| control     | yes     |  no    | yes     |  no    |
+| extern      | yes     | yes    | yes     | yes    |
+| table       |  no     |  no    |  no     |  no    |
+| value types | yes     | yes    | yes     | yes    |
+|-------------|---------|--------|---------|--------|
+
+The next table lists restrictions on what types can be instantiated
+where.  The answer for `package` is always "no" because there is no
```
mbudiu-vmw: I would be very explicit saying that this refers to the P4 instantiation from Section 9.3 of the spec.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+"inside a package" where instantiations can be written in P4~16~.  One
+can definitely make constructor calls and use instances of stateful
+types as parameters when instantiating a package, and restrictions on
+those types are in the table above.
+
+For externs, their implementations are typically written in some other
```
mbudiu-vmw: I would say that externs only describe the interface to an extern object, and not its implementation; thus there is no place to instantiate objects within an extern.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+language, and so what can be instantiated inside of them is not the
+concern of the P4~16~ language.
+
+The entries for "value types" are yes for parsers and controls,
+because you can define variables of any of those types inside of
+parsers and controls.  Variables of these types may not be defined at
```
mbudiu-vmw: Technically you don't instantiate variables, you just declare them. Syntactically it is similar indeed. But according to this definition you cannot instantiate value types ever, so the last row should be "no". Perhaps it could be a "no" with a footnote, that variables can be declared.

Also, constants can be declared at the top-level.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+|-------------|-------------------------------------------------|
+|             | can be called at run time from this place in a P4 program |||||
+|             |-----------|---------|---------|------------|--------|
+|             |           | control | action  | action not |        |
+|             | parser    | apply   | used by | used by    |        |
+| This type   | state     | block   | table   | any table  | extern |
++-------------+:---------:+:-------:+:-------:+:----------:+:------:+
+| package     | N/A       | N/A     | N/A     | N/A        | N/A    |
+| parser      | yes       |  no     | no      | no         | no     |
+| control     |  no       | yes     | no      | yes        | no     |
```
mbudiu-vmw: I think that actions should not be allowed to call controls. So the fourth entry should be "no".
I understand that you are treating here actions as functions; perhaps we should rather change the language to add functions rather than carry this baggage with two kinds of actions going forward.

Note that top-level actions cannot call controls because they have no way of naming control instances; you can only call a control instance.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+| extern      | yes       | yes     | yes     | yes        | no     |
+| table       |  no       | yes     | no      | yes        | no     |
```
mbudiu-vmw: In the same vein, actions should be unable to call tables.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+parameters, with the addition that `void` can be a return type for
+methods that return no value.
+
+The next table lists restrictions on what kinds of calls can be made
+from which places in a P4 program.  Calling a parser, control, or
+table means invoking its `apply()` method.  The row for `extern`
+describes where extern method calls can be made from.
+
+|-------------|-------------------------------------------------|
+|             | can be called at run time from this place in a P4 program |||||
```
mbudiu-vmw: You could add one more column: parsers outside any state and controls outside the apply block.
This includes, for example, variable initializers.
```
control c() {
   Random() rand;
   bit<32> x = rand.get();
   apply { ... }}
```

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+It is allowed to write actions that are not used by any table.  Such
+"actions not used by any table" have exactly the same restrictions on
+what they can call as a control `apply` block does.  A potential
+convenience of creating such an action is that it can be defined
+within the control that calls it, giving the action access to all
+parameter values and local variables defined earlier within its
+enclosing control.  By contrast, control definitions cannot be nested.
+
+Note that while the extern row shows that extern methods can be called
+from many places, particular externs may have additional restrictions
```
mbudiu-vmw: These restrictions are features of the target architecture which declares the externs; the target documentation should specify these restrictions.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+not listed in this table.  Any such restrictions should be documented
+in the description for each extern.  In most cases, it is expected
+that the restriction will be "from a parser state only" or "from a
+control apply block or action only".
```
mbudiu-vmw: Note that v1model only allows checksum externs to be instantiated within the verify and update checksum blocks. So restrictions can be more complicated.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+calls that can be made.  Many of them are described earlier in this
+document, but are collected here for easy reference.
+
+The stateful types of objects in P4~16~ are packages, parsers,
+controls, externs, and tables.  All other types are referred to as
+"value types" here.
+
+Some guiding principles:
+
+- Packages can be instantiated, but have no run-time behavior, so
```
jafingerhut: Understood.  That bullet point wasn't explaining very many of the later restrictions, so I'll just delete it.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,160 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+  is no use in passing one type to the other in constructor parameters
+  or run-time parameters.
+- At run time, after a control is called, and before that call is
+  complete, there can be no recursive calls between controls, nor from
+  a control to itself.  Similarly for parsers.  There can be loops
+  among states within a single parser.
+- Externs are not allowed to call parsers or controls, so there is no
+  use in passing objects of those types to them.
+- Tables are always instantiated directly in their enclosing control,
+  and cannot be instantiated at the top level.  There is no type by
```
jafingerhut: Done.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Done.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Done.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Done.

p4-16/spec/P4-16-spec.mdk
jafingerhut: You mention variable initializers as an example.  Are there other examples you know of?

p4-16/spec/P4-16-spec.mdk
jafingerhut: Makes sense.  I've made the change by having only a single column for 'action', since as you argue there should be no distinction between what you can call from inside of an action, regardless of whether the action is called from a table or not.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Done.

p4-16/spec/P4-16-spec.mdk
jafingerhut: I've added a few words regarding this that may address your comment.  Take a look at the updated version to see.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Same as previous comment of yours -- take a look at the updated text to see if it seems clear enough.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Done.  Check the contents of the column to see if it allows everything you think ought to be.  I've only included extern calls so far.  Unless there is some place for such calls other than initializers, the other kinds of calls don't make sense since they cannot return a value.

jnfoster:
Are we classifying this is a minor change that can be merged without a discussion on p4-design? It's certainly more than a typo correction.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,153 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+| control     | yes     |  no    | yes     |  no    |
+| extern      | yes     | yes    | yes     | yes    |
+| table       |  no     |  no    |  no     |  no    |
+| value types | yes     | yes    | yes     | yes    |
+|-------------|---------|--------|---------|--------|
+
+The next table lists restrictions on where one may perform
+instantiations (see Section [#sec-instantiations]) of different types.
+The answer for `package` is always "no" because there is no
+"inside a package" where instantiations can be written in P4~16~.  One
```
cc10512: No package inside a package is a limitation that will haunt us. We briefly touched on how hard is to express architectures with multiple repeated components (e.g. multiple pipelines) and being able to express them as nested packages will greatly simplify the definition of such architectures. I realize that it is not currently used in PSA, but do we really want to fully restrict doing it in the language?

I'm thinking about the following example:
```
const N_PIPES = 8;
package Pipe (Parser p, Ingress i, Egress e, Deparser d);
package MultiPipe (package Pipes p[N_PIPES]);
```
I do read this as an instantiation within the package. It is true that it is unlikely that these will be instantiations by a user program, but rather they'll be instantiated by the architecture. However, for better or worse, we are using the _same_ language for both user programs and architecture definitions. Are we at the point where we should consider certain language features to be applicable only in architecture definition files?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,153 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+run-time parameters to other callable things that have run-time
+parameters: parsers, controls, extern methods, and actions.
+
+|-------------|-------------------------------------------------|
+|             | can be a run-time parameter to this callable thing ||||
+|             |-----------|---------|---------|---------|
+| This type   | parser    | control | method  | action  |
++-------------+:---------:+:-------:+:-------:+:-------:+
+| package     |  no       |  no     | no      | no      |
+| parser      |  no       |  no     | no      | no      |
```
cc10512: We already support calling subparsers from parsers. If we want to support modularity, we have to allow that someone writes a parser in a separate module and passes that instance to the "main" parser. I can't think of any other method by which existing written code can take a not-yet-seen parser, except if it can be passed at runtime.

jafingerhut: Agreed that we should support calling subparsers from parsers.  The contents of this table do not prevent that, and a different table explicitly says that parsers should be callable from parsers.

Not sure if you were aware, but p4c currently gives an error if you attempt to pass a control to another control as a run-time parameter, or a parser to another parser as a run-time parameter.  Short P4_16 programs you can try compiling to verify this can be found here: https://github.com/jafingerhut/p4-guide/tree/master/run-time-params

Maybe that is a bug in the current compiler, but keep reading below for evidence that maybe it is by design.

It does permit passing controls to other controls as _constructor_ parameters, and parser to other parsers as constructor parameters.  Is that perhaps sufficient to achieve the modularity you are hoping for?

See this relevant issue: https://github.com/p4lang/p4-spec/issues/364 which I opened partly in response to the following comment from @jnfoster, who believes that passing controls or parsers as run-time parameters should be illegal, because it enables one to write mutually recursive calls between them: https://github.com/p4lang/p4-spec/issues/361#issuecomment-318768783

p4-16/spec/P4-16-spec.mdk
jafingerhut: I'm not sure, but perhaps the terminology in my text is confusing the issue here?

The first table says that a package can be a constructor parameter for another package.  Isn't that what you are showing examples of?

The next table by saying "can be instantiated in this place" means where in the code you can write an instantiation statement.  An instantiation statement and constructor call are similar, but not identical, I think?

p4-16/spec/P4-16-spec.mdk
cc10512: Thanks @jafingerhut. I believe it is by design, but as you know, the current design of both language and compiler assume the availability of the entire program. In such case, yes, a constructor parameter is sufficient. However, for true modularity (and I realize this might be a P4_18 goal) we might need to have only the type declared at compile time, and the actual instantiation to happen at "link time", when we bring the modules together. That's not quite runtime, but it is likely closer to that then compile time. And yes, I agree that for the current version of P4_16 you captured the actual behavior. I'd like @jnfoster to chime in, he started looking at modularity.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6416,6 +6416,154 @@ ck16.update( { h.ipv4.ttl, h.ipv4.proto } );
...
+| control     |  no       |  no     | no      | no      |
+| extern      | yes       | yes     | yes     | no      |
+| table       |  no       |  no     | no      | no      |
+| value types | yes       | yes     | yes     | yes     |
+|-------------|-----------|---------|---------|---------|
+
+The restrictions on return types of extern methods is the same as for
+parameters, with the addition that `void` can be a return type for
+methods that return no value.
```
jafingerhut: The table above and this sentence imply that it should be allowed to return an extern from an extern method call.  I'm not sure that is intended to be allowed?

If not, this sentence should probably be replaced with: "Extern method calls may return only value types, or nothing (specified by a return type of `void`)."

mbudiu-vmw: Yes, let's start with the most restrictive language possible, and make it more lenient if there is a compelling use case. I think that functions should not be allowed to return extern types.

jafingerhut: Done.

jnfoster: Do we want to distinguish constructor parameters from apply parameters here?

jafingerhut: I think we do, and I've tried to do so with these changes, but please suggest ways to make it clearer, if it is not.

There are currently 4 tables in these additions.  They are, in order:

(a) restrictions on constructor parameter types.  The table headings are "This type..." and "can be a constructor parameter for this type"

(b) restrictions on where in the source code instantiations are allowed.  The table headings are "This type" and "can be instantiated in this place"

(c) restrictions on run-time parameter types, which is what I think you mean by "apply parameters"?  The headings are "This type" and "can be a run-time parameter to this callable thing"

(d) restrictions on where calls to different kinds of callable things may occur.  The headings are "This type" and "can be called at run time from this place in a P4 program"

jnfoster: Sorry, my fault. I was looking at GitHub diffs and missed it. Let me rebuild and check closely.

PR #373, opened 2017-08-05
Draft proposal for clone/resubmit/recirculate

hanw:
Based on the discussion in #369, here is a draft proposal for clone/resubmit/recirculate under the assumption that we will add ingress deparser and egress parser to the PSA architecture.

Issue #370, opened 2017-08-03
[PSA] Default Port Id type


samar-abdi:
The current default port ID type is a 10 bit bitstring. Is there a strong reason for this? It would be preferable to have a default 32 bit port ID type for couple of reasons:
1. SDN apps that need to be ported from the Openflow version to an equivalent P4 version already use a uint32 to define port numbers.
2. P4Runtime defines uint32 port ids for packet IO. 

Of course, one could override the Port ID type, but that would be needed across all P4 programs and seems like an unnecessary cognitive load/ potential for errors.

jafingerhut:
Do you mean that you would prefer it if the psa.p4 file that one #include'd in P4 programs defined a port ID type to be the same as bit<32>, regardless of the target device?

If so, I believe the motivation for allowing it to be significantly smaller in P4 programs is that bits of storage space will be at a premium in many such implementations.  If one had a 32-bit port ID as an action parameter for a table with a million entries, for example, and all of those port ID values were in the range 0 to 63 for that device, then you are wasting 58 Mbits ~= 7 Mbytes of relatively small on-chip memory in an ASIC or FPGA that achieves the highest packet rates.  Don't compare that 7 Mbytes to general purpose CPU DRAM sizes (typically tens or hundreds of gigabytes) -- compare it it on-chip on CPUs with typically several tens of Mbytes, maybe low hundreds of Mbytes total.  That 7Mbytes would be filled with all 0s all of the time.

Are you concerned that the control plane will need to trim 32-bit port ids down to platform-specific bit widths in order to create P4Runtime API calls?

samar-abdi:
While this is a valid concern, I am not aware of direct tables with actions that have a port ID parameter. Port IDs are typically used in L3 routing actions, and they are best modeled as indirect tables, so we will end up with only few tens of action entries even if there are a million flows.

By the same token, if an action takes src and dst mac as parameters, a direct table with a million entries will require ~96MB for storing actions alone!

The problem is in control plane portability as you say in your last comment. Since P4Runtime uses uint32 (in network byte order) for port ID, a P4Runtime server on the switch will have to interpret and map that uint32 value to a 10-bit value. And different switches may do so differently.

jafingerhut:
PortID is only one of several fields proposed in the PSA right now to be a type bit<W> where the width in bits W can differ across PSA implementations.

[...]

PR #384, opened 2017-08-15
Proposed PSA pseudocode for how metadata controls where packets go ...

jafingerhut:
... after ingress and egress control block are complete.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -124,11 +124,8 @@ part of the PSA pipeline.
 
 Even though the PRE can not be programmed using P4, it can be
 configured both directly using control plane APIs and by setting
-intrinsic metadata. In this specification we opt to define the
-operations available in the PRE as method invocations. A target
-backend is responsible for mapping the PRE extern APIs to the
-appropriate mechanisms for performing these operations in the
-hardware.
+intrinsic metadata, and the `psa.p4` include file provides some
+actions to help set these metadata fields for some common use cases.
 
 The PRE extern object has no constructor, and thus it cannot be
 instantiated in the user's P4 program. The architecture instantiates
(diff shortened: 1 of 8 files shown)


p4-16/psa/PSA.mdk
```diff
@@ -146,49 +143,105 @@ of native and non-native operations. Target documentation should highlight such
...
+    }
+    // Continue below, regardless of whether a clone was created or not.
+    if (ostd.drop) {
+        drop the packet
+        return;   // Do not continue below.
+    }
+    if (ostd.resubmit) {
+        resubmit the packet, i.e. it will go back to starting with the
+            ingress parser;
+        // TBD: Specify if anything is different about the resubmitted
```
cc10512: Is the egress_port for the resubmitted packet assumed to have been set already?

p4-16/psa/PSA.mdk
```diff
@@ -146,49 +143,105 @@ of native and non-native operations. Target documentation should highlight such
...
+[INCLUDE=psa.p4:Metadata_ingress_output]
+
+    psa_ingress_output_metadata_t ostd;
 
-The semantics of calling the PRE APIs is equivalent to setting
-intrinsic metadata fields/bits and assuming that the PRE looks up the
-fields in the following order: drop, truncate, multicast, clone,
-output_port.
+    if (ostd.clone) {
+        create a copy of the packet to the clone target
```
cc10512: what is the `egress_port` of the cloned packet? can it be set here or is it assumed to be the same as the original packet?

cc10512:
Also, it would be good to address @samar-abdi's comment here: https://github.com/p4lang/p4-spec/issues/370#issuecomment-322272859

p4-16/psa/PSA.mdk
jafingerhut: Good questions, and not the only ones I have about how all of this works, as is hopefully obvious from the TBD comments I have in the text.  I'm happy to add more TBD questions to this pull request, unless maybe @hanw has suggested answers he wants me to incorporate in this PR.

My hope in this PR was to at least first get a start at what the possibilities are for controlling what happens to a packet after it is finished with ingress or egress processing, the metadata fields that control it, and the pseudocode that defines at least the relative priorities of these actions, e.g. clone vs. drop vs. resubmit.

p4-16/psa/PSA.mdk
jafingerhut: Ditto to my previous reply above -- I don't have an answer yet, and am hoping @hanw has some thoughts on the matter, but not sure whether they need to be in this PR, or a later one that adds more detail after this one is merged.

p4-16/psa/PSA.mdk
hanw: We need additional metadata to control the `egress_port` for the cloned packet. The cloned packet should not shared the same `egress_port` metadata with the original packet. I added the metadata `clone_spec` to the ingress_metadata/egress_metadata for PRE in PR #373

hanw: A common use case for cloning is to send the packet to control CPU. In this case, the `clone_spec` should be set to a value that represents the control CPU port.

p4-16/psa/PSA.mdk
hanw: My understanding is the resubmitted packet is coming into the ingress pipeline from the same port. There is no need to set the `egress_port` on the original packet because it is resubmitted for processing.

p4-16/psa/PSA.mdk
jafingerhut: Planning to merge this in, and address these comments in the merged version, hopefully before the next PSA working group meeting.

PR #389, opened 2017-08-20
Psa fix recirc behavior

jafingerhut:


jafingerhut:
@hanw I am making a change to the psuedocode for recirculate, similar to the one you made recently for resubmit.  At least from the text description, neither of them make an additional copy of a packet.

PR #390, opened 2017-08-20
Proposed changes to CloneSpec_t

hanw:
- Remove CloneMethod_t from psa.p4
- Added CloneType_t to specify if cloned packet is modified version or the original packet.
- Added CPU as a field to CloneSpec_t enum.


hanw:
superseded by #393

PR #397, opened 2017-08-23
Proposal for updating PSA_Switch package and controls in psa.p4

jafingerhut:


jafingerhut:
@hanw @vgurevich Could you look this over and see if it makes sense?

It updates the package and control definitions near the end of psa.p4 as follows:

+ remove verifyChecksum control, in preparation for the checksum calculations being done in the ingress parser instead (and similar checks could be done in the egress parser, too, if one wished)

+ It keeps computeChecksum after egress control block, before the egress deparser.  We can fold it into egress processing if we want, I suppose.  I don't have a strong opinion either way, but perhaps you do.

+ It adds explicit ingress deparser, and egress parser.

+ It makes two slightly different signatures for ingress and egress parsers -- the only difference is the input metadata each receives.  Ingress parser gets ingress_port as a field of its input metadata, whereas egress parser gets egress_port instead.

I have updated one of the example P4_16 programs that use psa.p4 (psa-example-counters.p4), and it compiles with this new psa.p4 file.  If these changes look reasonable, I can update the other P4_16 programs in the p4-16/psa/examples/ directory similarly.

jafingerhut:
@hanw There is no egress_port from the ingress_deparser.  The only output from the ingress deparser is a packet out.

Do you mean the egress_port from the ingress control block?  If the ingress control blocks sets metadata in such a way that only 1 normal packet goes to egress processing, then the egress_port value there should be exactly the same as the egress_port value sent to the egress deparser, which is the same as the egress_port value sent to the egress control block, yes?  Sure, we need some way (maybe tables, maybe text, maybe pseudocode) to specify exactly how these metadata fields are initialized before the egress parser and egress control blocks start executing, not just for the normal case, but for the clone cases as well.

mbudiu-vmw:
You should make sure that BMv2 will support relatively soon checksum units in the parser.
BTW: I would not finalize this architecture until there is at least a sketch of a compiler that generates code for a BMv2 version of it.

jafingerhut:
@mbudiu-vmw Nothing being merged into the PSA spec now is final.  I'm trying to at least move forwards to a PSA spec _draft_ that has all of the proposed pieces, that is at least close to internally consistent with itself.  Agreed that we want at least an early implementation of a compiler that works with whatever psa.p4 we end up with.

p4-16/psa/psa.p4
```diff
@@ -568,12 +579,13 @@ control ComputeChecksum<H, M>(inout H hdr, inout M user_meta);
...
-package PSA_Switch<H, M>(Parser<H, M> p,
-                         VerifyChecksum<H, M> vr,
+package PSA_Switch<H, M>(IngressParser<H, M> ip,
                          Ingress<H, M> ig,
+                         Deparser<H> igdep,
+                         EgressParser<H, M> ep,
                          Egress<H, M> eg,
                          ComputeChecksum<H, M> ck,
-                         Deparser<H> dep);
+                         Deparser<H> egdep);
```
vgurevich: This is a long discussion that we should probably have in-person. Here are some things to think about:
1. Should H and M really be the same at ingress and egress? 
2. Are we saying that checksum is computed only at the egress or that it is instantiated/called twice? Perhaps, that's why it might be better to fold it into deparsing.

jafingerhut: 1. I'm happy making separate H and M for ingress and egress.  IH, IM, EH, EM.
2. Please pick something that you consider implementable, and please let's write that into the draft.  So far, I thought that the thinking was that deparser code was going to be restricted to nothing but emit() method calls, but I can't think of any objection to allowing checksum calculations in there as well.  Neither can I think of any objection to allowing checksum calculations in Ingress and/or Egress control blocks.

vgurevich: I would think that if you have a deparser, checksum (re)calculation should be nearby. 

Having it in a separate control is easiest.
Having it in a deparser is generally better than in ingress/egress control because we need to ensure that the fields do not change after we calculate the checksum. Imposing this restriction inside a general ingress/egress control would be difficult, but might be OK in the deparser (basically, we'd allow one extra extern there in addition to packet_out and other _outs).

jafingerhut: If checksum calculation and fixing up were allowed in ingress and egress, I don't see a reason why the compiler would need to impose any restrictions on where it was done.  If the programmer wanted it to be after every other modification, they could simply make it occur last sequentially in their ingress/egress apply block.

That said, I have no problems with allowing checksum fixup calls in the deparsers.  That will require making the headers 'inout' direction, else the header fields cannot be modified, but that is easy to do.

jafingerhut: I've added another commit that eliminate the computeChecksum control block, and makes the headers have direction 'inout' in the Deparser control blocks, to enable header fields to be modified in those controls.

vgurevich: My point was that if the programmer decides to put checksum calculation anywhere but at the very end of the ingress/egress control, then we will have a problem. Checking for "the very end" whatever it might mean will be painful and we do not want to express it in the language.

I think it is much easier to say (and check) the rule "this extern should be used in that control only".

I think you are making an important point that recalculating checksums in the deparser requires the headers to become inout. Perhaps, that's a sign that we should have it separate and keep the deparser clean (i.e. do not allow it to modify headers) in that regard?

jafingerhut: If a programmer decides to put checksum calculation before the last place that they modify the relevant headers, then the programmer has created a problem, and must live with their mistake.  If we allow checksum calculation in a deparser, we are saying "do checksum fixups before the emit calls, or it will be incorrect."  That seems roughly the same as allowing it in ingress/egress, and saying "do checksum fixups after all header modifications, or it will be incorrect."

I've added computeChecksum back in, but this time both for ingress and egress.  Eventually we will stop changing this.

vgurevich: Coolz! Thanks, Andy!

p4-16/psa/psa.p4
```diff
@@ -568,12 +579,14 @@ control ComputeChecksum<H, M>(inout H hdr, inout M user_meta);
...
-                         Egress<H, M> eg,
-                         ComputeChecksum<H, M> ck,
-                         Deparser<H> dep);
+package PSA_Switch<IH, IM, EH, EM>(IngressParser<IH, IM> ip,
+                                   Ingress<IH, IM> ig,
+                                   ComputeChecksum<IH, IM> ic,
+                                   Deparser<IH> id,
+                                   EgressParser<EH, EM> ep,
+                                   Egress<EH, EM> eg,
+                                   ComputeChecksum<EH, EM> ec,
```
vgurevich: Is it really OK to have the same name (Compute Checksum and Deparser) twice?

jafingerhut:
I could not find this comment on the Github site to reply to, only in my
email, so replying via email.

The ComputeChecksum<IH, IM> and ComputeChecksum<EH, EM> are names of types,
not names of parameters.  The names of the parameters are unique: "ic" and
"ec".

I don't think it is any more unusual than having two parameters "ic" and
"ec" both with type "bit<16>" to an action.

All of the code in psa.p4 and the all of the example programs that #include
it mentioned in the "check" target in p4-spec/p4-16/psa/Makefile are
compiled by me using a recent version of p4test, and it has no problem with
this code at all (except warnings that some fields of the struct ostd may
not be initialized, but that is expected right now).

Andy

vgurevich:
@jafingerhut 

It was late at night and I realized my mistake a second after I put the comment in and therefore deleted it. :) You are absolutely correct.

PR #398, opened 2017-08-24
Proposed restrictions on where externs may be used in PSA programs

jafingerhut:


p4-16/psa/PSA.mdk
```diff
@@ -120,6 +120,81 @@ Additional supported match_kind types
 
 # PSA Externs
 
+## Restrictions on where externs may be used { #sec-extern-restrictions }
+
+All instantiations in a P4~16~ program occur at compile time, and can
+be arranged in a tree structure we will call the instantiation tree.
+The root of the tree `T` represents the top level of the program.  Its
+children are the node for the package `PSA_Switch` described in
+[#sec-programmable-blocks], plus any externs instantiated at the top
```
cc10512: nit: s/plus/and

p4-16/psa/PSA.mdk
```diff
@@ -120,6 +120,81 @@ Additional supported match_kind types
...
+level of the program.  The children of the `PSA_Switch` node are the
+parsers and controls passed as parameters to the `PSA_Switch`
+instantiation.  If any of those parsers or controls instantiate other
+parsers, controls, and/or externs, create child nodes for them,
```
cc10512: "create child nodes for them" can be interpreted as asking the P4 programmer to create the nodes. We should says something like "the compiler creates nodes for them as children of the instantiating nodes" (or something to that effect).

p4-16/psa/PSA.mdk
```diff
@@ -120,6 +120,81 @@ Additional supported match_kind types
...
+
+A PSA implementation is allowed to reject programs that instantiate
+externs, or attempt to call their methods, from anywhere other than
+the places mentioned in the table below.
+
+For example, `Counter` being restricted to "Ingress, Egress" means
+that every `Counter` instance must be instantiated within either the
+Ingress control block or the Egress control block, or be a descendant
+of one of those nodes in the instantiation tree.  If a `Counter`
+instance is instantiated in Ingress, for example, then it cannot be
```
cc10512: s/then it can not be named/then it can not be referenced/ ?

p4-16/psa/PSA.mdk
```diff
@@ -120,6 +120,81 @@ Additional supported match_kind types
...
+named, and thus its methods cannot be called, from any non-Ingress
+control block.
+
+|-------------------|----------------------------------------------|
+| Extern type       | Where it may be instantiated and called from |
++:------------------+:---------------------------------------------+
+| `ActionProfile`   | Ingress, Egress |
+| `ActionSelector`  | Ingress, Egress |
+| `Checksum`        | IngressParser, EgressParser, Deparser |
```
cc10512: should we split checksum verification and update? they have different scopes.

jafingerhut: At least with the current Checksum method calls defined in the latest PSA draft, both checksum verification and update rely on a get() method call.

I understand that the collection of Checksum method calls might be changing soon.  I don't know what the final version will look like, but am happy to update this table when that becomes clearer.

p4-16/psa/PSA.mdk
jafingerhut: Done.

p4-16/psa/PSA.mdk
jafingerhut: I am changing it to "the instantiation tree contains child nodes for them".

p4-16/psa/PSA.mdk
jafingerhut: Done.

PR #400, opened 2017-08-25
Add example of using Checksum extern to detect wrong IPv4 header chec…

jafingerhut:
…ksum

The same example also demonstrates:

+ How to use the parser error that is passed into the Ingress control
  block.
+ Example action that converts the 'error' type to a bit vector, so
  the resulting bit vector can be used for other purposes, e.g. an
  index into a Counter array.

### DIFF
--- a/p4-16/psa/examples/psa-example-parser-checksum.p4
+++ b/p4-16/psa/examples/psa-example-parser-checksum.p4
@@ -1,0 +1,232 @@
+/*
+Copyright 2017 Cisco Systems, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <core.p4>
+#include "../psa.p4"
+
+
+typedef bit<48>  EthernetAddress;
+
+header ethernet_t {
+    EthernetAddress dstAddr;
+    EthernetAddress srcAddr;
+    bit<16>         etherType;
+}
+
+header ipv4_t {
+    bit<4>  version;
+    bit<4>  ihl;
+    bit<8>  diffserv;
+    bit<16> totalLen;
+    bit<16> identification;
+    bit<3>  flags;
+    bit<13> fragOffset;
+    bit<8>  ttl;
+    bit<8>  protocol;
+    bit<16> hdrChecksum;
+    bit<32> srcAddr;
+    bit<32> dstAddr;
+}
+
+header tcp_t {
+    bit<16> srcPort;
+    bit<16> dstPort;
+    bit<32> seqNo;
+    bit<32> ackNo;
+    bit<4>  dataOffset;
+    bit<3>  res;
+    bit<3>  ecn;
+    bit<6>  ctrl;
+    bit<16> window;
+    bit<16> checksum;
+    bit<16> urgentPtr;
+}
+
+struct fwd_metadata_t {
+}
+
+struct metadata {
+    fwd_metadata_t fwd_metadata;
+}
+
+struct headers {
+    ethernet_t       ethernet;
+    ipv4_t           ipv4;
+    tcp_t            tcp;
+}
+
+
+// BEGIN:Parse_Error_Example
+// Define additional error values, one of them for packets with
+// incorrect IPv4 header checksums.
+error {
+    UnhandledIPv4Options,
+    BadIPv4HeaderChecksum
+}
+
+typedef bit<32> PacketCounter_t;
+typedef bit<8>  ErrorIndex_t;
+
+const bit<9> NUM_ERRORS = 256;
+
+// Action convert_error_to_idx returns a unique ErrorIndex_t value (a
+// bit vector) for each possible value of the error type.  The error
+// type cannot be directly put into packet headers or used as an index
+// into counter, meter, or register arrays, so such a conversion
+// 'function' is necessary to get a value that can be used for these
+// purposes.
+
+action convert_error_to_idx (in ParserError_t err,
+                             out ErrorIndex_t idx)
+{
+    if (err == error.NoError) {
+        idx = 1;
+    } else if (err == error.PacketTooShort) {
+        idx = 2;
+    } else if (err == error.NoMatch) {
+        idx = 3;
+    } else if (err == error.StackOutOfBounds) {
+        idx = 4;
+    } else if (err == error.HeaderTooShort) {
+        idx = 5;
+    } else if (err == error.ParserTimeout) {
+        idx = 6;
+    } else if (err == error.BadIPv4HeaderChecksum) {
+        idx = 7;
+    } else if (err == error.UnhandledIPv4Options) {
+        idx = 8;
+    } else {
+        // Unknown error value
+        idx = 0;
+    }
+}
+
+parser IngressParserImpl(packet_in buffer,
+                         out headers parsed_hdr,
+                         inout metadata user_meta,
+                         in psa_ingress_parser_input_metadata_t istd,
+                         out psa_parser_output_metadata_t ostd)
+{
+    Checksum<bit<16>>(HashAlgorithm.ones_complement16) ck;
+    state start {
+        buffer.extract(parsed_hdr.ethernet);
+        transition select(parsed_hdr.ethernet.etherType) {
+            0x0800: parse_ipv4;
+            default: accept;
+        }
+    }
+    state parse_ipv4 {
+        buffer.extract(parsed_hdr.ipv4);
+        // TBD: It would be good to enhance this example to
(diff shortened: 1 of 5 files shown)


jafingerhut:
@jnfoster @vgurevich @hanw Please look over this example code and see if there are any obvious problems with it.

I fully expect that the API to the Checksum extern will be changed soon, and it would be good to update the small part of this example that uses that extern when that happens.  This sample program compiles cleanly with the latest p4test and checked-in psa.p4 file (except for warnings about fields of the ostd struct not being initialized, which is normal).  I would prefer to check this in and then update the Checksum extern method calls with a later PR, unless someone already has another PR ready to go for that.

The rest of the example demonstrates simpler parser error handling in the ingress control block, following some code that @vgurevich has in another Github issue.

p4-16/psa/examples/psa-example-parser-checksum.p4
```diff
@@ -0,0 +1,232 @@
...
+                parsed_hdr.ipv4.protocol,
+                parsed_hdr.ipv4.hdrChecksum,
+                parsed_hdr.ipv4.srcAddr,
+                parsed_hdr.ipv4.dstAddr });
+        // The verify statement below will cause the parser to enter
+        // the reject state, and thus terminate parsing immediately,
+        // if the IPv4 header checksum is wrong.  It will also record
+        // the error error.BadIPv4HeaderChecksum, which will be
+        // available in a metadata field in the ingress control block.
+        verify(ck.get() == 0, error.BadIPv4HeaderChecksum);
```
vgurevich: Thanks, this is pretty much what I had in mind. However, there is a little difference with the code I've written and I am just wondering if the compiler will be able to do the right thing in both cases. I hope it is possible.

My code looked like this:

```
ck.update({ parsed_hdr.ipv4.version,
                parsed_hdr.ipv4.ihl,
                parsed_hdr.ipv4.diffserv,
                parsed_hdr.ipv4.totalLen,
                parsed_hdr.ipv4.identification,
                parsed_hdr.ipv4.flags,
                parsed_hdr.ipv4.fragOffset,
                parsed_hdr.ipv4.ttl,
                parsed_hdr.ipv4.protocol,
                // parsed_hdr.ipv4.hdrChecksum,
                parsed_hdr.ipv4.srcAddr,
                parsed_hdr.ipv4.dstAddr });
        // The verify statement below will cause the parser to enter
        // the reject state, and thus terminate parsing immediately,
        // if the IPv4 header checksum is wrong.  It will also record
        // the error error.BadIPv4HeaderChecksum, which will be
        // available in a metadata field in the ingress control block.
        verify(ck.get() == parsed_hdr.ipv4.hdrChecksum, error.BadIPv4HeaderChecksum);
```

p4-16/psa/examples/psa-example-parser-checksum.p4
```diff
@@ -0,0 +1,232 @@
...
+// bit vector) for each possible value of the error type.  The error
+// type cannot be directly put into packet headers or used as an index
+// into counter, meter, or register arrays, so such a conversion
+// 'function' is necessary to get a value that can be used for these
+// purposes.
+
+action convert_error_to_idx (in ParserError_t err,
+                             out ErrorIndex_t idx)
+{
+    if (err == error.NoError) {
```
vgurevich: Why not an exact match table instead?

p4-16/psa/examples/psa-example-parser-checksum.p4
```diff
@@ -0,0 +1,232 @@
...
+    Counter<PacketCounter_t, ErrorIndex_t>
+        ((bit<32>) NUM_ERRORS, CounterType_t.packets) parser_error_counts;
+    apply {
+        if (istd.parser_error != error.NoError) {
+            // Example code showing how to count number of times each
+            // kind of parser error was seen.
+            ErrorIndex_t error_idx;
+            convert_error_to_idx(istd.parser_error, error_idx);
+            parser_error_counts.count(error_idx);
+            ingress_drop(ostd);
```
vgurevich: This is probably easier to do with an exact match table and a direct counter...

jafingerhut: Makes sense, and I can change the example to do that.  It simply didn't occur to me while writing the example.  @antoninbas The P4Runtime API should be able to handle exact matches on error type, then?

p4-16/psa/examples/psa-example-parser-checksum.p4
jafingerhut: I would expect that a checksum extern API could be implemented to work in the way that you show, and can change the example to demonstrate it that way.

p4-16/psa/examples/psa-example-parser-checksum.p4
jafingerhut: Example changed to use exact match table and direct counter.

p4-16/psa/examples/psa-example-parser-checksum.p4
jafingerhut: Example code changed to leave ipv4.hdrChecksum out of checksum calculation, instead using it to compare output of checksum calculation against that value.

p4-16/psa/examples/psa-example-parser-checksum.p4
jafingerhut: Done.

p4-16/psa/examples/psa-example-parser-checksum.p4
```diff
@@ -0,0 +1,234 @@
...
+// incorrect IPv4 header checksums.
+error {
+    UnhandledIPv4Options,
+    BadIPv4HeaderChecksum
+}
+
+typedef bit<32> PacketCounter_t;
+typedef bit<8>  ErrorIndex_t;
+
+const bit<9> NUM_ERRORS = 256;
```
vgurevich: BTW, this is another thing we'll need to add to the "how many" list. I'd suspect than in practice it will be much shorter, probably 16 or 32, even with user-defined errors

p4-16/psa/examples/psa-example-parser-checksum.p4
```diff
@@ -0,0 +1,234 @@
...
+    }
+    table parser_error_count_and_convert {
+        key = {
+            istd.parser_error : exact;
+        }
+        actions = {
+            set_error_idx;
+        }
+        default_action = set_error_idx(0);
+        const entries = {
```
vgurevich: Note that error_idx is not used anywhere anymore, but that's OK -- it serves as an example of how to go from an enum to an integer. :)

jafingerhut: Agreed, and a comment above this even mentions this explicitly, and explains that it is in the example solely to show an example of how to do this.  If you wanted to put an error index into a header of a packet to the CPU indicating what parser error was encountered, for example.

p4-16/psa/examples/psa-example-parser-checksum.p4
jafingerhut: This value could be different from one P4 program to another -- there is nothing preventing anyone from creating as many different values of type 'error' as they wish in their P4 programs.  I can't think of a strong need to address this topic in the PSA spec, unless someone thinks of a good reason to do so.

jafingerhut: Oh, if you mean into one of the proposed "PSA profiles" that contain proposed numerical values for "PSA compliant implementations should support at least this many of thing X", then yes, good point.

vgurevich: Yep, that's what I meant.

jafingerhut:
It looks like hopefully all of @vgurevich's comments have been addressed.  @jnfoster @hanw @cc10512 any comments before we merge this in?  As I mention above, I expect that the API to the checksum extern will likely change soon, and I can make whatever changes are needed to this example code when we document what that is in the PSA.  I'd rather 'merge early, merge often', at least until we think we are really close to a releasable v1.0 of the PSA spec.

jnfoster:
I just looked through the code, which looks fine to me. 

A few questions?
* What did we decide about incremental vs. batch-mode? I don't see a discussion of that here.
* Why is computeChecksum instantiated twice, once before ingress and once after ingress? If the parser verifies the checksum, isn't the computation redundant?

jafingerhut:
I didn't put a discussion of incremental vs. batch-mode here yet, as my main purpose so far was to show an example of checking an IPv4 header checksum in the parser, and detecting that error in the ingress code.  I am hoping that any discussion of incremental vs. batch-mode could be made in a later PR, "on top of" this one.

Regarding having two computeChecksum blocks, one on ingress and one on egress, that change was made with this recent PR: https://github.com/p4lang/p4-spec/pull/397 as suggested by @vgurevich, I believe primarily for symmetry when there is a separate ingress and egress deparser.  They need not be instantiated as the same control type, of course, but they are in this small example program.  That part of the complete P4 program is not shown in the example excerpt, since my intent in that example was to show how to use the checksum extern to calculate a fresh IPv4 header checksum from scratch.

jnfoster:
Got it. Thanks!

jafingerhut:
@vgurevich @jnfoster If you are OK with these changes (not as final text, but at least as an improvement on the current document, upon which to base future document edits), could one of you "Approve" the PR on Github, and then I will merge it in.  Github doesn't require it in order for me to merge, but I would prefer to have it as kind of a checkbox item before I merge.  Thanks.

vgurevich:
Thanks a lot, Andy!

PR #401, opened 2017-08-30
Minor syntax fixes for examples with psa_direct_counters table proper…

jafingerhut:
…ties


jafingerhut:
Went ahead and merged this in, as it is a minor typo fix.

PR #403, opened 2017-08-30
Proposed PSA mechanism for ingress control to choose class-of-service…

jafingerhut:
… queue

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -252,22 +252,26 @@ the contents of several metadata fields in the struct
     psa_ingress_output_metadata_t ostd;
 
     if (ostd.class_of_service value is not supported) {
-        drop the packet;
-        Optional: Record error status, e.g. update a count of packets
-            dropped for this reason, and record the bad value of
-            ostd.class_of_service in a small queue, dropping a new
-            value if the queue is full.
-        return;   // Do not continue below.
+        // use default class
+        ostd.class_of_service = 0;
+        Optional: Count occurrences and first few unsupported values
+            of ostd.class_of_service.
     }
     if (truncate) {
         Truncate the payload to at most truncate_payload_bytes long.
             This affects any copies made below.
     }
     if (ostd.clone) {
-        create a copy of the packet and send it to the packet buffer,
-            after which it will start egress processing with
-            egress_port=ostd.clone_port (which might be PORT_CPU to
-            send the clone to the control CPU).
+        if (ostd.clone_class_of_service value is not supported) {
+            ostd.clone_class_of_service = 0;
+            Optional: Count occurrences and first few unsupported
+                values of ostd.clone_class_of_service.
+	}
+        create a copy of the packet and send it to the packet buffer
+            with class of service specified by
+            ostd.clone_class_of_service, after which it will start
+            egress processing with egress_port=ostd.clone_port (which
+            might be PORT_CPU to send the clone to the control CPU).
         The contents of the packet will be the same as the packet as
            it began ingress parsing, prepended with any headers
            created by clone.emit() calls in the ingress deparser apply
@@ -293,7 +297,6 @@ the contents of several metadata fields in the struct
     }
     enqueue one packet for output port ostd.egress_port with class of
         service specified by ostd.class_of_service
-
 ```
 
 TBD: Need text defining, for each possible copy, exactly what the
@@ -395,10 +398,16 @@ the contents of several metadata fields in the struct
             This affects any copies made below.
     }
     if (ostd.clone) {
-        create a copy of the packet and send it to the packet buffer,
-            after which it will start egress processing with
-            egress_port=ostd.clone_port (which might be PORT_CPU to
-            send the clone to the control CPU).
+        if (ostd.clone_class_of_service value is not supported) {
+            ostd.clone_class_of_service = 0;
+            Optional: Count occurrences and first few unsupported
+                values of ostd.clone_class_of_service.
+	}
+        create a copy of the packet and send it to the packet buffer
+            with class of service specified by
+            ostd.clone_class_of_service, after which it will start
+            egress processing with egress_port=ostd.clone_port (which
+            might be PORT_CPU to send the clone to the control CPU).
         The contents of the packet will be the same as the packet as
            it comes out of the egress deparser, prepended with any
            headers created by clone.emit() calls in the egress
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -358,6 +358,11 @@ by the Ingress control block.
 The control plane API excerpt below is intended to be added as part of
 the P4Runtime API[^P4RuntimeAPI].
 
+TBD: Antonin Bas suggests exposing 'switch capabilities' type of
+information like that shown below not via the P4Runtime API, but
+instead via Openconfig data models.  Ask him for a more appropriate
+way to document this in the PSA spec.
+
 ```
 // The ClassOfServiceInfo message should be added to the "oneof"
 // inside of message "Entity".
@@ -1447,6 +1452,11 @@ implementations were allowed to do this with their timestamp values?
 
 The control plane API excerpt below is intended to be added as part of
 the P4Runtime API[^P4RuntimeAPI].
+
+TBD: Antonin Bas suggests exposing 'switch capabilities' type of
+information like that shown below not via the P4Runtime API, but
+instead via Openconfig data models.  Ask him for a more appropriate
+way to document this in the PSA spec.
 
 ```
 // The TimestampInfo and Timestamp messages should be added to the
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -251,6 +251,14 @@ the contents of several metadata fields in the struct
 
     psa_ingress_output_metadata_t ostd;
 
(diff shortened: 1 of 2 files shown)


p4-16/psa/PSA.mdk
```diff
@@ -199,6 +202,41 @@ copy going to the control CPU to have a software-defined header
 (defined in the P4 program) that is different than any headers on the
 packet copies going to the Egress control block.
 
+A PSA implementation may choose to implement multiple class-of-service
+queues for each output port, but it may also choose to implement only
+1 queue for each output port.  The Ingress control block may choose to
```
vgurevich: Are mandating VOQ here as a principle? I do not think that's the intention. Perhaps we can add "or use any other queuing arrangement".

p4-16/psa/PSA.mdk
```diff
@@ -199,6 +202,41 @@ copy going to the control CPU to have a software-defined header
...
+assign a value to the `ostd.class_of_service` field to change the
+packet's class of service to a value other than the default of 0.
+
+PSA only specifies how the Ingress control block can choose among
+multiple class-of-service queues for an output port, if more than one
```
vgurevich: I would remove "for an output port" and replace "that might exist on the device".

p4-16/psa/PSA.mdk
```diff
@@ -199,6 +202,41 @@ copy going to the control CPU to have a software-defined header
...
+exists.  It does not mandate a scheduling policy among these queues.
+Something at least as flexible as weighted fair queuing, with an
```
vgurevich: I would remove this sentence. 

I would add as a note, though, that queueing and scheduling have the potential (if not outright intended) to reorder packets. This can have visible effect on the P4 programs, especially those containing stateful elements, writable by the data plane (counters, meters, registers). Any tests written for such systems must account for such differences in behavior otherwise they will not be portable.

p4-16/psa/PSA.mdk
```diff
@@ -180,9 +180,12 @@ the contents of several metadata fields in the struct
     if (ostd.multicast_group != 0) {
         Make 0 or more copies of the packet according to the control
             plane configuration of multicast group ostd.multicast_group.
+            Every copy will have the same value of ostd.class_of_service
         return;   // Do not continue below.
     }
-    enqueue one packet for output port ostd.egress_port
+    enqueue one packet for output port ostd.egress_port with class of
+        service specified by ostd.class_of_service
```
vgurevich: I think it might be useful to add what should happen if a given class of service doesn't exist on a particular target. In other words, I have a program that send packet with class-of-service 7 and try to run this program on a system which has only one queue....

Option 1: If class-of-service doesn't exist, drop the packet
Option 2: If class-of-service doesn't exist, use one of the existing classes as per platform's choice.

p4-16/psa/PSA.mdk
jafingerhut: Are you thinking that a typical PSA implementation would have 1 separate FIFO queue per output port, or less than that?  If less than that, you can have significant head of line blocking for traffic leading to individual output ports.

If we take 1 FIFO queue per output port as a minimum, then I was thinking that if a PSA implementation supported 4 class of service queues, it would be 4 class of service queues per output port.  Anything less and you are introducing head of line blocking issues.

But I could certainly be bringing in unwarranted assumptions here that are best gotten out in the open before finishing the PSA spec.

p4-16/psa/PSA.mdk
jafingerhut: If it isn't outright intended to reorder packets, then there isn't much point to having multiple classes of service, is there?  The only reason to introduce them that I can think of _other_ than to explicitly allow reordering between different classes of service, is if the class of service affected some drop policy in the PRE.

p4-16/psa/PSA.mdk
vgurevich: As an example, there can be a separate queue or queues for multicast packets, etc, but ultimately, it is not relevant to this very simple discussion that we are allowing to have this field that and that's pretty much the only thing I would say. We are not going to define its semantics further.

p4-16/psa/PSA.mdk
vgurevich: I think we agree, that packets will have a god chance of being reordered.

PSA allows users to write programs that are sensitive to packet ordering (mostly via registers). Therefore, it might make sense to warn people, but, I might be wrong on that point.

vgurevich: Actually, after waking up in the morning I realized that even with a single queue the order of the packet on the egress can still be different (e.g. when packets from multiple ports go to the same port you never know the exact order), so, perhaps, this is so fundamental that no additional warnings are necessary.

p4-16/psa/PSA.mdk
```diff
@@ -199,6 +202,41 @@ copy going to the control CPU to have a software-defined header
...
+
+```
+// The ClassOfServiceInfo message should be added to the "oneof"
+// inside of message "Entity".
+
+// ClassOfServiceInfo is only intended to be read.  Attempts to update
+// this entity have no effect, and should return an error status that
+// the entity is read only.
+
+message ClassOfServiceInfo {
```
antoninbas: If this is just to retrieve switch capabilities information, we actually may not expose it in P4Runtime but in the Openconfig data models. For example, we plan on using Openconfig for port configuration (enable / disable ports, query port status, ...). IMO the COS information would belong there as well based on this proposal.

Also, is there no intention of exposing an API to configure scheduling and rate-limiting based on COS id? If this is considered out of scope of PSA, I'm not sure this document should include any discussion about COS runtime APIs.

p4-16/psa/PSA.mdk
jafingerhut: I have changed the text not to mention the queueing structure, except that a PSA implementation might implement different FIFO queues for different classes of service.

p4-16/psa/PSA.mdk
jafingerhut: I have changed the text not to mention the queueing structure, except that a PSA implementation might implement different FIFO queues for different classes of service.

p4-16/psa/PSA.mdk
jafingerhut: The sentence about recommended weighted fair queueing scheduling policy is still in there, but note that it is only a recommendation, not mandated.

p4-16/psa/PSA.mdk
jafingerhut: I added an 'if' statement to the pseudocode describing what happens to the packet at the end of ingress processing, that specifies option 1, with optional error status recorded about the drop.

p4-16/psa/PSA.mdk
jafingerhut: The intent is that it is only for retrieving capability information about the device, as the PSA currently allows that different PSA devices may implement a different number of class of services from each other.

I certainly think it would be useful if there were an API that all controllers could use to configure PSA devices that have multiple classes of service.  If you can get a proposal to pass Vladimir's variety of implementations he doesn't want to exclude, go for it :-)

jafingerhut: Would config information like the rate that timestamp_t values advance in a device also belong in an Openconfig data model, vs. the P4Runtime API?  I had earlier created some proposed P4Runtime API messages to retrieve that data from a device, too.

p4-16/psa/PSA.mdk
jafingerhut: From 2017-Sep-07 PSA working group meeting, updated the behavior in the case where the user program assigns a value to class_of_service that is not supported by the platform, to use class_of_service 0, instead of dropping the packet.

p4-16/psa/PSA.mdk
jafingerhut: @antoninbas @cc10512 I would recommend merging these changes in as written right now, and then do a later pass of editing to clean up control plane API issues, focused just on that topic.

p4-16/psa/PSA.mdk
```diff
@@ -299,6 +313,71 @@ copy going to the control CPU to have a software-defined header
...
 packet copies going to the Egress control block.
 
+A PSA implementation may implement multiple classes of service for
+packets sent to the packet buffer.  If so, the Ingress control block
+may choose to assign a value to the `ostd.class_of_service` field to
+change the packet's class of service to a value other than the default
+of 0.
+
+PSA only specifies how the Ingress control block can choose the class
+of service the packet is assigned.  PSA does not mandate a scheduling
```
cc10512: It seems to me that with the restriction to just Ingress, defining a BQE block at the end of the pipeline is not very useful. We still define drop/truncate on it such that we have symmetrical functionality for these operations with the Ingress, so we may want to keep it. But then, maybe we should rename it to something that matches the reduced functionality.

p4-16/psa/PSA.mdk
cc10512: I'm fine with including the definition, however, we may want to say that these are guidelines that will be proposed to the P4Runtime WG. E.g.:
> The control plane API excerpt below is intended as a guideline to be included as part of the P4Runtime API[^P4RuntimeAPI] once discussed and approved.

jafingerhut: I added a note marked with "TBD" in the PSA spec for now, reminding us to revisit this part of the proposed control plane API.  If it is better to explain it differently than it is written up currently, I'm open to suggestions, but would like to get the rest of these changes merged in before that decision is made, if that is acceptable.

PR #409, opened 2017-09-06
Action profile and selector specification

samar-abdi:


jafingerhut:
@samar-abdi The diffs as I see them on Github seem to include duplicates of recently merged changes, unrelated to your changes.  Would it be much trouble to try creating a PR with only your changes in Github's diffs?

I don't know if it would be helpful, as I am not sure whether the notes at the link below always give predictable results when trying to rebase a commit and update a Github PR, but am linking to it in case it does help: https://github.com/jafingerhut/dotfiles/blob/master/bin/git-rebase.sh

samar-abdi:
Sorry Andy, forgot to pull before adding my changes. You should have received my new PR now.

PR #406, opened 2017-08-31
PSA updates to simplify options for cloning packets

jafingerhut:

PR #415, opened 2017-09-13
cleanup: naming conventions, grammar, text, etc.

### LINKED ISSUE #315: [PSA] consistent naming convention

cc10512:
Adds a section on naming conventions and follows up on implementing it.

Fixes #315.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -463,13 +463,7 @@ common use case is to the send the cloned packet to the control CPU,
 in which case the `clone_port` should be set to the value that
 represents the control CPU port.
 
-The PSA specifies four types of cloning, with the packet sourced from
-different points in the pipeline and sent back to ingress or to the
-buffering queue in the egress. (TBD: Which are the four??)
-
 ```
-[INCLUDE=psa.p4:Cloning_methods]
-
 [INCLUDE=psa.p4:Clone_extern]
 ```
 
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -73,7 +73,8 @@ each such egress port, the packet passes through an egress match
 action pipeline and a checksum update calculation before it is
 deparsed and queued to leave the pipeline. Note: the checksum
 operations are available -- validation in the parser,
-and checksum computation and update in deparser (see also Table [])
+and checksum computation and update in deparser (see also
+Table [#table-extern-usage])
 however, they need to be explicitly invoked though the checksum extern
 in order to be executed.
 
@@ -168,6 +169,8 @@ instance is instantiated in Ingress, for example, then it cannot be
 referenced, and thus its methods cannot be called, from any
 non-Ingress control block.
 
+~ TableFigure { #table-extern-usage; \
+  caption: Summary of controls that can instantiate and invoke externs.}
 |-------------------|----------------------------------------------|
 | Extern type       | Where it may be instantiated and called from |
 +:------------------+:---------------------------------------------+
@@ -184,6 +187,7 @@ non-Ingress control block.
 | `Register`        | Ingress, Egress |
 | `ValueSet`        | IngressParser, EgressParser |
 |-------------------|----------------------------------------------|
+~
 
 PSA implementations need not support instantiating these externs at
 the top level.  PSA implementations are allowed to accept programs
@@ -424,7 +428,7 @@ recirculate, and resubmit in PSA.
 A PSA implementation provides a cloning mechanism to create copies of
 the original packet as independent packet instances. A cloned
 packet may be a duplicate of the original packet or a copy of the
-deparsed packet after egress pipeline. The cloning may mechanism submit
+deparsed packet after egress pipeline. The cloning mechanism submit
 the cloned packet to either the ingress parser or the buffering mechanism.
 In a PSA implementation, the cloning mechanism is logically implemented in
 the PRE, that is, even if targets chose to implement it as part of other
@@ -449,9 +453,10 @@ block.
 
 A PSA implementation provides two configuration metadata, `clone`
 and `clone_port`, to the PRE to control the cloning mechanism. When
-the `clone` bit is set, the clone mechanism generates a cloned packet with the
-optional attached metadata. When the `clone` bit is unset or
-uninitialized, cloning is disabled and a cloned packet is
+`clone` is set to true, the cloning mechanism generates a
+cloned packet with the
+optional attached metadata. When `clone` is set to false,
+cloning is disabled and a cloned packet is
 not generated even if the `emit` method is invoked in the deparser.
 The `clone_port` controls the destination of the cloned packet. A
 common use case is to the send the cloned packet to the control CPU,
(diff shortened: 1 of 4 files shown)


p4-16/psa/PSA.mdk
```diff
@@ -86,13 +91,30 @@ controls.
...
+- Struct types are named using lower case words separated by `_`
+  followed by `_t`. For example `psa_ingress_input_metadata_t`.
+- Actions, extern methods, extern functions, headers, structs, and
+  instances of controls and externs start with lower case and words
+  are separated using `_`. For example `send_to_port`.
+- Enum members, const definitions, and #define constants are all
+  caps, with words separated by `_`. For example 'PORT_CPU'.
+
+Architecture specific metadata (e.g. structs) are prefixed by `psa_`.
+
```
vgurevich: Great stuff! Do we also want to define a prefix?

cc10512: You mean for all the other names? I thought about putting it out for debate, but it seems a bit too much. I would favor a "namespace" rather than prefixing each control type with PSA. However, it is certainly something we can debate in the next meeting.

p4-16/psa/PSA.mdk
```diff
@@ -70,7 +71,11 @@ on where the packets should go. After the ingress pipeline, the packet
 may be buffered and/or replicated (sent to multiple egress ports). For
 each such egress port, the packet passes through an egress match
 action pipeline and a checksum update calculation before it is
-deparsed and queued to leave the pipeline..
+deparsed and queued to leave the pipeline. Note: the checksum
+operations are available -- validation in the parser,
+and checksum computation and update in deparser (see also Table [])
```
jafingerhut: I don't know how to make a cross-reference to a table, but I do know that a cross-reference to a section can be done like so:
```
[#sec-extern-restrictions]
```
Is the table in that section the one you intended to cross-reference here?  If so, the section that table in is quite short, so a cross-reference to the whole section should be helpful enough to find the table.

p4-16/psa/PSA.mdk
```diff
@@ -387,31 +408,36 @@ no payload.
...
-packet can be a duplicate of the original packet or a copy of the
-deparsed packet after egress pipeline. The clone mechanism can submit
-the cloned packet to ingress parser or buffering mechanism. In a PSA
-implementation, the clone mechanism is implemented in PRE.
-
-The clone mechanism can optionally attach metadata to the cloned
+A PSA implementation provides a cloning mechanism to create copies of
+the original packet as independent packet instances. A cloned
+packet may be a duplicate of the original packet or a copy of the
+deparsed packet after egress pipeline. The cloning may mechanism submit
```
jafingerhut: Typo "cloning may mechanism".

Also @hanw said that he thought clone operations should be restricted to one of these possibilities only:

+ clone initiated at the end of the ingress control block can only be a clone of the original packet that triggered the clone, as received by the ingress parser, optionally with additional 'unordered metadata' (not headers) attached to it, currently specified via a clone.emit() method call in the ingress deparser (see Note 1 below).  The next place in the P4 code that will process such a clone will always be the egress parser.

+ clone initiated at the end of the egress control block, which can only be a clone of the packet as output by the egress deparser.  The next place in the P4 code that will process such a clone will always be the egress parser.

(Note 1) I believe that emit() is a misleading name to use for anything _except_ emitting headers, which have a specified bit order between the header and what gets put into the packet.  If the intent is that clone.emit() can be of a struct, and the data never leaves the PSA device on the wire, and therefore the fields might be packed in arbitrary order, and it will never be parsed by any parser in the PSA device, then I think it would be clearer to use a different method name, e.g. clone.attach_metadata() or something that conveys the behavior a bit better.

p4-16/psa/PSA.mdk
```diff
@@ -421,41 +447,40 @@ error to instantiate the `clone` extern in the P4 control or parser
 block.
 
 A PSA implementation provides two configuration metadata, `clone`
-and `clone_spec`, to the PRE to control the cloning mechanism. The
-`clone` enables/disables the cloning mechanism. If the `clone`
-bit is set, the clone mechanism generates a cloned packet with the
-optional attached metadata. If the `clone` bit is unset or
-uninitialized, the clone mechanism is disabled and a cloned packet is
+and `clone_port`, to the PRE to control the cloning mechanism. When
+the `clone` bit is set, the clone mechanism generates a cloned packet with the
```
jafingerhut: Maybe replace "bit is set" with "bool is true", to match the currently proposed type of 'bool' for the clone field?

p4-16/psa/PSA.mdk
```diff
@@ -421,41 +447,40 @@ error to instantiate the `clone` extern in the P4 control or parser
...
+optional attached metadata. When the `clone` bit is unset or
+uninitialized, cloning is disabled and a cloned packet is
```
jafingerhut: Elsewhere in the PSA draft the `clone` bool cannot be uninitialized, as it is defined to be initialized to `false` at the beginning of the ingress and egress control blocks.

p4-16/psa/PSA.mdk
```diff
@@ -421,41 +447,40 @@ error to instantiate the `clone` extern in the P4 control or parser
...
+buffering queue in the egress. (TBD: Which are the four??)
 
 ```
-extern clone {
-  /// Write @hdr into the ingress/egress clone engine.
-  /// @T can be a header type, a header stack, a header union, or a struct
-  /// containing fields with such types.
-  void emit<T>(in T hdr);
-}
+[INCLUDE=psa.p4:Cloning_methods]
```
jafingerhut: If the restrictions on clone options that @hanw is proposing are kept (see this PR, which maybe should be merged in before this one? https://github.com/p4lang/p4-spec/pull/406 ), then these enum's in psa.p4 will be unnecessary, and the PR I linked proposes deleting those enum types completely.

cc10512:
@jafingerhut i addressed your bit undefined comments and the reference to the table. i'm fine with either ordering of PRs. Up to @hanw to decide, he seemed to be ok with this going in :)

p4-16/psa/PSA.mdk
hanw: Let's merge andy's PR first, then this one. I will follow up with another PR to clean up the clone section today.

p4-16/psa/PSA.mdk
vgurevich: Do we have support for namespaces? Namespaces are definitely cleaner than prefixes, but then we need to have it.

jnfoster:
Sadly, no.

p4-16/psa/PSA.mdk
vgurevich: So, I suggest we discuss it on Monday. The proposal is to add and consistently use psa_ or PSA_ prefix (depending on the particular naming convention).

PR #416, opened 2017-09-13
specifies to drop packets in case egress_port is invalid

### LINKED ISSUE #317: [PSA] What if no valid forwarding operation is specified?
What happens if neither send_to_port(), nor multicast(), nor drop() are called?

What happens if port, specified in send_to_port doesn't exist on a given device?

cc10512:
Fixes #317 (actually only the remaining clarification).

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -225,6 +225,11 @@ after the Ingress control block has completed executing, based upon
 the contents of several metadata fields in the struct
 `psa_ingress_output_metadata_t`.
 
+Note that if at the end of Ingress control block the egress_port is
+not set, or it is set to a value that is not supported by the target,
+the PSA specifies that the packet is dropped. A target implementation
+may optionally chose to count such packets.
+
 ```
 [INCLUDE=psa.p4:Metadata_ingress_output]
 
@@ -258,7 +263,11 @@ the contents of several metadata fields in the struct
             plane configuration of multicast group ostd.multicast_group.
         return;   // Do not continue below.
     }
-    enqueue one packet for output port ostd.egress_port
+    if (ostd.egress_port undefined)
+        Drop the packet and optionally count the packets that are
+        dropped because of it.
+    else
+        enqueue one packet for output port ostd.egress_port
 ```
 
 TBD: Need text defining, for each possible copy, exactly what the
@@ -434,7 +443,7 @@ represents the control CPU port.
 
 The PSA specifies four types of cloning, with the packet sourced from
 different points in the pipeline and sent back to ingress or to the
-buffering queue in the egress. 
+buffering queue in the egress.
 
 ```
 extern clone {


p4-16/psa/PSA.mdk
```diff
@@ -225,6 +225,11 @@ after the Ingress control block has completed executing, based upon
 the contents of several metadata fields in the struct
 `psa_ingress_output_metadata_t`.
 
+Note that if at the end of Ingress control block the egress_port is
```
jafingerhut: I don't see any way for a PSA implementation to detect that the egress_port "is not set".  How are you thinking it could do that?

p4-16/psa/PSA.mdk
```diff
@@ -258,7 +263,11 @@ the contents of several metadata fields in the struct
             plane configuration of multicast group ostd.multicast_group.
         return;   // Do not continue below.
     }
-    enqueue one packet for output port ostd.egress_port
+    if (ostd.egress_port undefined)
```
jafingerhut: Because of my previous comment above, I would propose that this if condition should be something like "if (ostd.egress_port value is not supported)".

Currently the PSA draft says that the initial value of ostd.egress_port when the ingress control block begins executing is undefined.  That means that it could be _any value_.  I cannot think of any way that an implementation can know that the value has been assigned a value by an assignment statement in the ingress control block, vs. not, unless we change the spec to say that the initial value of ostd.egress_port must be some value known to be unsupported by the implementation.

cc10512: The compiler can find if all paths to the end of ingress assign egress_port or not (remember Nate's talk). If not, the value remains "undefined". It is true that we don't have a way to represent undefined, so this might be a mute point, but I do think that if we agree on the "undefined" value in the beginning, we can find a way of detecting that it is still undefined at the end. That said, I'd be happy to change it to not supported here, but then we still have to address the initial value.

jafingerhut: By default, if you assign to none of these metadata fields in the ingress control block, it doesn't matter if ostd.egress_port is initially undefined, because the ostd.drop is defined to be initialized to true, and this pseudocode will never reach this if statement.

If you want it to be predictable across PSA implementations that a program whose ingress control block assigns ostd.drop to false, but never assigns a value to ostd.egress_port, I would think that the simplest way to guarantee this is to define in PSA that the initial value of ostd.egress_port is a defined value (perhaps a #define constant name whose value can vary from one PSA implementation to another?) that is guaranteed to be unsupported by that implementation.

jafingerhut: Regarding the compiler finding all paths -- I had been thinking that this pseudocode here in the PSA spec is to specify run-time behavior of the PSA implementation.  We could recommend that a compiler ought to catch this, sure.  That doesn't help a run-time behavioral spec tell you what to do when this happens, though.

cc10512: Sorry @jafingerhut, but I think we need to make a decision and move toward consistency: if we allow saying that the psa_[ingress/egress]_output_metadata can be undefined, I don't see why we can't say in pseudocode that the value may end up here undefined. If we want to be more precise and acknowledge that undefined values are really not representable in the architecture, then, let's define _what_ the undefined values mean.

And talking about defaults drop == true and egress_port == undefined, I hope that people realize that just setting the egress_port to a proper value is not sufficient, we also need to flip drop to false.  Explicitly.

vgurevich: +1 on explicitly flipping the drop to false. And we can code that in the "helper" functions.

jafingerhut: @cc10512 I agree we need to make a decision.

By an 'undefined' value I mean here that the value is _some_ sequence of bits of the appropriate width, but it is not predictably portable what those values will be, so you cannot rely on them.  However, unlike Perl (where you can compare a variable's value to 'undef', because there is special logic in Perl to have a variable's value equal to a special 'undef' value, and therefore you can test for it at run time), P4_16 has no 'undef' run-time value you can compare against.  There are the bit vector values that are possible, and every bit<10> value will have one of the 2^10 values at run time at any given moment.

I am perfectly happy to say in the PSA that the initial value of ostd.egress_port is 0, or whatever specific defined value you want.

The only thing I am objecting to is that pseudocode defining what a PSA implementation does _at run time_ should not say "if value X is undefined".  There is no way for a PSA implementation to implement that phrase _at run time_.

I am perfectly fine using a phrase like this in the PSA pseudocode: "if (ostd.egress_port is a value that is not associated with any output port) { drop the packet }".  That can be implemented at run time.

jnfoster:
Chiming in late, but I think what you were both moving toward was to have PSA define a constant that is the initial value of the egress spec and does not correspond to any port, physical or virtual. Although the semantics of "undefined values" is somewhat subtle in P4_16 (as we all know) at least we would have that if that field is never assigned to, then it still has that value.

I agree with @jafingerhut that it's problematic to talk about undefined values.

I also like this suggestion:

> I am perfectly fine using a phrase like this in the PSA pseudocode: "if (ostd.egress_port is a value that is not associated with any output port) { drop the packet }". That can be implemented at run time.

PR #417, opened 2017-09-13
specifies the returned values for uninitialized meters and registers

### LINKED ISSUE #324: [PSA] meter and register initial state
Meters and registers: what's the initial state?

cc10512:
Fixes #324

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -787,9 +787,9 @@ algorithm used with protocols such as IPv4, TCP, and UDP -- see [IETF
 RFC 1624](https://tools.ietf.org/html/rfc1624) for details.
 
 ```
-[INCLUDE=psa.p4:InternetChecksum_extern]    
-```            
-    
+[INCLUDE=psa.p4:InternetChecksum_extern]
+```
+
 ### Checksum examples { #sec-checksum-examples }
 
 The partial program below demonstrates one way to use the `Checksum`
@@ -806,7 +806,7 @@ the location within a packet that a parser error occurred. A P4
 program author can choose to record such location information
 explicitly. For example, one may define metadata fields specifically
 for that purpose -- e.g. to hold an encoded value representing the last
-parser state reached, or the number of bytes extracted so far -- and then 
+parser state reached, or the number of bytes extracted so far -- and then
 assign values to those fields within the parser state code.
 
 ```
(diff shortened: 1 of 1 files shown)


p4-16/psa/PSA.mdk
```diff
@@ -921,6 +921,9 @@ in such a way that they return a `GREEN` (also known as conform),
 conditions under which one of these three results is returned.  The P4
 program is responsible for examining that returned result, and making
 changes to packet forwarding behavior as a result.
+The value returned by an uninitialized meter is undefined. If a target
+implementation chooses to return some value, the PSA recommends the
+value should be `GREEN`.
```
antoninbas: not sure how relevant this it to you, but we plan in `P4Runtime` to _require_ the default value to be `GREEN`, at least for _direct_ meters.

cc10512: Thanks for the feedback. I do believe it is slightly better to encourage initialization rather than relying on a required default value. It is true that ultimately there is no "undefined" value -- the bits have to be set to some combination of 0s and 1s, and we say that should be `GREEN`. I can change the recommendation to a requirement if people have no objections.

cc10512:
@jafingerhut, your take on this?

jafingerhut:
@cc10512 @antoninbas When you say that you plan in P4Runtime to require the default value to be `GREEN`, at least for direct meters, do you mean that if you have a table with direct meters, and you create a new table entry, but have not yet specified any parameters for the meter (e.g. you have not yet used an appropriate control plane API to specify a committed and peak information rate, etc.), then no matter how high a rate that the meter is accessed, it should always return `GREEN`?  That is, effectively the meter's rate and/or burst sizes are infinite, or at least faster than any rate that the device can process packets and/or bits?

If so, that sounds like a reasonable default setting to me, perhaps both for direct and indexed meters that the control plane has not yet initialized.  It seems good to call this out as a special case for PSA implementers.

cc10512:
@jafingerhut, indeed, that is the intent: a meter that has not been initialized by the control plane will always return GREEN. The question for the PSA WG was whether we _mandate_ GREEN as the return value, or we _recommend_ it, and leave to a target implementation the decision of what should be the default. If the P4Runtime API requires it, I propose that we mandate it as well.

antoninbas:
@jafingerhut For P4 Runtime, I think the agreement right now is that adding an entry to a table will automatically _initialize_ direct counters to 0 and direct meters in a way that GREEN is always returned. Unless explicit initialization values are provided (https://github.com/p4lang/PI/blob/master/proto/p4/p4runtime.proto#L159).
We haven't discussed indirect resources yet. It seems that for the sake of consistency, we could specify the same default initialization values.
A note on meters: a hardware target may not be able to strictly implement this property (infinite rates), in which case the target should do "best-effort".

jafingerhut:
@cc10512 Mandating it in PSA seems like a reasonable thing to do.

@antoninbas At least the PSA spec says that meters should support packet/bit rates up to their highest port speed.  It is possible to access individual meters at a higher rate than that, of course, but at least that rate should be able to return green for all packets, if a P4 program is written in a way so that an individual meter is not accessed at a higher rate than all packet from/to a single port.  Seems like pretty good 'best effort' in most cases.

cc10512:
@jafingerhut fixed.

PR #419, opened 2017-09-15
Proposal for Digest mechanism

hanw:


p4-16/psa/PSA.mdk
```diff
@@ -1161,11 +1161,26 @@ corresponding table entries.
...
+data object to the control plane. The data object may contain a
+identifier to identify the receiver of the object. The data object can
+optionally attach metadata to provide more information about the
+source of the digest. The attached metadata can be one or more header
+fields from a packet or one or more metadata generated in the Ingress
+or Egress pipeline. The digest extern can only be invoked in the
+deparser.
+
+A common use case of the digest mechanism is MAC learning. In the MAC
+learning case, the receiver of the digest is the CPU port, represented
```
cc10512: why is it important here how is the CPU port represented? And actually `PortId_t` is not an enum.

p4-16/psa/PSA.mdk
```diff
@@ -1161,11 +1161,26 @@ corresponding table entries.
...
+by a enum type. The digest metadata is the source MAC address of the
+received packet.
 
-\TODO: is generating a new packet and sending it to the stream or is
-it adding a header to the current packet and sending it to the
-stream (copying or redirecting).
+It may be desirable to avoid sending a burst of identical digests to
+the control plane, to save the processing cycles of the control plane
+processor for other tasks. In this case, the PSA implementation can
+implement a deduplication mechanism to filter repeated digests.
```
cc10512: may implement. Since we're not providing any guarantees on latency and any support for aggregation, is it even worth talking about this? Maybe we can move this to a "note to implementers".

p4-16/psa/psa.p4
```diff
@@ -524,8 +524,8 @@ extern ActionSelector {
 
 // BEGIN:Digest_extern
 extern Digest<T> {
-  Digest(PortId_t receiver); /// define a digest stream to receiver
-  void emit(in T data);      /// emit data into the stream
+  Digest(PortId_t receiver);    /// define a digest stream to receiver
+  void add_metadata(in T data); /// emit data into the stream
```
cc10512: `void add_metadata<T>(in T data); /// ...`

vgurevich: While this might look OK from the data plane side, it makes it fairly difficult to generate proper control plane APIs, that are supposed to somehow extract the data that we put into the Digest receiver in the data plane using the names of the corresponding fields.

hanw:
This is superseded by #454

PR #421, opened 2017-09-18
Checksums

jnfoster:
This PR implements changes discussed in preceding working group meetings. Note that I missed the last meeting, so I am going from the notes and discussions with those who attended -- i.e., it possible something got lost in translation.

The major changes include:
* Remove `ComputeChecksum` control and require checksum computation and update to be done in deparser.

* Clarifying that incremental checksumming is only supported for CRC{16,32}.

To discuss:

* Should we indeed move checksum updates to deparser?

* What do to with deparser direction annotations? Can copy the headers to a new variable, but it's a shame that we have to do that.

* Is my undersatnding of the intended restrictions on incrmental checksums correct? In particular, calling `update` many times is fine, but calling `remove` is only okay for CRC{16,32}. 

* Add an example of incremental TCP checksum?

### DIFF
--- a/p4-16/psa/examples/psa-example-incremental-checksum.p4
+++ b/p4-16/psa/examples/psa-example-incremental-checksum.p4
@@ -1,0 +1,202 @@
+/*
+Copyright 2017 Cisco Systems, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <core.p4>
+#include "../psa.p4"
+
+
+typedef bit<48>  EthernetAddress;
+
+header ethernet_t {
+    EthernetAddress dstAddr;
+    EthernetAddress srcAddr;
+    bit<16>         etherType;
+}
+
+header ipv4_t {
+    bit<4>  version;
+    bit<4>  ihl;
+    bit<8>  diffserv;
+    bit<16> totalLen;
+    bit<16> identification;
+    bit<3>  flags;
+    bit<13> fragOffset;
+    bit<8>  ttl;
+    bit<8>  protocol;
+    bit<16> hdrChecksum;
+    bit<32> srcAddr;
+    bit<32> dstAddr;
+}
+
+header tcp_t {
+    bit<16> srcPort;
+    bit<16> dstPort;
+    bit<32> seqNo;
+    bit<32> ackNo;
+    bit<4>  dataOffset;
+    bit<3>  res;
+    bit<3>  ecn;
+    bit<6>  ctrl;
+    bit<16> window;
+    bit<16> checksum;
+    bit<16> urgentPtr;
+}
+
+struct fwd_metadata_t {
+    bit<32> old_srcAddr;
+}
+
+struct metadata {
+    fwd_metadata_t fwd_metadata;
+}
+
+struct headers {
+    ethernet_t       ethernet;
+    ipv4_t           ipv4;
+    tcp_t            tcp;
+}
+
+
+// Define additional error values, one of them for packets with
+// incorrect IPv4 header checksums.
+error {
+    UnhandledIPv4Options,
+    BadIPv4HeaderChecksum
+}
+
+typedef bit<32> PacketCounter_t;
+typedef bit<8>  ErrorIndex_t;
+
+const bit<9> NUM_ERRORS = 256;
+
+parser IngressParserImpl(packet_in buffer,
+                         out headers parsed_hdr,
+                         inout metadata user_meta,
+                         in psa_ingress_parser_input_metadata_t istd,
+                         out psa_parser_output_metadata_t ostd)
+{
+    state start {
+        buffer.extract(parsed_hdr.ethernet);
+        transition select(parsed_hdr.ethernet.etherType) {
+            0x0800: parse_ipv4;
+            default: accept;
+        }
+    }
+    state parse_ipv4 {
+        buffer.extract(parsed_hdr.ipv4);
+        verify(parsed_hdr.ipv4.ihl == 5, error.UnhandledIPv4Options);
+        transition select(parsed_hdr.ipv4.protocol) {
+            6: parse_tcp;
+            default: accept;
+        }
+    }
+    state parse_tcp {
+        buffer.extract(parsed_hdr.tcp);
+        transition accept;
+    }
+}
+
+// BEGIN:Incremental_Checksum_Table    
+control ingress(inout headers hdr,
+                inout metadata user_meta,
+                PacketReplicationEngine pre,
+                in  psa_ingress_input_metadata_t  istd,
+                out psa_ingress_output_metadata_t ostd) {
+    action drop() {
+      ingress_drop(ostd);
+    }
+    action forward(PortId_t port, bit<32> srcAddr) {
+      user_meta.fwd_metadata.old_srcAddr = hdr.ipv4.srcAddr;
+      hdr.ipv4.srcAddr = srcAddr;
+      send_to_port(ostd, port);      
+    }
+    table route {
+        key = { hdr.ipv4.dstAddr : lpm; }
+        actions = {
+          forward;
+          drop;
+        }
+    }
+    apply {
+        if(hdr.ipv4.isValid()) {
+          route.apply();
+        }
+    }
+}
+// END:Incremental_Checksum_Table        
+
+parser EgressParserImpl(packet_in buffer,
+                        out headers parsed_hdr,
+                        inout metadata user_meta,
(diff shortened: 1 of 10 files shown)


jafingerhut:
I would think that `remove` would be difficult to implement for CRCs, and in fact I think it is impossible unless you indicate the 'bit position range' of the data you are removing/inserting.

Restricting `remove` to whatever name we use for the IPv4/TCP/UDP Internet checksum seems reasonable.  I think it might currently be called `ones_complement16` in psa.p4 right now (there may be benefits to renaming that to whatever P4_14 uses now).

Changing the direction of the headers parameter to the deparser from 'in' to 'inout' is a straightforward way to allow the checksum fields to be modified in place.  I understand that has the down side that the compiler would then permit arbitrary header fields to be modified in the deparser, without issuing any errors for doing so.

An alternative is to require the checksums to be updated at the end of the ingress or egress control blocks, but the down side there is that it makes P4_16 statements like 'return' and 'exit' harder or impossible to use, without also skipping the checksum updating.

mbudiu-vmw:
You could also consider having two types of checksum blocks: `Checksum` and `IncrementalChecksum`, where the latter has a `remove` method, but the former does not.

p4-16/psa/examples/psa-example-parser-checksum.p4
```diff
@@ -211,32 +211,30 @@ control egress(inout headers hdr,
...
-        hdr.ipv4.hdrChecksum = ck.get();
     }
 }
 // END:Compute_New_IPv4_Checksum_Example
 
-control DeparserImpl(packet_out packet, in headers hdr) {
+// BEGIN:Compute_New_IPv4_Checksum_Example
+control DeparserImpl(packet_out packet, in headers hdr_) {
+    Checksum<bit<16>>(HashAlgorithm_t.ONES_COMPLEMENT16) ck;
+    headers hdr = hdr_; //TBD: or better, change declaration to inout?
```
cc10512: What are the implications of changing the argument to `inout`? Are we then going to rely on convention that certain operations are not allowed?

jnfoster:
Resolved at 9/18 meeting:

* Eliminate `ComputeChecksum`
* Remove `remove` method from checksum extern.
* Add a new extern for `csum16`checksumming that supports `remove`
* No support for TCP payload -- must be done incrementally

I will take these marching orders and update the PR.

jnfoster:
@jafingerhut 

I believe I've made the changes we agreed on.

* Are you okay with calling the one's-complement incremental extern "Internet Checksum"?

* If the Internet checksum is always 16 bits, does it make sense to delete the width type parameter `W`?

jafingerhut:
Given that it is used for IPv4 headers, and TCP and UDP, Internet Checksum seems as good a name as any.

I am not aware of any other checksums that are incrementally updatable and commonly used in networking protocols.  If such a thing arises, it seems reasonable to add a new extern to P4_16 for that purpose.  Thus fixing InternetChecksum to input that is multiples of 16 bits wide, and output that is 16 bits wide, seems reasonable.

jnfoster:
I've made the following changes:

* Modified the type signature for the deparser to the following:
```
control Deparser<H, M>(packet_out buffer, inout H hdr, in M user_meta);
```
Specifically I added the user metadata as an `in` parameter.

* Added an example of computing an incremental TCP checksum.

* Fixed various typos and ran examples through `p4test` to verify that they are syntactically correct.

cc10512:
@jnfoster what's the conclusion on adding the `init` method?

p4-16/psa/examples/psa-example-incremental-checksum.p4
```diff
@@ -0,0 +1,202 @@
...
+            hdr.ipv4.protocol,
+            //hdr.ipv4.hdrChecksum, // intentionally leave this out
+            hdr.ipv4.srcAddr,
+            hdr.ipv4.dstAddr });
+        hdrChecksum = ck.get();
+        // Update TCP checksum
+       ck.clear();
+       ck.remove(hdr.tcp.checksum);
+       ck.remove(user_meta.fwd_metadata.old_srcAddr);
+       ck.remove(hdr.ipv4.hdrChecksum);
```
jafingerhut: The TCP pseudo-header includes the IPv4 source and dest address, but not the IPv4 header checksum, so this remove call, and the ck.update(hdrChecksum) call below, should be removed.  RFC 793 is a good reference for the contents of the TCP pseudo-header.

p4-16/psa/examples/psa-example-incremental-checksum.p4
```diff
@@ -0,0 +1,202 @@
...
+            hdr.ipv4.fragOffset,
+            hdr.ipv4.ttl,
+            hdr.ipv4.protocol,
+            //hdr.ipv4.hdrChecksum, // intentionally leave this out
+            hdr.ipv4.srcAddr,
+            hdr.ipv4.dstAddr });
+        hdrChecksum = ck.get();
+        // Update TCP checksum
+       ck.clear();
+       ck.remove(hdr.tcp.checksum);
```
jafingerhut: Nate, you asked me in email earlier whether remove() is correct to do here, and at the time I wasn't sure.  I am pretty darned sure now that this is a correct way to do an incremental TCP checksum now, primarily due to RFC 1624 and my own writing and thinking on the properties of the one's complement checksum.

p4-16/psa/examples/psa-example-incremental-checksum.p4
jafingerhut: I am going to go ahead and approve and merge in these changes, and then make the small fix suggested above on top of that.

p4-16/psa/examples/psa-example-incremental-checksum.p4
vgurevich: @jafingerhut ,

While this example is totally valid, it is probably not the way incremental checksums are computed in practice.

Instead, the following happens:
1. In the parser you compute the value of payload checksum by removing all the fields that TCP checksum is computed over from hdr.tcp.checksum and store it in the metadata, e.g. meta.tcpPayloadChecksum.
2. In the control, some of the fields that participate in tcp checksum are modified (which is the whole reason to recompute the checksum in the first place)
3. In the deparser, the TCP checksum is initialized with meta.tcpPayloadChecksum and then all the fields that TCP checksum is computed over are added to that checksum. Finally you do ck.get() and store it in hdr.tcp.checksum and then hdr.tcp is emit()ed.

As an optimization, one can remove/add only the fields that are actually modified in step 2.

jnfoster:
Ooh, that's nice. Shall we change the example?

-N

On Sat, Oct 14, 2017 at 2:37 PM, Vladimir Gurevich <notifications@github.com

p4-16/psa/examples/psa-example-incremental-checksum.p4
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// BEGIN:Incremental_Checksum_Example
+control DeparserImpl(packet_out packet, inout headers hdr, in metadata user_meta) {
+    InternetChecksum() ck;
+    bit<16> hdrChecksum;
+    apply {
+        // Update IPv4 checksum
+        ck.clear();
+        ck.update({ hdr.ipv4.version,
```
vgurevich: Also, maybe for clarity we should format this like so:
```
ck.update({
    hdr.ipv4.version, hdr.ipv4.ihl, hdr.ipv4.diffserv,  /* 16-bit word #1 */
    hdr.ipv4.totallen, 
    hdr.ipv4.identification,
    hdr.ipv4.flags, hdr.ipv4.fragOffset,
    hdr.ipv4.ttl, hdr.ipv4.protocol,
    hdr.ipv4.srcAddr,
    hdr,ipv4.dstAddr   });
```

The point I am trying to make is that the `update()` method of the `InternetChecksum` extern can only take data that is short-sized (i.e. 16*N bits wide).

Obviously, we do not have a way to express it syntactically, but I think it should be documented and the compiler should enforce that bit-width of T is a multiple of 16.

jafingerhut: Agreed that the update() and remove() methods should restrict their inputs to be multiples of 16 bits long, and this ought to be documented.  That change is included in this PR: https://github.com/p4lang/p4-spec/pull/440

It does update the format as shown above, though -- certainly an easy change to make, and no objections from me.

p4-16/psa/examples/psa-example-incremental-checksum.p4
jafingerhut: @vgurevich So a separate parser state for parsing a TCP header after IPv4 vs. parsing a TCP header after IPv6, if both IPv4 and IPv6 are supported in a P4 program, then?  Either that, or you need a single parser state that has an 'if' condition that distinguishes those two cases, and I'm not sure whether you'd be interested in mandating support for if statements inside of a parser state.

vgurevich: @jafingerhut ,  Yes, something along those lines (unless we want to revisit the separate ChecksumVerification control :) ). If() statement is more readable, but it is way more dangerous. 

Another option is to instantiate the extern outside (of any parser state) and then remove() IPv4/IPv6 fields in the corresponding states whether or not there is a TCP/UDP header afterwards. Then in the TCP parsing state, you add() tcp.CheckSum and remove() the rest of tcp fields. Same for UDP.

It's all not very pretty, but can probably be partially blamed on the quirkiness of the original (i.e. TCP/UDP) design.....

We can try to write a couple of examples and see which one is more palatable.

PR #422, opened 2017-09-18
Specify usage of don't care arguments

mbudiu-vmw:


mbudiu-vmw:
This is fully implemented.
@gbrebner's proposal is different.

PR #423, opened 2017-09-18
Clarify clone semantics

hanw:
Two main changes:
- disallow invoking emit() method multiple times for the same cloned packet. The generic type T should give program enough expressive power to put arbitrary data to a single emit() method.
- clarify that PSA specifies ingress-to-egress and egress-to-egress clone.

Point to debate: should egress-to-ingress and ingress-to-ingress be specified as optional?


p4-16/psa/PSA.mdk
```diff
@@ -444,12 +444,11 @@ the responsibility of the programmer to parse the cloned packet
...
-specified metadata to the same cloned packet. The PSA architecture
-instantiates the `clone` extern in the ingress and egress deparser. A
-P4 program can use it in ingress and egress deparser only. It is an
-error to instantiate the `clone` extern in the P4 control or parser
-block.
+above types. Invoking `emit` method multiple times on the same cloned
+packet is not allowed. The PSA architecture instantiates the `clone`
+extern in the ingress and egress deparser. A P4 program can use it in
+ingress and egress deparser only. It is an error to instantiate the
+`clone` extern in the P4 control or parser block.
```
cc10512: The instantiation is still confusing: you say that the PSA instantiates the clone extern -- is that implicit or does the user have to instantiate explicitly? If it is implicit, then there is no need for the later part. If it is explicit, then we should say that PSA only allows their instantiation in the ingress/egress deparser and instantiating anywhere else is an error.

p4-16/psa/PSA.mdk
```diff
@@ -463,6 +462,13 @@ common use case is to the send the cloned packet to the control CPU,
 in which case the `clone_port` should be set to the value that
 represents the control CPU port.
 
+
+The PSA specifies two types of cloning: cloning from ingress to egress
+and cloning from egress to egress. With cloning from ingress to
```
cc10512: s/With/When/

p4-16/psa/PSA.mdk
```diff
@@ -489,11 +495,10 @@ to the resubmitted packet. It is the responsibility of the programmer
...
 stack`, `header union` or `struct` of the above types. Invoking the
-`emit` method multiple times will attach all specified metadata to the
-same resubmitted packet. The PSA architecture instantiates the
-`resubmit` extern in the ingress deparser. A P4 program can use it in
-ingress deparser only. It is an error to instantiate the `resubmit`
-extern in the P4 control or parser block.
+`emit` method multiple times is not allowed. The PSA architecture
+instantiates the `resubmit` extern in the ingress deparser. A P4
+program can use it in ingress deparser only. It is an error to
+instantiate the `resubmit` extern in the P4 control or parser block.
```
cc10512: Same comment as for cloning.

hanw:
This commit is superceded by #452

PR #427, opened 2017-09-21
[do not merge] restores PR #381: Proposed additions to P4_16 spec: restrictions on use of stateful objects

cc10512:
This PR re-applies the changes that were submitted as part of #381 and which need to wait for the language working group discussion.


cc10512:
Merging per our discussion at the 10/23 LDWG meeting.

PR #428, opened 2017-09-22
add an example for clone extern

hanw:
Will add more clone examples. Feel free to comment.

### DIFF
--- a/p4-16/psa/examples/psa-example-clone-to-port.p4
+++ b/p4-16/psa/examples/psa-example-clone-to-port.p4
@@ -97,6 +97,7 @@ control ingress(inout headers hdr,
     action do_clone (PortId_t port) {
         ostd.clone = true;
         ostd.clone_port = port;
+        ostd.clone_instance = 0;
     }
     table t() {
         key = {
@@ -135,26 +136,50 @@ control egress(inout headers hdr,
 
 control computeChecksum(inout headers hdr, inout metadata meta) {}
 
-control DeparserImpl(packet_out packet, clone cl, in headers hdr) {
-    // Alternatively, we can provide a constructor to Clone().
-    // Clone() clone;
+
+control DeparserImpl(packet_out packet, inout headers hdr) {
     apply {
-        cl.emit({hdr.ethernet.srcAddr, hdr.ipv4.srcAddr});
-        // Use a different method name (pack?)
-        // cl.pack({hdr.ethernet.srcAddr, hdr.ipv4.srcAddr});
-
-        // XXX(hanw): multiple clone instances?
-
-        packet.emit(hdr.ethernet);
+        packet.emit(hdr.eth);
         packet.emit(hdr.ipv4);
+    }
+}
+
+control IngressDeparserImpl(packet_out packet,
+    clone_out clone,
+    inout headers hdr,
+    in userMetadata meta,
+    in psa_ingress_output_metadata_t istd) {
+    DeparserImpl() common_deparser;
+    apply {
+        if (istd.clone_instance == 0 && istd.clone) {
+            clone.add_metadata({hdr.eth.srcAddr, hdr.eth.dstAddr});
+        }
+        if (istd.clone_instance == 1 && istd.clone) {
+            clone.add_metadata({hdr.eth.dstAddr, hdr.eth.srcAddr});
+        }
+        common_deparser.apply(packet, hdr);
+    }
+}
+
+control EgressDeparserImpl(packet_out packet,
+    clone_out clone,
+    inout headers hdr,
+    in userMetadata meta,
+    in psa_egress_output_metadata_t istd) {
+    DeparserImpl() common_deparser;
+    apply {
+        if (istd.clone_instance == 0 && istd.clone) {
+            clone.add_metadata({hdr.ipv4.srcAddr, hdr.ipv4.dstAddr});
+        }
+        common_deparser.apply(packet, hdr);
     }
 }
 
 PSA_Switch(IngressParserImpl(),
            ingress(),
            computeChecksum(),
-           DeparserImpl(),
+           IngressDeparserImpl(),
            EgressParserImpl(),
            egress(),
            computeChecksum(),
-           DeparserImpl()) main;
+           EgressDeparserImpl()) main;
--- a/p4-16/psa/examples/psa-example-clone-to-port.p4
+++ b/p4-16/psa/examples/psa-example-clone-to-port.p4
@@ -113,12 +113,12 @@ parser IngressParserImpl(packet_in buffer,
     
     state parse_clone_header {
         cp.apply(buffer, istd, user_meta.clone_header);
-	transition accept;
+        transition parse_ethernet;
     }
 
     state parse_ethernet {
         p.apply(buffer, parsed_hdr, user_meta);
-	transition accept;
+        transition accept;
     }
 }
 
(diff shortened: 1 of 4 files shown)


p4-16/psa/examples/psa-example-clone-to-port.p4
```diff
@@ -0,0 +1,160 @@
...
+    apply { }
+}
+
+control computeChecksum(inout headers hdr, inout metadata meta) {}
+
+control DeparserImpl(packet_out packet, clone cl, in headers hdr) {
+    // Alternatively, we can provide a constructor to Clone().
+    // Clone() clone;
+    apply {
+        cl.emit({hdr.ethernet.srcAddr, hdr.ipv4.srcAddr});
```
vgurevich: So, where are these used? I do not see anything special in the egress parser to process these...

cc10512:
@vgurevich, @jafingerhut, @tomrodeheffer, please review and provide feedback.

p4-16/psa/examples/psa-example-clone-to-port.p4
```diff
@@ -135,26 +136,50 @@ control egress(inout headers hdr,
...
-        // XXX(hanw): multiple clone instances?
+control IngressDeparserImpl(packet_out packet,
+    clone_out clone,
+    inout headers hdr,
+    in userMetadata meta,
+    in psa_ingress_output_metadata_t istd) {
+    DeparserImpl() common_deparser;
+    apply {
+        if (istd.clone_instance == 0 && istd.clone) {
+            clone.add_metadata({hdr.eth.srcAddr, hdr.eth.dstAddr});
```
vgurevich: This is wonderful, but where is this consumed?

jafingerhut:
Are the new fields you introduce clone_instance, resubmit_instance, and recirculate_instance intended to be used only in the deparsers?  And the only reason they exist is to allow the deparser to do different things for different user-defined "kinds" of clone/resubmit/recirc'ed packets?  Might be good to have a comment in psa.p4 and/or in the PSA document to mention these things, if so.

p4-16/psa/examples/psa-example-clone-to-port.p4
jafingerhut: Yes, same question from me.  An example of adding metadata should show it being used somewhere.

(1) Is it intended to be parsed explicitly in the user's P4 program?  If so, do we use the 'extract' method for that purpose?  What order does this metadata occur in relative to other contents of the packet?  Always before the rest of the packet and any headers it might contain?  If there is more than one collection of metadata that can be added to a clone, how does the parser know which case to parse?

(2) Or is it intended to be more like bridged metadata, where the values you specify in add_metadata are auto-magically copied by the implementation to the fields with the same names in the egress code?

If so, do the names and types of the fields have to match exactly between ingress and egress?

Should the compiler give an error if a field name is given to the clone.add_metadata() method that does not exist in egress metadata?  If the name does exist, but the types are different?

Your example gives fields to clone.add_metadata in the hdr struct.  Which fields should a programmer be allowed to give as arguments here?  Inside user-defined metadata seems like an obvious choice to permit, but you are also suggesting user-defined headers, too.

jafingerhut: If the answer to the question "Is it intended that fields given to clone.add_metadata() are 'auto-magically' copied from ingress to egress?" is _yes_, then isn't bridged metadata pretty much like saying "normal_packet.add_metadata(user_meta);", where user_meta is the struct that contains all user-defined metadata?  (and the compiler is allowed to optimize out any fields from either list of arguments, if it can tell that the field is not referenced in the egress code)

If so, such a method for normal packets seems like a way to make metadata added to clones, and opt-in bridged metadata for normal packets, be specified in a common way in P4 programs.

hanw:
@jafingerhut , Thanks for the feedback, and sorry for the delay in reply.

I have updated the clone example and added a new example for mirror-on-drop. 
The example does not require the `clone_instance` metadata, I will remove them from my proposal. Instead, user is responsible to define a tag to differentiate multiple clone headers, and parse them accordingly in the parser.

The parser first check the instance_type of incoming packet, if the packet is of type "CLONE", then the clone parser is invoked; otherwise, the normal packet parser is invoked. In the clone sub-parser, the sub-parser can optionally use the `lookahead` method to read the user-defined clone tag. The sub-parser then extract the clone header based on the value of the clone tag.

A mirror on drop is implemented as setting both the `ostd.drop` and the `ostd.clone` as 1s.

p4-16/psa/examples/psa-example-clone-to-port.p4
hanw: I have updated the example to show the parser implementation.

The user program is responsible to parse the cloned header using the extract method. The clone header is prepended before other contents of the packet. The user can add arbitrary amount of metadata to the cloned header (up to the limit that a target supports). It is up to the user to parse the headers properly. As we only define clone from ingress to egress, and egress to egress, the parsing should be done in the egress parser. In the future, if we decided to define ingress to ingress and egress to ingress clone, the same rule applies to ingress parser. The name and types of the fields do not need to match between the deparser and parser (although it might be a good practice). But the length of the extracted header must be the same. I don't think compiler need to give an error if user decides to 'cast' the clone header to a different type upon receiving the header.

jafingerhut:
I believe that the P4_16 spec only allows the `extract` method to take an instance of a variable of type header as a parameter.  The proposed example program is calling extract with a struct type as a parameter.

It also seems to assume that add_metadata copies the fields of the struct given as a parameter in the order given in the source code (because the example parser assumes that the first 8 bits are a type field).  I believe in a previous working group meeting that someone (I have forgotten who) mentioned that an advantage of adding metadata (as opposed to emitting a header) was that the order of fields in the struct did not matter, and the compiler was free to pack it in the most efficient way allowed by the target.

I am not saying these are insurmountable issues, but if we want to stick with things already defined in P4_16, and requiring a particular order of fields in the copied metadata is acceptable, it seems that emit and extract on header types already gives us those things, and a P4 programmer should already be familiar with those things and how they work?

hanw:
I fixed the example to use header for clone header. The example does assume the `type` field is at the start of the clone header. One option is to use annotation to 'pin' the type field to the beginning of the clone header while allowing the compiler to freely choose the layout for the rest of the header.

jafingerhut:
OK, but realize that with the P4_16 v1.0.0 spec, this limits what can be put into clone metadata as only what can be put into a header.  So only these things:

+ `bit<W>`
+ `int<W>`
+ `varbit<W>`

None of these things are allowed inside a header definition:

+ error
+ enum
+ bool
+ struct
+ header (i.e. no header within a header)

Not sure if that is considered too limiting.

Besides the annotation for 'pinning' a type field to the beginning, are you also thinking that there is an annotation on the header as a whole that indicates that the order of bit-packing is otherwise whatever the compiler considers best?  And if so, headers with that annotation should not be used for packets that come in from outside the PSA device, nor for headers emitted that go outside of the device?

jafingerhut:
@cc10512 I believe the proposal at the last working group meeting was that we merge this one in as is, with the understanding that @hanw is still working on it, and it is likely to change?

cc10512:
@jafingerhut Actually, from the notes:
- issues with the transitions in the example, because it was not parsing the rest of the packet
**AI**: Han to fix
I think we should wait for the example to be fixed, not push in incorrect examples.

hanw:
the example is fixed.

cc10512:
That was fast! Thanks @hanw.

Issue #425, opened 2017-09-18
PSA: Scrub PSA metadata for direction


jafingerhut:
Currently psa.p4 has psa_ingress_output_metadata_t as a direction 'out' parameter of the Ingress control block, and it also claims to specify initial values for some of the fields when the Ingress control block begins execution.

This is not possible in P4_16.

One fix: For all fields that we wish to define an initial value, and also let the user's Ingress control block code modify it, put them in a struct with direction 'inout'.

For all fields that we wish to have no defined initial value (i.e. they are initially undefined in the P4_16 spec sense of the word), put them in a separate struct with direction 'out'.

Obviously, if we end up defining initial values for all such fields, there would be no need for a struct with direction 'out'.

cc10512:
Fixed by #434

PR #434, opened 2017-10-04
Change direction of some Ingress and Egress parameters from out to inout

jafingerhut:
We want these structs to have fields with defined values when the
Ingress and Egress control blocks begin execution.  This cannot be
done if they are defined as direction 'out', but can if they are
defined as direction 'inout'.

Also adds a missing type parameter <T> to one emit method definition
in psa.p4 that caused compilation errors.


jafingerhut:
Intended to be a fix for issue https://github.com/p4lang/p4-spec/issues/425

jafingerhut:
@cc10512 This is a very small, focused change.  The only part anyone would need to review would be the change to the file psa.p4, which is about 3 lines of code being changed, primarily two 'out' parameters being changed to 'inout'.

PR #435, opened 2017-10-04
Simplify the Random extern for uniform distributions only

jafingerhut:
The text briefly mentions how a user's P4 program can use this to
generate pseudo-random values with non-uniform distributions.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -1133,17 +1133,17 @@ $2^W$ entries.
 
 ## Random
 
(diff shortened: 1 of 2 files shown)


jafingerhut:
As per decision discussed during 2017-Sep-07 PSA working group meeting: https://github.com/p4lang/p4-spec/wiki/PSA-Meeting-Minutes-Sep-7,-2017

jafingerhut:
@cc10512 This is a pretty small change, one that ends up with fewer options and a simpler design than the original, so should be fairly quick for someone to review and merge, if you have a few minutes.

PR #440, opened 2017-10-13
PSA: Minor corrections to incremental checksum example code

jafingerhut:
Also add that program to those compiled via 'make check'.


jafingerhut:
I may have gone a bit overboard with adding an appendix with a reference implementation for the InternetChecksum.  Unlike the Checksum / Hash externs, this one has the requirement to conform with a widely deployed standard.

cc10512:
@sethfowler can you please review the checksum section merged in #421 and this PR and let us know if you see any issues?

sethfowler:
I have no problem with it, though my preferred solution for taking the checksum of the payload into account is still to provide an extern that models the payload checksum. It seems that that approach was already considered and rejected, though.

jafingerhut:
@cc10512 OK to merge this one in, then?

cc10512:
Yes, from my perspective. @jnfoster?

jafingerhut:
I have added another example P4 source program to this pull request, one that is not yet mentioned in the PSA spec document.  Right now it is there just for people to look at and think about.  Here is a link to the file in the Github pull request: https://github.com/p4lang/p4-spec/pull/440/files#diff-3ce992c739f1198a7e59fa1d6c9cd63c

Take a look at the proposed incremental checksum calculation code in the Deparser, in particular.  I don't see any reasonable way to eliminate the 'if' statements inside the deparser, if that is where we want checksum calculation to be in PSA programs.  Especially check out the nested if statements for the UDP incremental checksum calculation in the deparser code.  I don't see any simpler way to do it correctly.

Is this the kind of example program that you want PSA compiler authors to support?  Nested if statements are no big deal for the ingress/egress control blocks from P4_14, I know, and it should not in principle be any harder to support them in a deparser control block.  Just wanted to call it out for special attention.

jnfoster:
Wow, that is some beautiful code. Amazing!

I admittedly don't have deep hardware experience, but even so I am not too worried about the conditionals. For example, on a target with a pre-deparser checksum unit, it shouldn't be hard for the compiler to hoist the conditional and checksum code to before the emit code (although it is not entirely trivial).

vgurevich:
@jafingerhut ,

Thanks a lot -- this is really a very nice, detailed and clearly written code! I agree that there is not much we can simplify here (at least within the current approach). @cc10512  and @sethfowler (and others) should comment on the feasibility of compiling it onto a real HW target. The main challenge would probably be to detect a specific pattern regardless of style the code is written in.

jafingerhut:
Glad you folks liked the code.

I wanted to mention that I have probably spent a bit too much of my time writing and verifying this example.  In order to get better assurance that it is correct, I implemented the InternetChecksum update and remove methods using P4_16 controls (a separate control for each size of the data argument) so that I could test this implementation packet in/out behavior against Wireshark's calculation of IPv4, TCP, and UDP checksums (for empty TCP and UDP payloads, to keep the implementation simpler).

After finding several issues with my implementation, and finding and working around a p4c issue [1] in the process, the implementation matches Wireshark's behavior for IPv4+TCP, IPv4+UDP, IPv6+TCP, and IPv6+UDP packets, so I have pretty high confidence it is correct.  I also used a yet-to-be-released-as-open-source tool for creating packets that exercise all paths in a P4 program (based on the Z3 SMT solver) to verify that my P4_16 controls for update and remove always produce the same result in calculated-from-scratch vs. incremental update calculation for up to 5 16-bit words of input (e.g. half of the IPv4 header -- the full size IPv4 header made the Z3 formulas big enough that it didn't solve in a day, so I cut the problem size down).

[1] https://github.com/p4lang/p4c/issues/983

jafingerhut:
@cc10512 I have a slight preference for merging this PR in as it is now, and then follow up with a separate PR that makes the changes we discussed in today's PSA working group meeting.  But I could make those changes on top of this PR if you like.

cc10512:
I trust you you will, so please go ahead.

PR #441, opened 2017-10-13
PSA: Cleanly drop a packet if ingress ends with unsupported egress_po…

jafingerhut:
…rt value

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -244,6 +244,23 @@ after the Ingress control block has completed executing, based upon
 the contents of several metadata fields in the struct
 `psa_ingress_output_metadata_t`.
 
+The function `platform_port_valid()` mentioned below takes a value of
+type `PortId_t`, returning `true` only when the value represents an
+output port for the implementation.  It is expected that for some PSA
+implementations there will be bit patterns for a value of type
+`PortId_t` that do not correspond to any port.
+
+A comment saying "recommended to log error" is not a requirement, but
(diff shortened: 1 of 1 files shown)


jafingerhut:
Most of this is just white space changes due to deeper indenting of pseudo-code lines.  The only substantive changes are to specify that packets are dropped, or clone packets are not made, if the corresponding port values are not supported by the platform.

jafingerhut:
And we can close this issue if this PR is merged: https://github.com/p4lang/p4-spec/issues/317

p4-16/psa/PSA.mdk
```diff
@@ -244,6 +244,23 @@ after the Ingress control block has completed executing, based upon
 the contents of several metadata fields in the struct
 `psa_ingress_output_metadata_t`.
 
+The function `platform_port_valid()` mentioned below takes a value of
```
cc10512: Is the intent to define `platform_port_valid()` as a _must support_ extern? We should then define it as:
```
bool platform_port_valid(in PortId_t p);
```
And it seems that we should allow this extern to be instantiated and called anywhere in the pipeline: parser, ingress/egress controls, deparser.

jafingerhut: My intent when writing it was simply to name it and explain it one time, since it is used twice in the pseudocode.  I wasn't thinking that this would be a function that someone would want to call in their P4 programs.

If you think they would want to, we can do as you suggest.

If there isn't much reason for a P4 program author to call it (I can't think of a good reason), I can add a sentence or so mentioning this.

PR #445, opened 2017-10-18
Added action profile and selector description with figures and examples.

samar-abdi:


p4-16/psa/PSA.mdk
```diff
@@ -1288,45 +1288,199 @@ to generate values.
...
+match type `selector` when defining the table match key. The match fields of
+type `selector` are composed into a field list in the order they are specified.
+The composed field list is passed as an input to the action selector
+implementation. It is illegal to define a `selector` type match field if the
+table does not have an action selector implemenation.  
+
+The control plane can add, modify or delete member and group entries for a
+given action selector instance. An action selector instance may hold at most
+`size` member entries as defined in the constructor parameter. The number of
+groups may be atmost the size of the table that is implemented by the selector.
```
antoninbas: typo, atmost

p4-16/psa/PSA.mdk
```diff
@@ -1288,45 +1288,199 @@ to generate values.
...
+Action selector members may only specify action types defined
+in the `actions` attribute of the implemented table. An action selector instance
+may be shared across multiple tables only if all such tables define the same
+set of actions in their `actions` attribute. Tables with an action selector
+implementation cannot define a default action. The default action for such
+tables is implicitly set to `NoAction`.
+
+The dynamic selection algorithm requires a field list as an input for generating
+the index to a member entry in a group. This field list is created by using the
+match type `selector` when defining the table match key. The match fields of
```
antoninbas: is that the long term plan for PSA? I though the consensus was to use P4 "virtual" functions. Or do we think this won't get added to the language spec in time?

antoninbas:
Not sure this is the right place, but here is the proposal I wrote up a while ago for action selectors using abstract functions to define the hash algorithm:

```
struct H { };
struct M {
    bit<32> hash1;
    bit<32> hash2;
}

// my proposal
extern action_selector_psa<T> {
    action_selector_psa(bit<32> size);
    abstract T my_hash();
}

control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {

    action drop() { smeta.drop = 1; }

    // an example instantiation
    action_selector_psa<bit<16> > (32w1024) as = {
        bit<16> my_hash() {
            bit<16> res;
            // meta is in scope, which is what we need
            hash(res, HashAlgorithm.crc16, 16w0, {meta.hash1, meta.hash2}, 16w65535);
            return res;
            // ideally return hash(...)
        }
    };

    table indirect_ws {
        key = { }
        actions = { drop; NoAction; }
        const default_action = NoAction();
        implementation = as;
    }

    apply {
        indirect_ws.apply();
    }

};
```

mbudiu-vmw:
Abstract extern methods are an experimental feature that is already supported by the compiler.
It probably should be considered for inclusion in P4-XX for XX > 16.

Using variables available in the context, however, should not be supported.

jafingerhut:
@mbudiu-vmw By "variables available in the context", I am guessing you mean by that, in the sample code that @antoninbas published, the use of `meta.hash1` and `meta.hash2` inside of the function `my_hash`?

If so, how would you propose to write code with equivalent behavior that avoided doing this?

jafingerhut:
@samar-abdi From the discussion at the last work group meeting, I believe the proposal was that if you could add notes about these restrictions:

+  All actions in the same group must be the same.  The action parameters for actions in the same group are allowed to differ, and the action can be different for different groups.

+ An action_selector instance may be used by multiple tables, but if so, the list of fields given with match_kind `selector` must be the same for all such tables, including their order.

And perhaps also address the typo that Antonin commented on, then we are ready to merge this in.

samar-abdi:
Done. Thanks for the reminder @jafingerhut

mbudiu-vmw:
If you need to capture variables from the environment they have to be passed explicitly as arguments. The problem is that we don't know when the virtual method is going to be called.

p4-16/psa/PSA.mdk
```diff
@@ -1288,45 +1288,203 @@ to generate values.
...
+
+\TODO: is generating a new packet and sending it to the stream or is
+it adding a header to the current packet and sending it to the
+stream (copying or redirecting).
+
+```
+[INCLUDE=psa.p4:Digest_extern]
+```
+
+
```
jafingerhut: @samar-abdi Sorry, one more nit to pick -- the group of lines added above looks unrelated to the rest of this.  Not sure if it is easy to remove from your PR?

jafingerhut:
@cc10512 With Samar's latest changes from today, I think this one is ready to merge in.  I'm happy to do it if you have no objection.

I have copied the comments above related to virtual functions to a separate issue here for easier future reference: https://github.com/p4lang/p4-spec/issues/446

PR #448, opened 2017-10-22
add example to implement clone without an extern

hanw:
Unfortunately, this example triggers a bug in front end. I will file an issue in p4c.
```
psa-example-clone-without-extern.p4(217)
control EgressDeparserImpl(packet_out packet,
        ^^^^^^^^^^^^^^^^^^
libc++abi.dylib: terminating with uncaught exception of type Util::CompilerBug: COMPILER BUG: ../ir/visitor.cpp:145
Transform called const visit function -- missing template instantiation in gen-tree-macro.h?
```

### DIFF
--- a/p4-16/psa/examples/psa-example-clone-without-extern.p4
+++ b/p4-16/psa/examples/psa-example-clone-without-extern.p4
@@ -1,0 +1,233 @@
+/*
+	Copyright 2017 Barefoot Networks, Inc.
+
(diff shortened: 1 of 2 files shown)


jafingerhut:
@hanw Given the resolution of how to attach metadata to clone, resubmit, and recirculated packets, is this example still something we want?  Or should we close this without merging?

hanw:
Yes, closed.

Issue #285, opened 2017-06-06
Question: How undefined is the result of accessing a field in an invalid header?


jafingerhut:
A quote from the P4-16 v1.0.0 spec:
```
The result of reading or writing a field in an invalid header is undefined.
```
Reference: https://github.com/p4lang/p4-spec/blob/master/p4-16/spec/P4-16-spec.mdk#L3287-L3288

I could interpret this in at least a few different ways, and I am sure other interpretations are possible:

(1) If you attempt to read a field in an invalid header, _that value read_ is undefined, but except for that (and any other places in the program that the value's undefined-ness spreads to, e.g. the result of an arithmetic expression using that value), the behavior of the P4 program on that packet is defined.  If you attempt to write to a field in an invalid header, that write is effectively a no-op.

(2) If you attempt to read or write a field in an invalid header, the behavior of the rest of the P4 program on that packet is completely unspecified, even if the program never attempts to access that header or any of its fields again.

(3) The device behavior is completely unspecified from that point an ongoing, even for later arriving packets.  The device must be reset before its behavior becomes defined again.

I was wondering whether (1) is the way people think the statement was intended?  In particular, the part about attempting to write to a field in an invalid header being a no-op?  If so, it seems more precise to replace that sentence of the spec with something that has the same meaning as statement (1).

jnfoster:
From our discussions this fall and winter, interpretation (1) is what was
intended.

jafingerhut:
Do you know whether the statement "If you attempt to write to a field in an invalid header, that write is effectively a no-op." is intended to be true even if the assignment is to a field in an invalid header that is a member of a header_union?

For example, suppose we define:
```
header h1_t {
    bit<8> f1;
}
header h2_t {
    bit<8> f2;
}
header_union u_t {
    h1_t h1;
    h2_t h2;
}
u_t u;
u.h1 = { 8w5 };
// u.h1 is now valid, u.h2 is invalid
u.h2.f2 = 7;
// Is the state of u.h1.f1 now guaranteed to be 5?  Or is it undefined because of the attempted
// assignment to a field of the invalid header u.h2?
```
If so, that means that at least assignments to fields of headers within header unions must be conditional upon the validity of the header in which the the field lies.

[...]

PR #450, opened 2017-10-24
P4_16 spec: Clarify behavior of reads/writes of fields in invalid hea…

jafingerhut:
…ders

This is intended to resolve issue #285

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3286,11 +3286,21 @@ or reading an uninitialized field in a header that is currently valid,
 is that some unspecified value will be used for that field.  This
 unspecified value could differ from one such read to another.
 
-Writing to a field in a currently invalid header must not change any
-state that is currently defined in the system, neither in header
-fields nor anywhere else.  Such a write is allowed to modify state
-whose values are not defined, e.g. the values of fields in headers
-that are currently invalid.
+If a write is performed to a field in a currently invalid header, and
+that header is part of a `header_union`, and a different member header
+of that `header_union` is currently valid, then any or all of that
+valid header's field values may change as a result of the write.  Such
+a write must not change any other state that is currently defined in
+the system, neither in header fields nor anywhere else.
+
+If a write is performed to a field in a currently invalid header, and
+that header is not part of a `header_union`, that write must not
+change any state that is currently defined in the system, neither in
+header fields nor anywhere else.
+
+Either of the kinds of writes to fields in a currently invalid header
+is allowed to modify state whose values are not defined, e.g. the
+values of fields in headers that are currently invalid.
 
 It is expected that some P4~16~ implementations will have options that
 reduce the safety in order to enable more programs to be supported.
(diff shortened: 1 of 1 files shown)


vgurevich:
@jafingerhut ,

The wording about what happens when you write an invalid header field seem to be significantly different from what was discussed before. It's been a while and I can't find any "records of agreement", but I think that the last posting in #285 by @jnfoster  is what he wanted to discuss, but not that final resolution. 

I still believe that accessing a field in an invalid header should be similar to accessing a pointer that we know does point to the heap, but has not been officially returned to us by malloc(). That is: reads will succeed, but you do not know what they will return (we are in agreement here), while writes will succeed too, but you do not know where they will go, except that they will go into the heap (in this case headers/metadata) and not in .text/stack (i.e. not program code, not tables and not externs).

jafingerhut:
One of the options that we discussed was that by default the behavior of writes to invalid header fields was safe, i.e. it didn't corrupt any currently-defined state of the program.

As an option, some implementations are expected to provide options to opt-in to the less safe behavior that you describe.  That is described in the last paragraph of the current proposed change in this pull request.

The current pull request doesn't define precisely what those less-safe options are, or how you would opt-in, e.g. it could be a command line option in your compiler (or the default behavior of your compiler), some kind of whole-program P4_16 annotation, etc.

vgurevich:
@jafingerhut ,

We can certainly mandate safe behavior and make unsafe behavior optional. I did see it in your pull request. What I am afraid of is that the users of many practical (especially hardware) implementations will be forced to (almost always) specify this option, thus defeating our good intentions.

jafingerhut:
I have made more commits to this PR, attempting to address comments made during the 2017-Nov-27 P4_16 language work group meeting.  Here are the notes I took during that meeting describing the suggested changes:

Add clarification that for fields with enum type, an unspecified value
might not be equal to _any_ of the enum values defined within the enum
declaration.  This may violate compiler assumptions in if-then-else or
select expressions intended to be exhaustive.
 
Make explicit that reading a header field could be in a header_union
or a header stack, in addition to the already-mentioned header type.
 
Perhaps: Add one example of a ternary field match in a table with all
bit positions wild-carded, as one way to ignore an unspecified value.
 
 
 
For a header union, specify that invalid header field write must not
change validity of any header_union members.
 
Remove the last added paragraph.  Add a paragraph near the beginning
of the specification, with wording similar to this:
    "It is understood that some implementations may be unable to
    implement the behavior defined here in all cases.  They should
    document where they deviate from this specification."
 
No objections to leaving the 'safe behavior' proposed in the pull
request as the default of the language spec.

jnfoster:
I'm happy.

jafingerhut:
@jnfoster - I saw email about a comment you made about a paragraph that was in my first proposed PR, but was removed in a later commit at the suggestion of the WG.  Hopefully you are able to see the latest proposed changes in Github without that paragraph.

jnfoster:
Yes, I'm happy now. 

Thanks! (And sorry for accidentally closing the PR; best not to let absent minded professors around GitHub repos ;-)

mbudiu-vmw:
Why haven't we already merged this?

jnfoster:
Agree. This was discussed and approved by the LDWG.

PR #451, opened 2017-11-01
PSA add initial values of metadata, and details of packet contents

jafingerhut:


p4-16/psa/PSA.mdk
```diff
@@ -1748,3 +1748,357 @@ types, as the P4 program author wishes.
...
+~
+
+For egress packets, the choice of egress port, or the port to the CPU,
+is made by the immediately previous processing step (ingress for NU,
+NM, or CI2E packets, egress for CE2E packets), and egress processing
+cannot change that choice to a different port.  It can change the
+packet to RECIRC instead, or drop the packet.
+
+A single packet received by a PSA system from a port can result in 0,
+1, or many packets going out, all under control of the PSA program.
```
cc10512: Nit: P4 program

p4-16/psa/PSA.mdk
```diff
@@ -1748,3 +1748,357 @@ types, as the P4 program author wishes.
...
+For example, a single packet received from a port could cause all of
+the following to occur, if the PSA program so directed it:
+
++ The original packet received as NFP from port 2.  Ingress processing
+  creates a CI2E clone destined for the CPU port (copy 1), and a
+  multicast NM packet to multicast group 18, which is configured in
+  the PacketReplicationEngine to have copies made to ports 5 (copy 2)
+  and 7 (copy 3).
++ Copy 1 performs egress processing, which does not create a CE2E
```
cc10512: Readability: do we need to mention what it does not do? if so, i would move it after explaining what the copy does. E.g.: Copy 1 performs egress processing and sends the packet on path NTCPU to the CPU port. It does not create a CE2E clone.

p4-16/psa/PSA.mdk
```diff
@@ -1748,3 +1748,357 @@ types, as the P4 program author wishes.
...
+implementation is free to perform the steps above in many possible
+orders.
+
+There is no mechanism defined in PSA to prevent a single received
+packet from creating packets that continue to recirculate, resubmit,
+or clone from egress to egress indefinitely.  This can be prevented by
+suitable testing of your PSA program, and/or creating in your PSA
+program a "recirculate limit" metadata field that is carried along
+with copies of a packet, similar to the IPv4 Time To Live header
+field.
```
cc10512: I believe we should allow for target implementations to also limit the number of resubmit/recirc.

jafingerhut: Something like this?

"A PSA implementation may drop resubmitted, recirculated, or egress-to-egress clone packets after an implementation-specific maximum number of times of the same original packet.  If so, it should maintain counters of packets dropped for these reasons, and preferably record some debug information about the first few packets dropped for these reasons (perhaps only 1)."

p4-16/psa/PSA.mdk
jafingerhut: I think I wrote "PSA program" after noticing at least one other place in the document that uses that phrase.  In some cases, when saying that something is or is not allowed in a PSA program, it seems more precise than saying P4 program -- we don't want the PSA spec to limit what programs people expect to be able to write in other architectures.

I'm fine either way, and we can do a global search and replace "PSA program" -> "P4 program" if you prefer.

p4-16/psa/PSA.mdk
jafingerhut: I suppose we could leave out the "does not make a clone" phrases here, with the assumption that none is made if it doesn't say anything about a clone.

p4-16/psa/PSA.mdk
jafingerhut: Made the replacements in the new text of this PR.

p4-16/psa/PSA.mdk
jafingerhut: Done.

p4-16/psa/PSA.mdk
jafingerhut: I added the text of the previous comment.

cc10512: Thanks, sounds good.

p4-16/psa/PSA.mdk
cc10512: I think we should use P4 program and PSA implementation consistently throughout the document. If we agree, we can have these changes done during the editorial phase as well.

PR #452, opened 2017-11-01
Revised resubmit/clone/recirculate sections

hanw:


p4-16/psa/PSA.mdk
```diff
@@ -544,146 +546,168 @@ no payload.
...
+
+TBD: does PSA specify that a packet can be resubmitted at most once?
+
+PSA specifies that the resubmit operation can only be used in the
+ingress pipeline. The egress pipeline cannot resubmit packets.
+
+Packet resubmission is intended to increase the capacity and
+ flexibility of the packet processing pipeline. For example, because
+ the same packet is processed by the ingress pipeline multiple times,
+ it effectively increase the amount of operations on the packet by N
```
cc10512: , where N the number of times the packet is resubmitted.

p4-16/psa/PSA.mdk
```diff
@@ -544,146 +546,168 @@ no payload.
...
+used in the second pass.
+
+A PSA implementation provides the `resubmit` extern to specify the
+attached metadata. The `resubmit` extern provides a `add_metadata`
+method which accepts metadata with generic type `T`. In a PSA
+implementation, the metadata is prepended to the resubmitted packet.
+It is the responsibility of the programmer to parse the resubmitted
+packet correctly to extract the attached metadata. The attached
+metadata can be of type `header`, `header stack`, `header union` or
+`struct` of the above types. Invoking the `add_metadata` method
```
cc10512: how does this work with the serialization ideas? do we care to restrict to serializable data only? The scenario I'm thinking about is that the final packet after processing the resubmitted data might go out on the wire with that data as part of the header.

p4-16/psa/PSA.mdk
```diff
@@ -544,146 +546,168 @@ no payload.
...
+[INCLUDE=psa.p4:Resubmit_extern]
+
+need example
+
 ```
 
+### Packet Recirculation { #sec-recirculate }
+
+Packet recirculation sends the deparsed packet at the end of egress
+pipeline to the ingress parser. Unlikely the previous two mechanisms,
```
cc10512: Nit: unlikely -> unlike

p4-16/psa/PSA.mdk
hanw: I am going to rewrite this paragraph after today's discussion on #453. If the proposal in #453 is accepted, then we are not restricted to resubmit serializable data (header, header stack, header union, struct), but also enum, error, etc. The API to clone will be `pack` instead of `add_metadata`. If the user wants to send out enum, he/she has to pick an encoding for the enum/error, and handle the encoding at the receiving end.

p4-16/psa/PSA.mdk
```diff
@@ -544,146 +546,168 @@ no payload.
...
+which port to send the cloned packet to after egress processing. The
+`clone_port` could be any addressable port in the pipeline. For
+instance, it can be a port to control CPU or a port to the next hop
+router.
+
+PRE can optionally attach metadata to the cloned packet. The extra
+metadata is specified with the `clone` extern using the `add_metadata`
+method. The `add_metadata` method accepts a generic type `T`.
+
+TBD: if we decide to use pure metadata, then this paragraph needs to
```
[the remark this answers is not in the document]
hanw: This is referring to the example in #448, where there is no extern for clone metadata. A data type is shared between the deparser and parser, which can be written at the deparser, and read at the parser. #453 by Andy solved the same problem with an extern and `pack` method, which I liked.

PR #454, opened 2017-11-01
revise digest section and API

hanw:


p4-16/psa/examples/psa-example-digest.p4
```diff
@@ -0,0 +1,216 @@
...
+
+control IngressDeparserImpl(packet_out packet,
+                            digest_out packed_meta,
+                            inout headers hdr,
+                            in metadata meta,
+                            in psa_ingress_output_metadata_t istd)
+{
+    CommonDeparserImpl() common_deparser;
+    apply {
+        if (meta.digest_id == 0) {
```
vgurevich: The main problem here is code safety. Let's extend this example for several values of meta.digest_id? What's the mechanism that ensures that (a) we pack exactly what we want for the right digest_id and (b) that the control plane unpacks exactly the same thing?

I think we need to compare this to P4_14, where each field_list is a named object and thus each digest gets a separate API, based on the field_list that was used. Here it is all by number instead and thus is very error-prone...

cc10512: @vgurevich I believe this is the best PSA can do at the moment until we sort out field references in the language. We discussed that at the last LDWG and it is a topic that shall be addressed in the near future.

p4-16/psa/PSA.mdk
```diff
@@ -1732,6 +1732,43 @@ program could discard the most significant bits of the timestamp so
...
+contains two pieces of information: the receiver identifier and the
+digest message. The receiver identifier is an opaque id used by the
+control plane to route the digest message to the right receiver. The
+digest message can contain any from the dataplane.
+
+In PSA, digest is performed at the end of the ingress pipeline in the
+deparser. The digest extern provides a `pack` method to specify the
+content in the 
+
+The compiler decides the best serializatoin format to send the data
```
cc10512: typo: serializatoin -> serialization

p4-16/psa/PSA.mdk
```diff
@@ -1732,6 +1732,43 @@ program could discard the most significant bits of the timestamp so
...
+On the control plane side, the control plane API calls the `unpack`
+method, and pass in a reference to the data object to be read.
+
+The `unpack` method return 0 if succeed. The field names in the
+control plane API is derived from the field names in the dataplane
+`pack` method.
+
+#### Control Plane
+
+The control plane should provide a
```
cc10512: Unfinished sentence.

PR #455, opened 2017-11-01
move PRE and BQE declaration from ingress/egress to top level package

hanw:
I am feeling itchy about having to write down `PacketReplicationEngine pre` every time I create an ingress block. I hope we can move them to the top level package definition, where they belong. Ideally, users shouldn't even need to create an instance of PRE, but I haven't figured out how to do that.

### DIFF
--- a/p4-16/psa/psa.p4
+++ b/p4-16/psa/psa.p4
@@ -592,7 +592,6 @@ parser IngressParser<H, M>(packet_in buffer,
                            out psa_parser_output_metadata_t ostd);
 
 control Ingress<H, M>(inout H hdr, inout M user_meta,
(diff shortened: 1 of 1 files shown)


jafingerhut:
Looks reasonable to me, given that there are no methods for those objects callable from P4 code.

jafingerhut:
@cc10512 OK to merge this one in, then?

cc10512:
Yes, looks good to me.

hanw:
Need to update all examples the make sure they still compiles?

jafingerhut:
Yes, I can do that separately, unless someone else beats me to it.

PR #459, opened 2017-11-02
Update PSA example program for counters for latest psa.p4

jafingerhut:


jafingerhut:
@hanw @cc10512 Before updating all of the example PSA programs, I wanted to run these suggested updates to one example program by your first, since the rest will likely be copy and paste from this one.

In particular, look at the instantiation of PSA_Switch near the end of this proposed change to the program psa-example-counters.p4 of this PR.  This is syntactically correct according to p4test, but requires adding a 0-argument constructor for PacketReplicationEngine and BufferingQueueingEngine extern objects in psa.p4.  Is there another way that avoids the need for an explicit construct you were thinking of instead of this?

hanw:
We can provide a macro in psa.p4
```
#define PSA_SWITCH(x, y) PSA_Switch(x, PacketReplicationEngine(), y, BufferingQueueingEngine())
```

and in user program they can write
```
PSA_SWITCH(ip, ep) main;
```

hanw:
I could not figure out another way in p4.

jafingerhut:
@hanw @cc10512 I added the convenience macro that Han described, and used it in the example.  Calin, if you are good with this, I can update the rest of the example PSA programs using this updated psa-example-counters.p4 as a pattern.

PR #464, opened 2017-11-03
PSA editing round 1

jafingerhut:


jafingerhut:
Some notes about things I have changed here, in case it helps reviewing.  These are in the commit notes for each commit, copied here for your convenience:

The changes to PSA.mdk here are:

+ Removing things that were repeated in multiple places

+ Changing things written earlier where a different decision was made
  later.

+ Deleted some TBDs that I believe were resolved, with additional text
  answering the issue where appropriate.

+ Added at least one new TBD I had not thought of before on clone
  payload truncation.

+ Add a few cross references to make it easier for the reader to find
  more details.

+ Added a note that an implementation can drop a packet after ingress,
  not under control of the P4 program, due to factors such as packet
  buffer being full.

+ Mention that PORT_CPU is a valid port to which ingress may send a
  packet, but PORT_RECIRCULATE is not.

+ Add a similar note about a PSA implementation being allowed to drop
  packets outside of a P4 program's control after egress processing is
  complete, similar to the note describing this behavior after ingress
  processing is complete.

Changes to file psa.p4 are:

+ Add new InstanceType_t enum names, for more precise indication to P4
  programs which of the paths a packet took to get where it is now.

+ Delete some comments that were obsolete, or are now answered in one
  of the recently added packet path details sections.

p4-16/psa/PSA.mdk
```diff
@@ -83,10 +83,10 @@ Table [#table-extern-usage]).
...
 
 A programmer targeting the PSA is required to instantiate objects for
-the programmable blocks that conform to these APIs. Note that the
-programmable block APIs are templatized on user defined headers and
-metadata. In PSA, the user can define a single metadata type for all
-controls.
+the programmable blocks that conform to these APIs (see section
+[#sec-programmable-blocks]).  Note that the programmable block APIs
+are templatized on user defined headers and metadata. In PSA, the user
+can define a single metadata type for all controls.
```
jafingerhut: Only change in this paragraph is adding the cross-reference.  Sorry, I re-formed the paragraph text and this diff doesn't make it obvious that everything else remained the same.

p4-16/psa/PSA.mdk
```diff
@@ -176,21 +176,21 @@ grouped together by similar values of "Processed next by".
...
+|              | from ingress to egress      |         | plus at most one of a |
 |--------------|-------------|               |                    |
-| NM           | normal multicast-replicated | egress  | one of a RECIRC,     |
-|              | packet from ingress to egress |       | NTP, or NTCPU packet |
+| NM           | normal multicast-replicated | egress  | RECIRC, NTP, or NTCPU |
+|              | packet from ingress to egress |       | packet. |
 |--------------|-------------|               |                    |
-| CI2E         | clone from ingress to egress |        |          |
+| CI2E         | clone from ingress to egress |        | See section [#sec-after-egress] for details. |
 |--------------|-------------|               |                    |
```
jafingerhut: Same for this table -- nothing new except added a cross-reference.

p4-16/psa/PSA.mdk
```diff
@@ -398,7 +400,17 @@ The function `platform_port_valid()` mentioned below takes a value of
 type `PortId_t`, returning `true` only when the value represents an
 output port for the implementation.  It is expected that for some PSA
 implementations there will be bit patterns for a value of type
-`PortId_t` that do not correspond to any port.
+`PortId_t` that do not correspond to any port.  This function returns
+true for `PORT_CPU`, but false for `PORT_RECIRCULATE`.
+`PORT_RECIRCULATE` is only intended as a value for `ingress_port` when
+ingress processing begins for a recirculated packet (see Table
+[#ingress-initial-values].  See section [#sec-after-egress] for how P4
+programs cause a packet to be recirculated.
```
jafingerhut: The additions above are worth reading and thinking about to see if they sound correct.

p4-16/psa/PSA.mdk
```diff
@@ -433,30 +445,23 @@ assist P4 developers in debugging their code.
...
-                ostd.clone_class_of_service, after which it will start
-                egress processing with egress_port=ostd.clone_port (which
-                might be PORT_CPU to send the clone to the control CPU).
-            The contents of the packet will be the same as the packet
-                as it began ingress parsing, prepended with any headers
-                created by clone.emit() calls in the ingress deparser
-                apply block.
+            create a clone of the packet and send it to the packet
+                buffer with class of service ostd.clone_class_of_service,
+                after which it will start egress processing.
```
jafingerhut: The details I removed here are stated in the new 'initial values at beginning of ingress' section.  I think they are easier to find and refer to there than in this pseudocode.

p4-16/psa/PSA.mdk
```diff
@@ -433,30 +445,23 @@ assist P4 developers in debugging their code.
...
+    // Any clone created above is unaffected by the code below.
     if (ostd.drop) {
         drop the packet
         return;   // Do not continue below.
     }
     if (ostd.resubmit) {
         resubmit the packet, i.e. it will go back to starting with the
             ingress parser;
-        // TBD: Specify if anything is different about the resubmitted
-        // packet vs. other copies that might be made below.
```
jafingerhut: This TBD is done in the new 'initial values at start of ingress' section.

p4-16/psa/PSA.mdk
```diff
@@ -467,19 +472,21 @@ assist P4 developers in debugging their code.
...
-contents of the packet will be.
+TBD: Should it be possible to truncate a cloned packet differently
+than the normal packet that goes out?
 
-TBD: Should it be possible to truncate a cloned or resubmitted packet
-differently than the normal packet that goes out?
+TBD: The definition of `ostd.truncate_payload_bytes` refers to payload
+bytes.  This makes sense for normal packets passing through the
+ingress deparser, but what does it mean for clone packets, where the
+packet is supposed to be the original pre-IngressParser packet.
```
jafingerhut: If you know a good answer for this, let me know.  One possibility: have a separate metadata value output by ingress control that specifies the truncation length for CLONE_I2E packets independently from the truncation length of normal packets.

p4-16/psa/PSA.mdk
```diff
@@ -489,23 +496,35 @@ copy going to the control CPU to have a software-defined header
...
+Whenever the pseudocode above indicates that a packet should be sent
+to a particular place, a PSA implementation may under some
+circumstances instead drop the packet.  For example, the packet buffer
+may be too low on available space for storing new packets, or some
+other congestion control mechanism such as RED (Random Early
+Detection) or AFD (Approximate Fair Dropping) may select the packet
+for dropping.  It is recommended that an implementation maintain
+counters of packets dropped, preferably with separate counters for as
+many different reasons as the implementation has for dropping packets
+outside the control of the P4 program.
```
jafingerhut: Good to double-check me on this addition.  I believe it must be true, since many traffic patterns must cause queues to build up.

p4-16/psa/PSA.mdk
```diff
@@ -756,45 +778,43 @@ the contents of several metadata fields in the struct
             create a copy of the packet and send it to the packet
                 buffer with class of service specified by
                 ostd.clone_class_of_service, after which it will start
-                egress processing with egress_port=ostd.clone_port (which
-                might be PORT_CPU to send the clone to the control CPU).
-            The contents of the packet will be the same as the packet
-                as it comes out of the egress deparser, prepended with any
-                headers created by clone.emit() calls in the egress
-                deparser apply block.
+                egress processing.
```
jafingerhut: As earlier, the details removed are are I think better represented in the recently added 'initial values at beginning of egress processing' section.

p4-16/psa/PSA.mdk
```diff
@@ -756,45 +778,43 @@ the contents of several metadata fields in the struct
...
-differently than the normal packet that goes out?
+As for the handling of a packet after ingress processing, a PSA
+implementation may drop a packet after egress processing, even if the
+pseudocode above says that a packet will be sent.  For example, you
+may attempt to clone a packet after egress when the packet buffer is
+too full, or you may attempt to recirculate a packet when the ingress
+pipeline is busy handling other packets.  It is recommended that an
+implementation maintain counters of packets dropped, preferably with
+separate counters for as many different reasons as the implementation
+has for dropping packets outside the control of the P4 program.
```
jafingerhut: Similar comment as one I called out above -- there must be cases where a packet can be dropped even if the P4 program didn't request it.

p4-16/psa/PSA.mdk
```diff
@@ -835,6 +853,12 @@ For Ethernet ports, any padding required to get the packet up to the
 minimum frame size required is done by the implementation, as well as
 calculation of and appending the Ethernet frame CRC.
 
+It is expected that typical P4 programs will have explicit checks to
+avoid sending packets larger than a port's maximum frame size.  A
+typical implementation will drop frames larger than this maximum
+supported size.  It is recommended that they maintain error counters
+for such dropped frames.
```
jafingerhut: Again, seems obvious that something like this must be the case, but comment if you disagree.

p4-16/psa/psa.p4
```diff
@@ -40,10 +40,9 @@ typedef bit<16> EgressInstance_t;
 typedef bit<48> Timestamp_t;
 typedef error   ParserError_t;
 
+const   PortId_t         PORT_RECIRCULATE = 254;
 const   PortId_t         PORT_CPU = 255;
 
-// typedef bit<unspecified> InstanceType_t;
-// const   InstanceType_t   INSTANCE_NORMAL = unspecified;
```
jafingerhut: These lines look obsolete to me, as both things are defined differently later in psa.p4 now.

p4-16/psa/psa.p4
```diff
@@ -55,17 +54,19 @@ typedef bit<unspecified> PacketLength_t;
 typedef bit<unspecified> EgressInstance_t;
 typedef bit<unspecified> Timestamp_t;
 
+const   PortId_t         PORT_RECIRCULATE = unspecified;
```
jafingerhut: New definition needed if we want this value to be used as the ingress_port at the beginning of ingress, for recirculated packets, as I think Vladimir suggested.

p4-16/psa/psa.p4
```diff
@@ -55,17 +54,19 @@ typedef bit<unspecified> PacketLength_t;
...
 
 // BEGIN:Metadata_types
 enum InstanceType_t {
-    NORMAL,     /// Packet is "normal", i.e. none of the other cases below
-    CLONE,      /// Packet was created via a clone operation
+    NORMAL,     /// Packet received by ingress that is none of the cases below.
+    NORMAL_UNICAST,   /// Packet received by egress that is none of the cases below, and unicast
+    NORMAL_MULTICAST, /// Packet received by egress that is none of the cases below, and multicast
+    CLONE_I2E,  /// Packet created via a clone operation in ingress, destined for egress
+    CLONE_E2E,  /// Packet created via a clone operation in egress, destined for egress
```
jafingerhut: Having all of these values makes distinguishing these cases at the beginning of ingress or egress processing possible.

p4-16/psa/PSA.mdk
```diff
@@ -338,46 +346,43 @@ packet contents and metadata when a packet begins ingress processing.
...
 
-For Ethernet ports, packet_in contains the Ethernet frame starting
-with the Ethernet header.  It does not include the Ethernet frame CRC.
-TBD whether the payload is always the minimum of 46 bytes (64 byte
-minimum Ethernet frame size, minus 14 bytes of header, minus 4 bytes
-of CRC), or whether an implementation is allowed to leave some of
-those bytes out.
+For Ethernet ports, `packet_in` for FP and NFCPU path packets contains
+the Ethernet frame starting with the Ethernet header.  It does not
+include the Ethernet frame CRC.  TBD whether the payload is always the
```
cc10512: Given that we're planning for a programmable switch and even though most applications are likely to include Ethernet, I don't see a strong need to constrain the minimum packet size for PSA. It is likely to be target dependent in any case. And if one wants to build a custom protocol in which the entire packet consists of headers, we should not prevent it. There are target limits on how big a header can be as well.

p4-16/psa/PSA.mdk
cc10512: Since it is used only in the pseudo-code, I think we're ok, even though we still consider PORT_RECIRCULATE a port, so it will give readers a pause.

p4-16/psa/PSA.mdk
cc10512: Shall we just disallow truncation of cloned packets?

p4-16/psa/PSA.mdk
```diff
@@ -489,23 +499,35 @@ copy going to the control CPU to have a software-defined header
 (defined in the P4 program) that is different than any headers on the
 packet copies going to the Egress control block.
 
+Whenever the pseudocode above indicates that a packet should be sent
+to a particular place, a PSA implementation may under some
```
cc10512: s/place/port?

p4-16/psa/PSA.mdk
cc10512: It's a fair point -- no actual device has infinite resources to buffer all packets.

p4-16/psa/PSA.mdk
jafingerhut: I am guessing PSA implementations will be free to drop resubmitted and recirculated packets, too, which at least right now in the PSA spec aren't described as ports.  Maybe s/to a particular place/on a particular packet path/ ?

p4-16/psa/PSA.mdk
jafingerhut: I would think that perhaps one common situation in which truncating packets is desired is when sending a clone to the CPU -- the CPU software might often only care about the headers, not the payload contents.

jafingerhut: One way to specify this would be to have a clone_truncate_length metadata field, like we have other clone-specific metadata fields.  For ingress, it would specify the maximum number of packet bytes to be cloned, starting at the beginning of the first header byte, not payload bytes.  For egress, it would specify the maximum number of payload bytes, but independently from packet truncation of normal packets, so a P4 programmer could make a clone that is truncated, and a normal packet that is not, in the same egress processing pass of a single packet.

cc10512: Wouldn't it be easier to create a digest?

jafingerhut: I have no problem saying in the PSA that truncation is not supported for clone packets, and suggesting the use of the Digest extern if/when someone wishes for such functionality.  If someone wants truncation of cloned packets, we can bring it up in the PSA WG in the future.

PR #465, opened 2017-11-03
PSA Move example implementation of Counter extern to an appendix

jafingerhut:


jafingerhut:
This is a pretty straightforward change, just moving some text about an example implementation of the Counter extern into an appendix.

PR #473, opened 2017-11-07
PSA small fixes

jafingerhut:


jafingerhut:
@cc10512 Besides these small changes, note that some of the text in these short sections on PRE and BQE are currently inaccurate -- they were written when PRE and BQE externs had no constructors, but at least at the moment they do.  I recall you making a comment that you would like to find a way to eliminate those constructors, so I'm OK with leaving these sections inconsistent with psa.p4, if you think we might find a way to make this text correct again.

cc10512:
@jafingerhut you and I both don't like inconsistencies, and it would be fantastic to get rid of them. AFAIK, the main reason we added the constructors was because package (as opposed to control) does not seem to be able to instantiate these externs as `packet_in` is instantiated by the compiler. That is because they are not part of the language, they are part of the architecture, but really we have no mechanism to tell the compiler these should be initialized. One possible solution, I believe is to add an annotation on these externs and say they should be automatically instantiated, and given a name so that the control plane can address them. Then we can remove them from the package definition and get rid of the pesky macro that hides their instantiation too! With that, the sections are consistent and the PSA package definition is a bit cleaner. I would like to hear what @hanw and @mbudiu-vmw think about such a proposal.

hanw:
Another option is to specify a higher level 'main' package that is provided by the compiler implicitly which instantiates the PRE() and BQE() and PSA(). the main package in PSA then becomes the user-level 'main' package, which only consists of the programmable blocks. 

For example,
```
package Top (PSA(), PRE(), BQE()) top;

# if keep ingress and egress package
package PSA_Switch(Ingress(), Egress()) main; 
# if not
package PSA_Switch(ingressParser(), ingress(), ingressDeparser(), egressParser(), egress(), egressDeparser()) main;
```

Alternatively, since PRE() and BQE() are only concerning the API WG. We should leave a hole in the top level package for the API WG to fill in their package.

```
package Top (PSA(), API()) top;
package API (PRE(), BQE()) api_main;
```

mbudiu-vmw:
The user must be able to supply all constructor arguments at compile-time to all blocks, including packages. So if BQE has no constructor, the user cannot instantiate it. So it cannot be "made available" to a package.

I think @hanw 's proposal is good.

The core question is: why do you want to expose these engines at all in the data-plane program?

cc10512:
@mbudiu-vmw the only reason they are in the arch file is because there are control-plane APIs that configure them in such a way that data-plane functionality is affected. That calls for them to be part of the architecture, otherwise the behavior is not fully defined.

@hanw, I read again the spec and we mentioned at one point that these are like `packet_in`. `packet_in` and `packet_out` do not have constructors. While your proposal is quite elegant, it does occur to me that we are sometimes considering the order of arguments important, sometimes not. And since we do not really need to instantiate these externs in the user program, I am still looking for a mechanism in which we can eliminate the constructors :)

mbudiu-vmw:
Data-plane run-time values are not instantiated by the user, they are instantiated by the architecture.
That's why packet_in does not need a constructor.

It is useful to have a constructor, it indicates that some kind of resources are being allocated statically.

hanw:
As far as I can see, the order of the arguments in the package is important for two reasons: it helps to check the package instance has the right signature; it implicitly represents the logical order of the packet processing (from left to right). Frankly, I think relying on the order of arguments to represent packet processing order is not a wise idea. If we need a notion of packet processing order, we should make it explicit in the architecture or user program using the apply() mechanism, not by inferring from the ordering of the arguments. If we accept that the ordering of arguments does not have meanings on how packets are processed, we can have the following definition for the top-level block (If we can make the constructor parameter to work on `package`).

```
package Top(PSA())(PacketReplicationEngine pre, BufferQueueingEngine bqe) top;
```

mbudiu-vmw:
We have a proposal to have named arguments in the next version of the language.
So just wait a little bit.

PR #474, opened 2017-11-07
fix clone example

hanw:


jafingerhut:
Those warnings exist when compiling all of the current examples.  Annoying, but I suspect it would be more annoying what we would have to change in the examples to prevent the warnings from occurring.

hanw:
once we have a numbering system for all the warnings, we can add a flag --suppress-warnings WARN_ID to the compiler.

cc10512:
It's fine to have these warnings, since clearly we're illustrating just one aspect. It's already great to have examples that compile.

Issue #477, opened 2017-11-07
PSA Truncation of different kinds of packets


jafingerhut:
Should it be possible to truncate ingress-to-egress clone packets?  If so, probably having a separate clone_truncate_length_bytes metadata field in ingress output metadata would be good to have, because specifying a clone truncation length independently of a normal packet truncation length seems useful in some cases, and because the ingress-to-egress clone is based on a pre-IngressParser packet, so the truncate_payload_bytes value wouldn't start at the same place (the beginning of the unparsed portion of the packet) as it would for normal packets.

Alternately, explicitly say that truncation of ingress-to-egress clone packets is not supported in PSA, and make that clear in the post-ingress pseudocode and any other places it might be mentioned.

Should it be possible to truncate egress-to-egress clone packets?  The issue of different starting point as above does not exist here.  Still if it is supported, it could be useful to enable truncation of cloned packets to a different length than normal packets, if it is supported.

Otherwise, again make it clear that this kind of truncation is not supported.

jafingerhut:
Should be able to close this issue when this PR is approved and merged in: https://github.com/p4lang/p4-spec/pull/538

jafingerhut:
PR https://github.com/p4lang/p4-spec/pull/538 has been merged in.  Closing issue.

Issue #466, opened 2017-11-03
PSA Document interaction between direct counters/meters and action profile/selector


jafingerhut:
Do we allow a table with an action profile or selector to have direct counters and/or meters?

If so, how many direct counters does a table using an action profile have?  That is, what is each counter value associated with in a 1-to-1 relationship (if the table has no action profile or action selector, the 1-to-1 relationship is between counter values and table entries.)

When an action is executed that updates a direct counter for a table using an action profile, which counter does it update?

Same questions as above, but for tables using an action selector.

jafingerhut:
@samar-abdi @vgurevich Just calling out your names to draw your attention to this issue.  I worry that there are differences in how people implement action profiles and action selectors that the current most accurate answer is "it depends on how you implement action profiles and action selectors in your target".

vgurevich:
@jafingerhut ,

The question you pose in the second paragraph is the most important one: do direct counters/meters correspond 1-1 to match entries (represented by the table itself) or 1-1 to the action entries? In tables without action profiles, there is a 1-1 correspondence between match and action entries, so that's no problem, but in tables with action profiles there is many-to-1 relationship instead.

Given that incrementing the counter or executing a meter is clearly an action, I'd say that most probably direct counters/meters should have 1-1 correspondence to the action entries and therefore, if a table has an action profile, then they should become additional attributes (constructor arguments) to this action profile. However, I can also see an argument of treating these resources separately, like an additional specialized action, associated 1-1 with a match entry. In that case, they should remain table properties.

Personally, I feel that unless we can make a very clear choice between these two designs, it might be prudent to table this issue and simply do not allow mixing action profiles (that already are almost at the boundary of permissible in my opinion) and direct resources. We can always allow them later should the need arise.

samar-abdi:
A direct counter needs to be explicitly incremented and a direct meter explicitly read from inside the action body. There is no implicit modification of direct counter or execution of direct meter. At least that is how these resources are implemented in v1model. As for the addressing of the direct meter or counter, it is the match key of the table that instantiates them. 

Here is an example:

table foo {
  key = {
    h.f1 : lpm;
    h.f2 : selector;
  }
  actions  = {a; b;}
  counters = c;
  implementation = as;
}

action a(in PortId_t p) {
  ostd.egress_port = p;
  ostd.drop = false;
  c.count();
}

action b() {
  ostd.drop = true;
}

action_selector(HashAlgorithm.crc16, 1024, 10) as;

------------
[...]

PR #480, opened 2017-11-07
PSA Remove most TBDs in document

jafingerhut:
In most comments, Github issues have been created to replace them, but
some are simply deleted since no one has raised any concerns about
them.  See my own comments below if you are curious whether a Github issue has been created to replace the removed TBD, or if not, why not.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -392,12 +392,6 @@ through the rest of the parser into the `Ingress`  control block as one
 might normally expect.  This will be left as an option to the user in
 their P4 programs, not required behavior for all P4 programs.
 
-~TBD
-Is this true for RESUB and RECIRC packets, or is it intended that
-there by a way in PSA to specify some collection of fields to be
-preserved with a resub/recirc'ed packet?
-~
-
 ## Behavior of packets after ingress processing is complete {#sec-after-ingress}
 
 The pseudocode below defines where copies of packets will be made
@@ -491,17 +485,6 @@ assist P4 developers in debugging their code.
 
 Note: Truncation is not currently supported for cloned packets. The functionality of a truncated
 cloned packet can be replaced with sending a digest.
-
-~TBD
-
-If it is planned to be possible at the end of ingress to send a
-packet to be replicated via a multicast_group, and also have a copy go
-to the control CPU, give an example showing this case (after showing
-some simpler common cases).  Ideally it should be possible for the
-copy going to the control CPU to have a software-defined header
-(defined in the P4 program) that is different than any headers on the
-packet copies going to the `Egress` control block.
-~
 
 Whenever the pseudocode above indicates that a packet should be sent
 on a particular packet path, a PSA implementation may under some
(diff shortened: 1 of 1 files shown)


p4-16/psa/PSA.mdk
```diff
@@ -392,12 +392,6 @@ through the rest of the parser into the `Ingress`  control block as one
 might normally expect.  This will be left as an option to the user in
 their P4 programs, not required behavior for all P4 programs.
 
-~TBD
-Is this true for RESUB and RECIRC packets, or is it intended that
-there by a way in PSA to specify some collection of fields to be
-preserved with a resub/recirc'ed packet?
-~
-
```
jafingerhut: Removed this one.  It seems like it should be covered by whatever we come up with for passing metadata with packets, which is a prominent issue in the open issues section.

p4-16/psa/PSA.mdk
```diff
@@ -806,11 +789,6 @@ the contents of several metadata fields in the struct
     enqueue one packet for output port istd.egress_port
 ```
 
-~TBD
-Should it be possible to truncate a cloned packet differently
-than the normal packet that goes out?
-~
-
```
jafingerhut: Created issue #477 while deleting this TBD.

p4-16/psa/PSA.mdk
```diff
@@ -912,10 +890,6 @@ which port to send the cloned packet to after egress processing. The
 instance, it can be a port to control CPU or a port to the next hop
 router.
 
-~ TBD
-Decide how to add other metadata to the cloned packet (see [#sec-open-metadata]).
-~
-
```
jafingerhut: It seems like it should be covered by whatever we come up with for passing metadata with packets, which is a prominent issue in the open issues section.

p4-16/psa/PSA.mdk
```diff
@@ -1359,12 +1325,6 @@ inside an action of its owner table.
 The counter value updated by an invocation of `count` is always the
 one associated with the table entry that matched.
 
-~ TBD
-How to describe which counter value is updated for tables with
-action profiles and direct counters?  Or should this combination even
-be allowed?
-~
-
```
jafingerhut: I think Github issue #466 already covers this.

p4-16/psa/PSA.mdk
```diff
@@ -2002,21 +1948,6 @@ synchronization, e.g. via PTP[^PTP] or NTP[^NTP].
...
-~ TBD
- Some time synchronization methods avoid sudden "jumps" by
-temporarily speeding up or slowing down the rate of increase by a
-small percentage, until the desired synchronization is achieved.
-Which ones?  citation?).  Would anyone mind if PSA
-implementations were allowed to do this with their timestamp values?
-~
-
```
jafingerhut: I am OK just deleting these TBDs with questions about timestamps.  No one has made any comment about them, nor complained about the rest of the timestamp section.  Perhaps that is because they've read it and
like it, or perhaps such details are not a concern for them.  Either way, I don't see keeping the TBDs in there helping anything.

PR #481, opened 2017-11-07
PSA Add a table listing P4 table properties added by PSA

jafingerhut:


p4-16/psa/PSA.mdk
```diff
@@ -1095,6 +1095,27 @@ Rationale:
...
+~ TableFigure {#table-table-properties; \
+  caption: Summary of PSA table properties.}
+|-----------------------|------|---------------------------------------|
+| Property name         | Type | See also |
++:----------------------+:-----+:--------------------------------------+
+| `psa_direct_counters` | list of `DirectCounter` instance names | Section [#sec-direct-counter] |
+|-----------------------|----------------------------------------------|
+| `psa_direct_meters` | list of `DirectMeter` instance names | Section [#sec-meters] |
+|-----------------------|----------------------------------------------|
+| `implementation`      | instance name of one `ActionProfile` | Sections [#sec-action-profile], [#sec-action-selector] |
```
vgurevich: Should we consider a different name (e.g. psa_implementation)?

jafingerhut: Updated this PR to change this occurrence of `implementation`, and other occurrences in example code and the action profile/selector text that I could find, to `psa_implementation`.

jafingerhut:
In the P4_14 v1.0.0 spec, I see mentions of an "example" table property called 'implementation', e.g. this one:

"As another example, an implementation property could be used to pass additional information to the compiler back-end."

I don't see anything that says P4_16 requires this table property to do anything, but I could be missing it.

jafingerhut:
Sounds like it might be useful to have a similar table in the P4_16 spec of exactly the list of table properties that are actually required by the language itself?

mbudiu-vmw:
There aren't many: Key, actions and default_action.

vgurevich:
What about size and entries?

mbudiu-vmw:
Size is not part of the spec. I guess that entries is, though. For software switches, which can allocate memory dynamically, size is not needed.

vgurevich:
In that case, I think that 'size' (or psa_size) should become a part of PSA. 

But, frankly, I think size should be in the language spec. SW implementations can ignore it (or use min_size/max_size like in P4_14), but this is a vital parameter for any HW resource allocation and leaving it to a vendor implementation sounds a little risky to me.

jafingerhut:
I believe this PR is consistent with the discussions held during the 2017-Dec-06 PSA WG meeting: All table properties psa_direct_counters, psa_direct_meters, psa_implementation should have psa prefix.  Feel free to comment on it further, but if you have no other comment I think it should be approved and merged in.

For 'size' we discussed that this should be taken up in the P4_16 language WG.  I have created an issue for the P4_16 spec here: https://github.com/p4lang/p4-spec/issues/514

jafingerhut:
@cc10512 Do you think these changes look reasonable to go in?

PR #486, opened 2017-11-10
Change the spec page title.

jklr:


index.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>P4 Language Specification</title>
+    <title>P4 Language Specification and Related Specifications</title>
```
cc10512: I'm obviously not a native speaker, but I would actually think that "P4 Language and Related Specifications" is more appropriate.

PR #488, opened 2017-11-11
Add telemetry specs.

jklr:


jklr:
@cc10512  yes, that's the plan onwards.

PR #490, opened 2017-11-13
Update Telemetry Report Format draft

jklr:
Fixed a broken reference url.


jafingerhut:
@jnfoster Not sure of the protocol here, but perhaps you may welcome the idea of working group chairs approving and merging in changes inside their relevant sub-directories of this repository, without requiring your approval?  At least, that is how Calin Cascaval and I have been operating for the PSA specification, also in this repository.

jnfoster:
Yes. I think we can even be a bit more liberal: each WG can identify a group of key committers who are permitted to review and merge changes within p4-spec.

jklr:
Thanks for the notes on the WG spec review policies.

A bigger question to me is whether it's better to host the specs in each WG specific repo. 
A few things to consider:
- Apps WG is expected to generate more sub-sub-directories, one for each application area.
- Application specs are likely to change quite frequently, more frequently than language/arch/API specs.
- Application spec or document will accompany with reference code and test program hosted in the separate github repo. It can be confusing where to conduct discussions and track issues btw the p4-spec repo vs. the apps repo.

+ @mhira1

jafingerhut:
For issues on language specs vs. PSA so far I have been usually marking PSA-specific issues and commits with "PSA" prefix, just to be able to quickly distinguish them for language-specific issues/PRs.

It seems reasonable to me to introduce a separate repository for P4 applications areas, or even one per application area, if you wish.

cc10512:
Agreed. For that level of complexity, you may want your own sandbox :)

jklr:
Sounds good. How about publishing to p4.org website? Don't we need to host the material in p4-spec?
For application specs, we can post links to the specs hosted in the apps WG repo.

jnfoster:
No you don't need to host it in p4-spec. As long as you either maintain the
current PDFs somewhere accessible, or set up your own GitHub pages and
Travis CI, we can grab the files when we build the website HTML.

PR #499, opened 2017-11-27
Update text for recirc and resubmit

hanw:


p4-16/psa/PSA.mdk
```diff
@@ -950,41 +940,33 @@ used in the second pass. See Section [#sec-open-metadata] for the
...
-deparser.
-
-```
-[INCLUDE=psa.p4:Resubmit_extern]
-
-```
+PRE to enable the resubmission mechanism. If true, the original packet
+is resubmitted with the optional resubmit metadata. If false, the
+resubmission mechanism is disabled and the value assigned to the
+`resubmit_meta` is ignored.
```
jafingerhut: I think the way the proposal for carrying metadata is now, it would be more precise to say "and no assignments to `resubmit_meta` should be performed"

p4-16/psa/PSA.mdk
```diff
@@ -950,41 +940,33 @@ used in the second pass. See Section [#sec-open-metadata] for the
...
-[INCLUDE=psa.p4:Recirculate_extern]
+Packet recirculation is a mechanism to repeat both ingress and egress
+processing on a packet.
 
-```
+Packet recirculation happens at the end of the egress pipeline. When a
+packet is recirculated, the packet finishes both ingress and egress
+processing and re-enters the ingress parser after being deparsed by
+the egress deparser. The `ingress_port` of the recirculated packet is
+set to `RECIRCULATE`. The `packet_path` of the recirculated packet is
```
jafingerhut: At least right now psa.p4 uses the value PORT_RECIRCULATE for the ingress_port of such packets.

p4-16/psa/PSA.mdk
```diff
@@ -950,41 +940,33 @@ used in the second pass. See Section [#sec-open-metadata] for the
...
+changed to `RECIRCULATE`.
 
-Users can leverage the existing `emit` methods in `packet_out` to
-attach additional metadata to the recirculated packet. Upon receiving
-the recirculated packet in ingress parser, the `packet_path` of the
-recirculated packet is set to `RECIRCULATE`. It is the user's
-responsibility to parse the recirculated packet correctly.
+~ TBD
+Whether use two different enums for recirc port and recirc path
+~
```
jafingerhut: recirc path is an enum value in psa.p4 as proposed now in another PR.  PORT_RECIRCULATE is not an enum, but a particular value of type PortId_t.

p4-16/psa/PSA.mdk
```diff
@@ -950,41 +940,33 @@ used in the second pass. See Section [#sec-open-metadata] for the
...
 
+The `recirculate` flag specifies whether a packet should be
+recirculated. If true, the processed packet should be recirculated at
+the end of the egress pipeline. The `egress_port` must be set to the
+dedicated port for recirculation in the pipeline. Some target may
+restrict the `recirculate` flag and `egress_port` to be set in the
```
jafingerhut: Nothing else in the current draft document says anything about having to assign a value to `egress_port` in order to recirculate a packet.  It seems easier to only require the P4 program to assign true to the `recirculate` flag, and then they don't have to do anything else.

Also, the field `egress_port` is an in parameter to the egress control block, so it is illegal in P4_16 to attempt to assign a value to it.

p4-16/psa/PSA.mdk
```diff
@@ -911,36 +897,40 @@ the cloned copy to `clone_port`.
...
+One use case of packet resubmission is to increase the capacity and
+flexibility of the packet processing pipeline. For example, because
+the same packet is processed by the ingress pipeline multiple times,
+it effectively increase the amount of operations on the packet by N
+folds, where N is the number of times the packet is resubmitted.
+
+Another use case is to deploy multiple packet processing algorithms on
+the same packet. For example, the original packet can be parsed,
+modified and resubmitted in the first pass. Then, the resubmitted
+packet can be parsed and modified using a different algorithm.
```
jafingerhut: It seems a little bit misleading to say that the original packet can be "modified" here, doesn't it?  I mean, I know it can modify the packet before resubmitting it, but all such modifications will be lost, true?

p4-16/psa/PSA.mdk
jafingerhut: You mention extra metadata that can be associated with a resubmitted packet, but not for a recirculated packet.  It would be good to mention for both or neither -- for both seems a good choice.

hanw:
Thanks, Andy, I have addressed all the comments.

p4-16/psa/PSA.mdk
```diff
@@ -911,36 +897,41 @@ the cloned copy to `clone_port`.
...
+the original packet. The `ingress_port` of the resubmitted packet is
+the same as the original packet. The `packet_path` of the resubmitted
+packet is changed to `RESUBMIT`.
+
+The ingress parser distinguishes the resubmitted packet from the
+original packet with the `packet_path` field in
+`ingress_parser_intrinsic_metadata_t`. The ingress parser can choose a
+different algorithm to parse the resubmitted packet. Similarly, the
+ingress pipeline can choose to process the resubmitted packet with
+different actions as opposed to the ones used to process the original
```
jafingerhut: Might be nice to mention that not only can you tell the original packet from a resubmitted packet because of `packet_path`, but the user's program can make it possible to distinguish between a packet resubmitted the first time, or second time, etc. by the extra metadata they can associate with the packet.  As far as I can tell, that is the _only_ way to cause a packet to be processed differently on the 2nd or N-th times being resubmitted.

p4-16/psa/PSA.mdk
```diff
@@ -950,41 +941,35 @@ used in the second pass. See Section [#sec-open-metadata] for the
...
+is generated during the first pass through the ingress and egress
+pipeline, and used in the second pass. See Section
+[#sec-open-metadata] for the discussion.
+
+The `recirculate` flag specifies whether a packet should be
+recirculated. If true, the processed packet should be recirculated at
+the end of the egress pipeline with the optional metadata. If false,
+the recirculation mechanism is disabled and no assignments to
+`recirc_meta` should be performed. Some target may restrict the
+`recirculate` flag to be set in the ingress pipeline.
```
jafingerhut: The `recirculate` flag is not visible or accessible in the ingress pipeline in the current psa.p4 definition.  How do you propose a target require the programmer to assign a value to `recirculate` in the ingress pipeline?

jafingerhut: Follow-up comment: If someone wants to control whether a recirculate happens in their ingress code, a way to do that is to store in some user-defined metadata field that is carried from ingress to ingress with a normal packet (if the normal packet path is the relevant one here), and then in egress code use that metadata field value to control whether it assigns the value true to the `recirculate` field.

jafingerhut:
@hanw Sorry, but I had some updates to the cloning section, at least, in my now-merged PR for how to associate metadata with packets, which is why this PR now has conflicts with the latest master.  Do you want me to take a look at merging them?

hanw:
Thanks, Andy!

p4-16/psa/PSA.mdk
```diff
@@ -970,80 +959,84 @@ The partial program below demonstrates how to clone a packet.
...
+
+Another use case is to deploy multiple packet processing algorithms on
+the same packet. For example, the original packet can be parsed and
+resubmitted in the first pass with additional metadata to select one
+of the algorithms. Then, the resubmitted packet can be parsed,
+modified and deparsed using the selected algorithm.
 
 To facilitate communication between the first pass and the second pass
-through the ingress pipeline, the resubmission mechanism supports
+through the ingress pipeline, the resubmission mechanism supportsr
```
hanw: typo?

p4-16/psa/PSA.mdk
```diff
@@ -970,80 +959,84 @@ The partial program below demonstrates how to clone a packet.
...
-the recirculated packet in ingress parser, the `packet_path` of the
-recirculated packet is set to `RECIRCULATE`. It is the user's
-responsibility to parse the recirculated packet correctly.
+Similar to packet resubmission, packet recirculation also supports
+attaching optional metadata with the recirculated packet. The metadata
+is generated during the first pass through the ingress and egress
+pipeline, and used in the second pass. See Section
+[#sec-open-metadata] for the discussion.
 
+The `recirculate` flag specifies whether a packet should be
```
hanw: I can write an example to recriculate without the `recirculate` flag, if it works, then we can remove this paragraph in another PR?

p4-16/psa/PSA.mdk
jafingerhut: Yes.  I did the conflict resolution using Github's web interface, not normal git commands.  I'm happy to fix this in a later PR, unless you find it easy to figure out how to use git to pull in those changes, and add another commit on top.

p4-16/psa/PSA.mdk
jafingerhut: Sounds good.

jafingerhut:
@cc10512 I've approved this.  Not sure if you want to take a pass through it before it is merged in?  I expect there will be further edits in these sections even after this is merged, to make everything consistent.

PR #500, opened 2017-11-27
v1.0.5

jnfoster:
Initialize fresh directory for v1.0.5 maintenance update to P4_14.

To make these changes easy to review, this commit contains no other modifications other than copying the v1.0.4 directory to the v1.0.5 directory.


jafingerhut:
I have double-checked that this PR copies all files from p4-14/v1.0.4 directory to a new p4-14/v1.0.5 directory, with no changes to the contents of those files.  Also that no other files are changed.  As long as it is not auto-published in a place that you do not want it to be when merged into master, it seems good to go.

PR #501, opened 2017-11-27
 Harmonize P4_14 and P4_16 semantics of header stacks

jnfoster:
This PR takes a first step toward the "P4_16 v1.0.0 is right" option identified by @jafingerhut.

This PR will be easier to review after #500 is merged.

### DIFF
--- a/p4-14/v1.0.5/tex/p4.pt
+++ b/p4-14/v1.0.5/tex/p4.pt
@@ -704,19 +704,25 @@ depend on the context.
 To help programers write parsers for header stacks, P4 provides two
 keywords, \texttt{next} and \texttt{last}, that may be used to refer
 to header instances within a stack. These references are advanced
-automatically as the parser invokes \texttt{extract} in the parser. It
-is illegal to use these keywords outside of a parser.
-    
-More specifically, the keyword \texttt{next} refers to the header
+automatically as the parser invokes \texttt{extract} in the parser.
+
+More precisely, the keyword \texttt{next} refers to the header
 instance at index \texttt{0} initially, and is advanced each time that
-\texttt{extract} is invoked on an instance within the stack. If
+\texttt{extract} is invoked on any instance within the stack. If
 \texttt{extract} has been invoked more than $N$ times, where $N$ is
 the size of the stack, then evaluating \texttt{next} results in a
 \texttt{p4_pe_index_out_of_bounds} exception. The \texttt{last}
 keyword refers to the instance referred to by \texttt{next} prior to
 the last invocation of \texttt{extract}, if it exists, and otherwise
-results in \texttt{p4_pe_index_out_of_bounds} exception.
-
+results in \texttt{p4_pe_index_out_of_bounds} exception. It is illegal
+to use \texttt{next} and \texttt{last} outside of a parser.
+    
+Note that while it is legal to extract to a constant index in a header
+stack and extract to the element referred by \texttt{next}, it is not
+recommended. It will likely be difficult for programmers to keep track
+of which element in the header stack \texttt{next} refers to when
+operations are combined in this way.
+    
 \SUBSECTION{Field Lists}{fieldlists}
 
 In many cases, it is convenient to specify a sequence of fields. For example, 
@@ -1797,10 +1803,9 @@ Add a header to the packet's Parsed Representation
 \paramdoc{header_instance}{(HDR) The name of the header instance to add.}
 }
 { % Description
-If the \texttt{header_instance} is not an element in a header stack, the indicated 
-header instance is set valid. If the header instance was invalid, all its 
-fields are initialized to 0. If the header instance is already valid, it is 
-not changed.
+The indicated  header instance is set valid. If the header instance was invalid, all its 
+fields are initialized to 0, and if the header instance was already valid, it is 
+not changed. 
 }
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -1831,10 +1836,9 @@ Mark a header instance as invalid.
 \paramdoc{header_instance}{(HDR) The name of the header instance to remove.}
 }
 { % Description
-If the \texttt{header_instance} is not an element in a header stack, then the indicated 
-header instance is marked invalid. It will not be available for matching in 
-subsequent \matchaction stages. The header will not be serialized on egress. 
-All field values in the header instance become uninitialized.
+The indicated  header instance is marked invalid. It will not be available for
+matching in subsequent \matchaction stages. The header will not be serialized
+on egress. All field values in the header instance become uninitialized.
 }
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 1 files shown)


p4-14/v1.0.5/tex/p4.pt
```diff
@@ -705,6 +700,22 @@ References at run time to a header instance (or one of its fields) which is
...
+    
+To help programers write parsers for header stacks, P4 provides two
+keywords, \texttt{next} and \texttt{last}, that may be used to refer
+to header instances within a stack. These references are advanced
+automatically as the parser invokes \texttt{extract} in the parser. It
+is illegal to use these keywords outside of a parser.
+    
+More specifically, the keyword \texttt{next} refers to the header
+instance at index \texttt{0} initially, and is advanced each time that
+\texttt{extract} is invoked on an instance within the stack. If
```
jafingerhut: If it is intended to allow extract on a constant index within a header stack (which I believe Chris Dodd mentioned there are extant uses of in P4_14 programs, e.g. for 2-entry header stacks of VLAN tag headers), then perhaps that should be explicitly mentioned somewhere as permitted?

Also, perhaps the "is advanced each time" condition should be qualified a bit more, e.g. "is advanced each time that \texttt{extract} is invoked on the [next] element of a header stack."

p4-14/v1.0.5/tex/p4.pt
```diff
@@ -1790,12 +1801,6 @@ If the \texttt{header_instance} is not an element in a header stack, the indicat
 header instance is set valid. If the header instance was invalid, all its 
 fields are initialized to 0. If the header instance is already valid, it is 
 not changed.
-
-If \texttt{header_instance} is an element in a header stack, the effect is to push 
-a new header into the stack at the indicated location. Any existing valid 
-instances from the given index or higher are copied to the next higher index. 
-The given instance is set to valid. If the array is fully populated when this 
-operation is executed, then no change is made to the Parsed Representation.
```
jafingerhut: The paragraph before this one only defines the behavior of add_header on a header that is not an element of a header stack.  I thought perhaps that the intent was to continue to allow add_header to be performed on a specified index of a header stack, but that its behavior would be specified to make that one element of the header stack become valid if it was invalid before, but there would be no effect on any other elements of the header stack.

p4-14/v1.0.5/tex/p4.pt
```diff
@@ -1830,10 +1835,6 @@ If the \texttt{header_instance} is not an element in a header stack, then the in
 header instance is marked invalid. It will not be available for matching in 
 subsequent \matchaction stages. The header will not be serialized on egress. 
 All field values in the header instance become uninitialized.
-
-If the \texttt{header_instance} is an element in a header stack, the effect is to 
-pop the indicated element from the stack. Any valid instances in the stack 
-at higher indices are copied to the next lower index.
```
jafingerhut: Similar comment here as for add_header: I thought the intended new behavior was to make the one element of the header stack specified as the argument to the remove_header() operation become invalid, if it was valid before.  Such an operation would have no effect on any other elements of the header stack.

p4-14/v1.0.5/tex/p4.pt
```diff
@@ -2160,21 +2161,20 @@ continues as per the current control function specification.
 
 \primact{push(array, count)}
 { % Summary
-Push all header instances in an array down and add a new header at the top.
+Push all header instances in an array down and add a new invalid header at the top.
```
jafingerhut: One reasonable alternative is to continue to define push() in P4_14 programs to add a new valid header at the top, i.e. to continue to allow this behavior to be different than P4_16's definition of the push_front operation.

There may be extant P4_14 programs that are relying on push() having the effect of making the new inserted headers valid, e.g. I think switch.p4 relies upon this.

p4-14/v1.0.5/tex/p4.pt
jnfoster: Fixed the same way as with `add_header`

p4-14/v1.0.5/tex/p4.pt
jnfoster: Good suggestions both.

Added the paragraph:
```
Note that while it is legal to extract to a constant index in a header
stack and extract to the element referred by \texttt{next}, it is not
recommended. It will likely be difficult for programmers to keep track
of which element in the header stack \texttt{next} refers to when
operations are combined in this way.
```

Clarified by adding:
```
and is advanced each time that \texttt{extract} is invoked on any instance within the stack. 
```

p4-14/v1.0.5/tex/p4.pt
```diff
@@ -1711,7 +1728,7 @@ Drop a packet (in the egress pipeline). \\ \hline
 \texttt{no_op} &
 Placeholder action with no effect. \\ \hline
 \texttt{push} &
-Push all header instances in an array down and add a new header at the top. \\ \hline
+Push all header instances in an array down and add a new invalid header at the top. \\ \hline
```
jafingerhut: With recent change to preserve P4_14 push behavior of adding valid headers, this addition of "invalid" should be removed.

p4-14/v1.0.5/tex/p4.pt
```diff
@@ -2160,21 +2165,20 @@ continues as per the current control function specification.
...
-instances without knowing in advance how many elements are already valid. 
+This primitive is used to initialize elements at the top of a header stack.
 Existing elements will be shifted by \texttt{count}: an
-element at index N will be moved to index N+\texttt{count}. 
-\texttt{count} number of new elements will be pushed at indices 0 to 
-\texttt{count}-1, zeroed out and set valid.
-This primitive leaves the array's size constant; if an array is already full,
-elements pushed to indices beyond the static array size will be lost.
+element at index N will be moved to index N+\texttt{count} and 
+\texttt{count} invalid elements are created at indices 0 to
```
jafingerhut: Similar comment as earlier: replace "invalid" with "valid".

jnfoster: Good catch. Fixed.

p4-14/v1.0.5/tex/p4.pt
jnfoster: Good catch. Fix.

p4-14/v1.0.5/tex/p4.pt
```diff
@@ -705,7 +700,29 @@ References at run time to a header instance (or one of its fields) which is
...
-
+    
+To help programers write parsers for header stacks, P4 provides two
+keywords, \texttt{next} and \texttt{last}, that may be used to refer
+to header instances within a stack. These references are advanced
+automatically as the parser invokes \texttt{extract} in the parser.
+
+More precisely, the keyword \texttt{next} refers to the header
+instance at index \texttt{0} initially, and is advanced each time that
+\texttt{extract} is invoked on any instance within the stack. If
```
jafingerhut: Suppose someone does extract() on a header stack element with a constant index.  This sentence seems to imply that this should cause next to be advanced.  Is that the intended behavior?  I would have guessed that extract on header stack elements with constant indices would not advance next.

jnfoster:
This is what P4_16 does.

Let's discuss it in a little more than an hour.

-N

jafingerhut:
Sure, happy to discuss at the meeting.

P4_16 v1.0.0 spec section 11.8.1 "Fixed width extraction" has pseudocode showing that nextIndex should be incremented only if 'next' is part of the l-value in the argument to extract, if I am understanding it correctly.

jnfoster:
Mea culpa. I should have known better than to contradict you on a P4 language issue! Your recollection is correct, of course.

jnfoster:
@jafingerhut have a look at 548b6a4?

jafingerhut:
Looks good to me.  Thanks!

jnfoster:
Just for hygiene, can you click the approve button? Then I'll merge!

jafingerhut:
Done.

PR #502, opened 2017-11-27
added an example for the resubmit packet path

hanw:
Demonstrated the resubmit packet path can be used to select different parsers to parse the same incoming byte streams.

test passed with the nov-27-2017 version of p4test.


hanw:
This example was committed with #499

PR #509, opened 2017-12-05
add an example to do recirculate in PSA

hanw:
One attempt to do recirculate in PSA and updated some text.

So far, it seems the `recirculate` flag is still necessary to inform the egress deparser that the ingress pipeline has made a decision to recirculate a packet by setting the `egress_port` to `PORT_RECIRCULATE`. The egress deparser needs to know if a packet is recirculation packet by either checking the `egress_port == PORT_RECIRCULATE` or checking `recirculate_flag == 1`. The issue with `egress_port` is that it is not in the `psa_egress_output_metadata_t`, and therefore not visible in egress deparser. On the other hand, the `recirculate` flag is visible in egress deparser. But the caveat is that the `recirculate` flag should be read-only and should be set by the environment. However, `recirculate` flag is currently in `psa_egress_output_metadata_t` which allows both read and write by the egress pipeline.

### DIFF
--- a/p4-16/psa/examples/psa-example-recirculate.p4
+++ b/p4-16/psa/examples/psa-example-recirculate.p4
@@ -117,7 +117,7 @@ control ingress(inout headers hdr,
                 inout psa_ingress_output_metadata_t ostd)
 {
     action do_recirc (PortId_t port) {
-        ostd.egress_port = PORT_RECIRCULATE;
+        send_to_port(ostd, PORT_RECIRCULATE);
     }
     table t {
         key = {
--- a/p4-16/psa/examples/psa-example-recirculate.p4
+++ b/p4-16/psa/examples/psa-example-recirculate.p4
@@ -1,0 +1,208 @@
+
+/*
+Copyright 2017 Barefoot Networks, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <core.p4>
+#include "../psa.p4"
+
+typedef bit<48>  EthernetAddress;
+
+header ethernet_t {
+    EthernetAddress dstAddr;
+    EthernetAddress srcAddr;
+    bit<16>         etherType;
+}
+
+header ipv4_t {
+    bit<4>  version;
+    bit<4>  ihl;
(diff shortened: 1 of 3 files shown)


jafingerhut:
A possibility: We could make the struct of type psa_egress_input_metadata_t that is an input to the egress deparser and egress control blocks, also an input of the egress deparser.  It contains the egress_port field, and could be used to implement the (egress_port == PORT_RECIRCULATE) check in the egress deparser.

If the decision to do recirculate or not must be made during ingress, that seems like potentially a better solution than requiring egress code to assign the recirculate flag to true (and it is a bug if it doesn't).

Either that, or say that the architecture assigns recirculate to true if egress_port == RECIRCULATE, regardless of what your P4 code does, and don't even make the recirculate flag visible in the egress control block.  (We would probably want packet_path to be equal to RECIRCULATE when the packet arrives at the egress parser -- the egress code could read that to know that the packet will be recirculated).

p4-16/psa/examples/psa-example-recirculate.p4
```diff
@@ -0,0 +1,208 @@
...
+    }
+}
+
+control ingress(inout headers hdr,
+                inout metadata user_meta,
+                in  psa_ingress_input_metadata_t  istd,
+                inout psa_ingress_output_metadata_t ostd)
+{
+    action do_recirc (PortId_t port) {
+        ostd.egress_port = PORT_RECIRCULATE;
```
jafingerhut: I think it would be better to use "send_to_port(PORT_RECIRCULATE);" here, as that also sets the drop flag to false.  As written now, the packet will be dropped after ingress because by default the drop flag is true when ingress begins.

p4-16/psa/psa.p4
```diff
@@ -125,8 +125,8 @@ struct psa_egress_output_metadata_t {
   // Egress control block begins executing.
   bool                     clone;         // false
   ClassOfService_t         clone_class_of_service; // 0
+  bool                     recirculate;   // set by environment
   bool                     drop;          // false
-  bool                     recirculate;   // false
```
jafingerhut: We can discuss in the PSA WG meeting, but I think it would be preferable to either:

+ remove the 'recirculate' flag entirely, relying on somehow getting the value of egress_port as an 'in' parameter to the egress deparser, and defining psa_recirculate() to return the value (egress_port == PORT_RECIRCULATE)

or

+ add a new 'in' parameter to the deparser that is a struct containing the recirculate flag.  If it should not be modifiable by the egress control block, it shouldn't be part of an out or inout parameter of the egress control block.

hanw: I prefer to remove `recirculate` flag and replace with an architecture defined `psa_recirculate()`. But, passing the entire `psa_egress_input_metadata_t` as `in` parameter to egress deparser is a bit too much. We may need yet another parameter to egress deparser.

jafingerhut:
@cc10512 @hanw I am going to try to merge this in, since I think there are good changes in this, and I also think it will be easier for me to make further proposed changes on top of this one after it is merged in.

PR #511, opened 2017-12-06
Update PSA example programs for latest version of psa.p4

jafingerhut:
Nothing fancy here -- just updating the other PSA example programs for the most recent changes in psa.p4.  Only one or two of them were updated when psa.p4 was last changed, to test those changes at least a little bit, but waiting for the changes to settle a bit before 'catching up' to the latest version with the rest of the examples.

### DIFF
--- a/p4-16/psa/examples/psa-example-incremental-checksum.p4
+++ b/p4-16/psa/examples/psa-example-incremental-checksum.p4
@@ -55,6 +55,9 @@ header tcp_t {
     bit<16> urgentPtr;
 }
 
+struct empty_metadata_t {
+}
+
 struct fwd_metadata_t {
     bit<32> old_srcAddr;
 }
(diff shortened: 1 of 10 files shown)


p4-16/psa/examples/psa-example-incremental-checksum.p4
```diff
@@ -82,6 +85,8 @@ parser IngressParserImpl(packet_in buffer,
                          out headers parsed_hdr,
                          inout metadata user_meta,
                          in psa_ingress_parser_input_metadata_t istd,
+                         in empty_metadata_t resubmit_meta,
```
cc10512: we really need optional/named arguments in the language spec. We shouldn't need to write this much boilerplate.

jafingerhut: We don't need to write it.  We just need to copy and paste it :-)

PR #513, opened 2017-12-07
Add a section with notes on PRE multicast replication behavior

jafingerhut:
Also restrictions on what kinds of items can be added to multicast
groups.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -604,6 +604,49 @@ message ClassOfServiceInfo {
 ```
 
 
+### Multicast replication
+
+The control plane may configure each `multicast_group` in the PRE to
+create the desired copies of packets sent to that group.  Each group
+begins empty.  Sending a packet to an empty group causes the packet to
+be dropped.  The control plane may add one or more pairs of the form
+`(egress_port, instance)` to a multicast group, and may also remove
+pairs from a group that were added earlier.
+
+Suppose a multicast group contains the following set of pairs:
+
+    (egress_port[0], instance[0]),
+    (egress_port[1], instance[1]),
+    ...,
+    (egress_port[N-1], instance[N-1])
+
+When a packet is sent to that group, `N` copies of the packet are
+made.  Copy number `i` that is sent to egress processing will have its
+`struct` of type `psa_egress_input_metadata_t` filled in with the
+field `egress_port` equal to `egress_port[i]`, and the field
+`instance` filled in with `instance[i]`.
+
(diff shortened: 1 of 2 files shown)


jafingerhut:
@samar-abdi I don't see how in Github to add you as a reviewer, but please take a look at this and add any comments you have.

p4-16/psa/PSA.mdk
```diff
@@ -604,6 +604,49 @@ message ClassOfServiceInfo {
...
+
+Suppose a multicast group contains the following set of pairs:
+
+    (egress_port[0], instance[0]),
+    (egress_port[1], instance[1]),
+    ...,
+    (egress_port[N-1], instance[N-1])
+
+When a packet is sent to that group, `N` copies of the packet are
+made.  Copy number `i` that is sent to egress processing will have its
```
vgurevich: This is a very strong requirement. I do not think that the (port, instance) pairs should be assumed to be in any order and I would not require the replication to follow ay order either.

cc10512: I didn't read the statement as imposing an ordering, just the pairing. Each copy will be uniquely associated with a pair.

jafingerhut: I agree that this text could be misread to imply that copies are generated in a particular order, and I can try to come up with a different way to describe it that doesn't imply that.

In Section 6.2 of the current PSA draft, there is a requirement that PSA implementations process packets in the egress control block in the same relative order that they were processed in the ingress control block, for multicast packet copies with the same (ingress_port, multicast_group, class_of_service, egress_port) values (OK, it doesn't say it exactly that way, but perhaps it should).  That has implications on how packet replication can be implemented.  It doesn't imply that the packet replication set is a list with replication order determined by the control plane, but I expect many implementations would choose to do something like that.

If you think _that_ shouldn't be a requirement, then we should discuss it, because otherwise you are probably going to make some multicast users pretty unhappy with your implementation.

jafingerhut: I've updated the PR with another commit intended to answer @vgurevich's comment.

Section 6.2 actually doesn't describe the FIFO properties as requirements, but 'expectations'.  That may be wishy-washy in language, and we may want to change it to a requirement, or just a recommendation.

jafingerhut: @vgurevich Do these modified changes look better to you?

vgurevich: @jafingerhut ,

Sorry for the late reply. Yes, the changes look good. Thank you!

PR #515, opened 2017-12-07
PSA Update clone operations to use clone_session_id

jafingerhut:
Some text in the clone section describes what the control plane must
configure for each clone session.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -481,18 +481,26 @@ assist P4 developers in debugging their code.
             packets, which are not required to support truncation.
     }
     if (ostd.clone) {
-        if (ostd.clone_class_of_service value is not supported) {
-            ostd.clone_class_of_service = 0;
-            // Recommmended to log error about unsupported
-            // ostd.clone_class_of_service value.
-        }
-        if (platform_port_valid(ostd.clone_port)) {
-            create a clone of the packet and send it to the packet
-                buffer with class of service ostd.clone_class_of_service,
-                after which it will start egress processing.
+        if (ostd.clone_session_id value is supported) {
+            cos = class_of_service configured for ostd.clone_session_id in PRE
+            ep = egress_port configured for ostd.clone_session_id in PRE
+            if (cos value is not supported) {
+                cos = 0;
+                // Recommmended to log error about unsupported cos
+                // value.
+            }
+            if (platform_port_valid(ep)) {
+                create a clone of the packet and send it to the packet
+                    buffer with the egress_port ep and
+                    class_of_service cos, after which it will start
+                    egress processing.
+            } else {
+                // Do not create a clone.  Recommended to log error
+                // about unsupported ep value.
+            }
         } else {
-            // Do not create a clone.  Recommended to log error about
-            // unsupported ostd.clone_port value.
+            // Do not create a clone.  Recommmended to log error about
+            // unsupported ostd.clone_session_id value.
         }
     }
     // Continue below, regardless of whether a clone was created.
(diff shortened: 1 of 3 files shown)


jafingerhut:
@samar-abdi As for other issue updating PSA on multicast replication, this PR contains proposed details for PRE configuration of clone sessions.

p4-16/psa/PSA.mdk
```diff
@@ -931,17 +946,28 @@ are those whose names begin with `clone` in types
...
+and `class_of_service` values that should be associated with packets
+cloned using that session.
+
+The `egress_port` may be any port that can be used for normal unicast
+packets, i.e. any normal port, `PORT_CPU`, or `PORT_RECIRCULATE`.  For
+the latter two values, the cloned packet will be sent to the CPU, or
+recirculated at the end of egress processing, as a normal unicast
+packet would at the end of egress processing.
+
+Since it is an expected common case to clone packets to the CPU, every
```
antoninbas: Thanks @jafingerhut

p4-16/psa/PSA.mdk
```diff
@@ -931,17 +946,28 @@ are those whose names begin with `clone` in types
...
 true, then a cloned packet should be generated at the end of the
-pipeline.  The `clone_port` specifies which port to send the cloned
-packet to, after egress processing.  The `clone_port` could be any
-addressable port in the pipeline.  For instance, it can be a port to
-the control CPU or a port to the next hop router.
+pipeline.  The `clone_session_id` specifies one of several possible
+clone sessions that the control plane may configure in the PRE.  For
+each clone session, the control plane may configure the `egress_port`
+and `class_of_service` values that should be associated with packets
+cloned using that session.
```
samar-abdi: Thanks @jafingerhut 
Are we mandating that these are the only two configurations that a PSA PRE will support? What about truncation?

jafingerhut: I am open to adding a truncation configuration value per clone session, if folks want it.  @hanw @vgurevich thoughts on that?

jafingerhut: In general, I think of PSA as saying "at least this much, but an implementation is free to do more".  Of course, taking advantage of more reduces portability, I know.  If you know of other configuration options besides the ones listed that you want to discuss, this is a good place to ask about them.

samar-abdi: OK, but we will have packet length in the P4Runtime API regardless, with the assumption that either switch SW or ASIC will support it.

antoninbas: We can have it in P4 Runtime if there is interest, but that kind of line-rate packet operations has to be supported in the hardware.

samar-abdi: There is a way to truncate copies by setting the ingress output metadata fields, so this LGTM. Thanks!

jafingerhut: @samar-abdi Is this something you see in the latest PSA draft spec somewhere, and if so, where?  I don't recall such a thing for clone packets, only for non-clone packets, but I could be forgetting it.

samar-abdi: I saw this for the truncate operation:

For all copies of the packet made at the end of ingress processing,
truncate the payload to be at most the specified number of bytes.
Specifying 0 is legal, and causes only packet headers to be sent, with
no payload.
-----------------------------------------

I suppose this applies to clones and multicast replicas.

samar-abdi: @jafingerhut 
Section 6.3.4

jafingerhut: We may want a single place that summarizes how to truncate operations are supposed to work.  For example, here is a little note stuck in Section 6.2 right now "Note: Truncation is not currently supported for cloned packets. The functionality of a truncated cloned packet can be replaced with sending a digest."  There are also notes about truncation not needing to be supported for resubmit and recirculate packets.

I suspect it may be a good idea to add truncate options/parameters in the clone session configuration state.  I will make an issue to discuss at next PSA WG meeting, so people have a chance to comment.

jafingerhut: @samar-abdi Looks like the current PSA draft contradicts itself in different sections on whether truncation must be supported for cloned packets, or not.  I have created this issue that will hopefully be discussed at next PSA WG, and resolved soon: https://github.com/p4lang/p4-spec/issues/519

samar-abdi: @jafingerhut I am ok with merging this PR and dealing with truncation separately. Thanks!

jafingerhut:
@cc10512 OK to merge these changes in, do you think?

cc10512:
Yup, I didn't change my mind :)

jafingerhut:
OK, merged it in, then.

PR #524, opened 2017-12-17
Many edits to clean up move of recirculate decision to ingress

jafingerhut:
Changes include:

+ Eliminate recirculate and resubmit extern objects from psa.p4.
+ Proposed addition of a new struct that contains only egress_port field, to be given as 'in' parameter to the egress deparser, so it can distinguish between recirculated packets vs. others.  It something like this to implement psa_recirculate() function correctly.  This is one of several ways to do it, but we could choose something else.
+ Many small changes to PSA.mdk mentioning things about recirculate, and updating them to the current planned way to do it.
+ A few other small corrections and occurrences of rewording that I hope are improvements in PSA.mdk.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -580,8 +580,8 @@ A similar FIFO property is expected to hold for multicast packets,
 i.e. those that follow the "Make 0 or more copies" path in the
 pseudocode above.  By a short period of time after the control plane
 has stopped modifying the set of members of multicast groups, they
-should be 'stable', and all packets with the same `(ingress_port,`
-`ostd.multicast_group,` `ostd.class_of_service,` `egress_port)` should
+should be 'stable', and all packets with the same (`ingress_port`,
+`ostd.multicast_group`, `ostd.class_of_service`, `egress_port`) should
 pass through the system in FIFO order, and be processed in the
 `Egress` control block in the same relative order that they were
 processed in the `Ingress` control block.
@@ -999,27 +999,10 @@ modified packet after egress processing, as output by the egress
 deparser.  In both cases, the cloned packet is submitted to the egress
 pipeline for further processing.
 
-Logically, PRE implements the mechanics of copying a packet.  PRE
-reads the original packet as it began ingress parsing, or the modified
-packet from the egress deparser.  PRE then decides which version of
-the packet to use to generate a copy based on the value of the
-metadata provided by the ingress and egress pipeline.  The metadata
-fields that control cloning are those whose names begin with `clone`
-in types `psa_ingress_output_metadata_t` and
+Logically, PRE implements the mechanics of copying a packet.  The
+metadata fields that control cloning are those whose names begin with
+`clone` in types `psa_ingress_output_metadata_t` and
 `psa_egress_output_metadata_t`.
-
-~ TBD
-The sentence beginning "PRE then decides which version of the packet
-to use to generate a copy" makes it sound like either you can create a
-clone of a packet in the ingress pipeline, or you can create a clone
-of a packet in the egress pipeline, but you cannot do both for the
-same packet.  The ingress and egress executions of packets, whether
-they occur due to the same original packet or different ones, are
-pretty much independent events, true?
-
-If so, it would be good to reword, or eliminate, that sentence and any
-text that could be misinterpreted in that way.
-~
 
 ```
 bool                     clone;
@@ -1126,9 +1109,9 @@ different headers than the packet had before recirculation.  This
 could be useful in implementing features such as multiple levels of
 tunnel encapsulation or decapsulation.
 
-Packet recirculation happens at the end of the egress pipeline, but
-whether a packet is recirculated must be chosen during ingress
-processing, by sending the packet to port `PORT_RECIRCULATE`.  When a
+Whether a packet is recirculated must be chosen during ingress
+processing, by sending the packet to port `PORT_RECIRCULATE`.  Packet
+recirculation happens at the end of the egress pipeline.  When a
 packet is sent to the recirculate port, the packet finishes egress
 processing, including the egress deparser, and then re-enters the
 ingress parser.  The `ingress_port` of the recirculated packet is set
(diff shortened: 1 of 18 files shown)


jafingerhut:
I have also added an appendix with reasoning given for some high level PSA design choices.  I don't think any of that will be surprising to anyone that has looked at high speed switching devices before -- it is there primarily for those who might be new to these things.

p4-16/psa/PSA.mdk
```diff
@@ -573,15 +580,15 @@ A similar FIFO property is expected to hold for multicast packets,
 i.e. those that follow the "Make 0 or more copies" path in the
 pseudocode above.  By a short period of time after the control plane
 has stopped modifying the set of members of multicast groups, they
-should be 'stable', and all packets with the same `(ingress_port,
-ostd.multicast_group, ostd.class_of_service, egress_port)` should pass
-through the system in FIFO order, and be processed in the `Egress`
-control block in the same relative order that they were processed in
-the `Ingress` control block.
+should be 'stable', and all packets with the same `(ingress_port,`
+`ostd.multicast_group,` `ostd.class_of_service,` `egress_port)` should
```
cc10512: We either quote the entire parentheses, or individual names. Commas should not be quoted: `ostd.multicast_group`, `ostd.class_of_service`, `egress_port`)

p4-16/psa/PSA.mdk
```diff
@@ -573,15 +580,15 @@ A similar FIFO property is expected to hold for multicast packets,
...
+pass through the system in FIFO order, and be processed in the
+`Egress` control block in the same relative order that they were
+processed in the `Ingress` control block.
```
cc10512: Do we really want to imply such ordering guarantees? Is it important for portability to rely on a FIFO impl?

p4-16/psa/PSA.mdk
```diff
@@ -971,22 +989,37 @@ its normal destination according to other features implemented by the
...
-are those whose names begin with `clone` in types
-`psa_ingress_output_metadata_t` and `psa_egress_output_metadata_t`.
+~ TBD
+The sentence beginning "PRE then decides which version of the packet
+to use to generate a copy" makes it sound like either you can create a
+clone of a packet in the ingress pipeline, or you can create a clone
+of a packet in the egress pipeline, but you cannot do both for the
+same packet.  The ingress and egress executions of packets, whether
+they occur due to the same original packet or different ones, are
+pretty much independent events, true?
```
cc10512: Correct, they are independent events and they need to be triggered as such. So maybe a rephrasing is required. How about removing that sentence and add after the "The metadata fields that control ..." sentence, the following: Based on the metadata settings the PRE will generate a copy of the packet using the correct version.

p4-16/psa/PSA.mdk
```diff
@@ -1085,30 +1118,30 @@ resubmission mechanism is disabled and no assignments to
...
+packet, after it has completed egress processing.  Unlike a resubmit,
+where the resubmitted packet contents are identical to the packet that
+arrived at the ingress parser, a recirculated packet may have
+different headers than the packet had before recirculation.  This
+could be useful in implementing features such as multiple levels of
+tunnel encapsulation or decapsulation.
+
+Packet recirculation happens at the end of the egress pipeline, but
+whether a packet is recirculated must be chosen during ingress
+processing, by sending the packet to port `PORT_RECIRCULATE`.  When a
```
cc10512: Nit: This sentence is a bit convoluted. I would probably start with the choice being made in ingress and recirculation at the end of egress. However, it is quite clear what's happening even with the current phrasing. Just the 'no but' rule bothering me here a bit :)

p4-16/psa/PSA.mdk
```diff
@@ -1231,8 +1264,14 @@ The PRE extern object has no constructor, and thus it cannot be
...
 instantiate it. The PRE is made available to the `PSA_Switch` package.
-A corresponding Buffering and Queuing Engine (BQE) extern
-is defined for the egress pipeline (see [#sec-bqe]).
+
+~ TBD
+The statement "The PRE extern object has no constructor" is currently
+wrong.  We should either make it correct by changing psa.p4 somehow,
+or update the text above to match psa.p4.
+
+The same comment applies for the BQE section below.
```
cc10512: Let's update the text. We need quite a bit more work in the language/compiler to support the implicitly constructed externs. Not likely to get my wish before the end of Jan.

p4-16/psa/PSA.mdk
```diff
@@ -2392,3 +2412,138 @@ out of range includes:
...
+have tight control loops monitoring the size of packets leaving egress
+processing to make small adustments in the rate that the scheduling
+algorithm operates for each port.  Either that, or they will just
+leave some fraction of the output port's capacity unused during times
+when all packet sizes are being decreased.
+
+Certainly if there are long durations of time when egress decides to
+drop all packets to an output port, that port will go idle.  The
+scheduling algorithm implementations are all built with a finite
+maximum packet scheduling rate.
```
cc10512: Illuminating (for me). Thanks for putting this together. We might end up expanding it.

p4-16/psa/examples/psa-example-recirculate.p4
```diff
@@ -102,6 +101,12 @@ parser IngressParserImpl(
 
     state parse_recirc_header {
         user_meta.recirc_header = recirc_meta;
+        // @hanw - You could write a P4 program like this, but then
+        // recirculated packets would have no headers parsed and valid
+        // during ingress processing, true?  It seems like a much
+        // better and more typical example if we use 'transition
+        // parse_ethernet' instead of 'transition accept' on the next
+        // line.  Do you agree?
```
cc10512: Is this still a question?

hanw: I agree this state should transit to parse_ethernet.

p4-16/psa/PSA.mdk
jafingerhut: I had the whole expression quoted before, but that seems to cause Madoko to put the entire thing on one line, with no line breaks allowed.  What would you prefer?

cc10512: the individual fields without comma and parentheses, just as I put in the comment. oh, well, i guess i should have used some sort of raw input, because github interpreted the quotes :)

p4-16/psa/examples/psa-example-recirculate.p4
jafingerhut: I think I know the answer, but would appreciate it if @hanw could look at this part and confirm my belief, and the suggested change to the example program.

p4-16/psa/PSA.mdk
jafingerhut: No problem.  It is an issue that has come up a couple of times, and nice to have a place to give an answer in a little more than a phrase, but a lot less than 10 pages.

p4-16/psa/examples/psa-example-resubmit.p4
```diff
@@ -157,15 +139,27 @@ control ingress(inout headers hdr,
...
 }
 // END:Resubmit_Example_Part2
 
-parser EgressParserImpl(packet_in buffer,
+parser EgressParserImpl(
+    packet_in buffer,
     out headers parsed_hdr,
     inout metadata user_meta,
     in psa_egress_parser_input_metadata_t istd,
+    in metadata normal_meta,
```
hanw: normal_meta is not used in egress parser, use `empty_metadata_t`?

p4-16/psa/PSA.mdk
jafingerhut: As currently written, this is not a requirement.  We can leave it out if it is too restrictive on implementations, but note that there are likely to be some P4 programs that do updates on registers in ingress and egress that might implicitly _assume_ such ordering properties, and fail on devices that don't.

jafingerhut: Commercial switches typically implement these ordering restrictions as packets are transmitted on the output ports, at least.  Whether or not they implement that by making the same restriction on the relative order that packets pass through their internal processing pipelines, or by allowing out of order processing, but then reorder before transmitting the packets on the wire, is an implementation detail.

If a commercial top-of-rack switch did _not_ implement this port-in-to-port-out ordering constraint, several UDP-based applications would break, and several (but not all) TCP implementations in operating systems would suffer factors of 10 or larger drops in application throughput, because they misinterpret too much packet reordering in the network as packet drops, thus a sign of network congestion, and reduce their TCP transmission window.  That is why LAGs and ECMP are typically selected among using a hash of packet header fields, not packet-by-packet round robin.

p4-16/psa/PSA.mdk
jafingerhut: Will soon add another commit to this PR that updates this text, and removed the TBD.

p4-16/psa/PSA.mdk
jafingerhut: I will soon include that change with another commit on this PR.  The PDF output from Madoko still looks reasonably formatted with that change.

p4-16/psa/PSA.mdk
jafingerhut: I have tried simply eliminating the sentence in the commit I will add to this PR soon.  Give it a look over.  With the other text in the rest of this section, I don't see much that the sentence was saying that couldn't be understood from the rest.

p4-16/psa/PSA.mdk
jafingerhut: I've made it two sentences, and reversed the order as you suggested.

p4-16/psa/examples/psa-example-recirculate.p4
jafingerhut: Thanks, Han.  I have changed the transition statement to parse_ethernet, and deleted the question about it in comments.

p4-16/psa/examples/psa-example-resubmit.p4
jafingerhut: Good catch. I made that change, too.

Issue #506, opened 2017-12-01
P4_16: Result of `(int<W>) X` right shifted by W or more bits when X is negative?


jafingerhut:
Section 8.6 "Operations on fixed-width signed integers" says this about left and right shift operations:

"Arithmetic shift left and right denoted by `<<` and `>>`. The left operand is signed and the right operand must be either an unsigned number of type `bit<S>` or a non-negative integer literal. The result has the same type as the left operand. Shifting by an amount greater than the width of the input produces a result where all bits are zero."

The only part I have a possible issue with is the last sentence.  Suppose X is a value of type `int<16>` and it is two's complement negative, so its most significant bit `X[15:15]` is 1.  Arithmetic right shifts on X of less than 16 bits should do sign extension, effectively copying the most significant bit during the right shifts.

The sentence in the spec seems to imply that the result of `X >> 17` should be 0, whereas from an arithmetic shift I would perhaps expect a result of all 1 bits (-1 in 2's complement) if X was negative.

It appears from this STF test: https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/arith5-bmv2.stf#L25-L26

that the combination of p4c-bm2-ss and simple_switch are doing what I would expect in this case: a right shift of a negative `int<32>` value by 255 bits results in all 1's.

It is the P4_16 spec language that I am questioning here.

mbudiu-vmw:
This is a bug, we should fix it.

jafingerhut:
Just to clarify your comment, do you believe it is a bug in the P4_16 language spec, or the current p4c and behavioral-model implementation?

mbudiu-vmw:
it is a bug in the spec.

jafingerhut:
Closing this issue, as PR #526 has made changes to address it.

PR #526, opened 2017-12-18
Fix for issue #506

mbudiu-vmw:


jafingerhut:
Looks good to me.

On a related topic, are logical shift _left_ and arithmetic shift _left_ completely equivalent in their bit-level behavior for all inputs, for integers represented in 2's complement?  I believe this is so, but wanted to make sure I wasn't missing anything.  If true, it might warrant a phrase in the language spec when mentioning arithmetic shift left.

mbudiu-vmw:
I think they are equivalent. This is not really a P4-specific feature.

ChrisDodd:
I would say no, since a signed arithmetic left shift is generally understood to shift just the value bits and leave the sign bit untouched, while a logical left shift will shift the whole word.  Hardware support for signed arithmetic left shift is rare (at least on CPUs), though Intel has a 'SAL' instruction which is an alias for 'SHL' (ie, ISN'T a signed arithmetic shift.)

Intel call

jafingerhut:
Given the fact that hardware implementations of arithmetic shift left that preserve the sign bit are rare on CPUs, it seems prudent for the P4_16 spec to be explicit about what it means by an arithmetic shift left operation, rather than merely using that term in English.

Note: I don't care much one way or the other how it ends up being defined -- I know how to do all the signed stuff I want with unsigned `bit<W>` types, like any good Verilog programmer :-)

jafingerhut:
I can create a separate issue for this, so this looks-good-to-me PR can proceed independently of that question.

ChrisDodd:
I guess you can claim that in the absence of overflow, they're the same.  So if you say that the result of an overflow on a signed left shift is undefined (an undefined value), then that should cover it.

PR #529, opened 2017-12-20
PSA Update example program psa-example-digest.p4 so it almost compile…

jafingerhut:
…s cleanly

### DIFF
--- a/p4-16/psa/examples/psa-example-digest.p4
+++ b/p4-16/psa/examples/psa-example-digest.p4
@@ -52,14 +52,17 @@ struct headers {
     ipv4_t           ipv4;
 }
 
-struct digest_t {
-    bit<48> srcAddr;
-    bit<9>  ingress_port;
+struct empty_metadata_t {
+}
+
+struct mac_learn_digest_t {
+    EthernetAddress srcAddr;
+    PortId_t        ingress_port;
 }
 
 struct metadata {
-    bit<3>   digest_id;
-    digest_t digest_h;
+    bit<3>             digest_id;
+    mac_learn_digest_t mac_learn_digest;
 }
 
 parser CommonParser(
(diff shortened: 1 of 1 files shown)


jafingerhut:
Most of these changes are just updating to the latest version of psa.p4, and fixing compiler errors, but I also made a few small changes to the content of the example that I think make it closer to a subset of a learning Ethernet bridge.

p4-16/psa/examples/psa-example-digest.p4
```diff
@@ -170,14 +220,15 @@ control IngressDeparserImpl(packet_out packet,
     CommonDeparserImpl() common_deparser;
```
hanw:
```
psa-example-digest.p4(219): error: digest: Type parameters must be supplied for constructor
    Digest(0) digest;
              ^^^^^^
psa-example-digest.p4(223): error: digest.pack: Cannot unify struct mac_learn_digest_t to T
            digest.pack(meta.mac_learn_digest);
```


Should be `Digest<mac_learn_digest_t>(0) digest`.

jafingerhut: I've created an issue to remind us about the editing still needed for Digest extern: https://github.com/p4lang/p4-spec/issues/530

Feel free to claim it if you have time.

jafingerhut:
@hanw We could do what you suggest to eliminate the ostd warnings, but I would prefer not to do such weird things in the example code if we don't intend them to be written in PSA P4 programs like that.

My expectation is that once PSA is implemented in p4lang/p4c, then those warnings should go away.  Is that your expectation, too?

hanw:
I don't think the warning for uninitialized `out` parameter is going to go away in a PSA backend if a user does not set a value to the parameter in their program.

jafingerhut:
@hanw Do you agree that it would be very strange to require a user to assign a value to ostd.parser_error explicitly in their program, or else they would get a warning?

hanw:
Yes, I agree. Are you suggesting the backend should assign a default value to ostd.parser_error?

jafingerhut:
I just looked at the sample architecture in the P4_16 language spec, and it doesn't even have the parser error as an 'out' parameter of the parser, but it does have parser_error as an 'in' parameter to the ingress control block.  If that makes sense to do in an architecture, we could define it that way in PSA, too.

PR #531, opened 2017-12-20
prefix constants with PSA_

cc10512:
Follows the agreement in the discussion on Dec 6, 2017.

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -739,7 +739,7 @@ packet contents and metadata when a packet begins egress processing.
 |                   | of ingress packet  | configuration      | of clone session       ||
 |                   |                    | of multicast group |                        ||
 +--------------+-------------+---------------+--------------+-----+
-| `packet_path`   | NORMAL | NORMAL   | CLONE_I2E | CLONE_E2E |
+| `packet_path`   | NORMAL_ | NORMAL_   | CLONE_I2E | CLONE_E2E |
 |                 | UNICAST | MULTICAST |           |           |
 +--------------+-------------+---------------+--------------+-----+
 +--------------+-------------+---------------+--------------+-----+
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -351,9 +351,9 @@ packet contents and metadata when a packet begins ingress processing.
 | IngressParser `istd` fields (type `psa_ingress_parser_input_metadata_t`) |||||
 +--------------+-------------+---------------+--------------+-----+
 | `ingress_port`      | `PortId_t` value of | `PORT_` | copied from    | `PORT_`       |
-|                     | packet's input port | `CPU`   | resub'd packet | `PSA_RECIRCULATE` |
+|                     | packet's input port | `CPU`   | resub'd packet | `RECIRCULATE` |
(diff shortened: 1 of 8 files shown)


cc10512:
I renamed the enum values because the enum types themselves are PSA specific and not renamed. I could go either way on these as well, and indeed, naming values as `PSA::PacketPath_t.NORMAL` would be ideal.

@vgurevich may we get a vote from you on the following choices:
1. `PSA::PacketPath_t.NORMAL` -- namespace for PSA, applies to the enum type. Requires namespace support in the language, which we discussed for externs.
2. `PSA_PacketPath_t.NORMAL` -- prefix PSA defined types with `PSA_`
3. `PacketPath_t.PSA_NORMAL` -- prefix the values with `PSA_`

hanw:
My vote goes to 1 or as-is right now.

jafingerhut:
@vgurevich Any opinion on the question in the issue above?  Options 2 and 3 do not require any changes to the P4_16 language spec.  I believe that option 1 would require adding namespace support to the P4_16 language spec.  Another alternative would be to leave "PSA" out of the enum types _and_ names, i.e. "PacketPath_t.NORMAL" would stay like that, with no occurrence of PSA in it.

I guess I would have a slight preference for option 2, if we want PSA in there somewhere.

vgurevich:
@jafingerhut ,

Sorry for the late reply. 

I think that option 1 is the best, but if I were to choose between 2 and 3, option 2 makes more sense to me. All enum value names effectively live in their own namespace, so prefixing them doesn't buy anything. As for the types, I believe that prefixing them with PSA_  will prevent unnecessary conflicts and misunderstandings when other architectures appear.

jnfoster:
Speaking not with my chair hat on, I would like to have namespace support in the language. So if PSA is a good driver for it, maybe a small group of people can get together and hammer out the design?

mbudiu-vmw:
We should base whatever design an existing design that works.
C# for example?
But we have other more high-priority items first in line, like supporting value sets.

cc10512:
@jafingerhut, @vgurevich, @hanw, can you please review again? I implemented option 2.

p4-16/psa/PSA.mdk
```diff
@@ -739,8 +739,8 @@ packet contents and metadata when a packet begins egress processing.
 |                   | of ingress packet  | configuration      | of clone session       ||
 |                   |                    | of multicast group |                        ||
 +--------------+-------------+---------------+--------------+-----+
-| `packet_path`   | NORMAL_ | NORMAL_   | CLONE_I2E | CLONE_E2E |
-|                   | UNICAST | MULTICAST |           |           |
+| `packet_path`   | NORMAL | NORMAL   | CLONE_I2E | CLONE_E2E |
+|                 | UNICAST | MULTICAST |           |           |
```
jafingerhut: Might be nice to restore the underscore characters that were removed from after the word 'NORMAL_' here.

jafingerhut:
@cc10512 I added one minor comment on 2 changes in one of the tables that could also be done (missing underscore character), but except for that looks great to me.  Thanks.

p4-16/psa/PSA.mdk
cc10512: Done

Issue #339, opened 2017-06-30
[PSA] "reject" state semantics


vgurevich:
P4_16 spec leaves the semantics of the "reject" state up to the architecture, so we need to define it.

Options:
1. Packets that reach "reject" state are dropped (with an optional counter or counters per error)
2. Packets that reach "reject" state are not dropped, but instead proceed to the ingress control that can choose to drop a packet, ingrement a counter, etc., based on the parser error that should be an input).

vgurevich:
As per brief discussion in the committee...

Option 1 is really simple and protects "unsuspecting" programs from malformed packets. However, given that the packets can reach "reject" state implicitly, it leaves no choices to the advanced users. 

Option 2 is more complicated and the programmer must write the code to deal with the malformed packets. However, the flexibility is enormously better.  For programmers who want to write safe code, while not having to deal with the complexity, PSA can always provide a global action, such as
```
action psa_packet_rejected_drop() {
    if (input_meta.parser_error != error.NoError) {
        pre.drop()
        exit;
    }
}
```

jafingerhut:
An issue from p4lang/p4c that has some related discussion. 
 Not sure if it helps inform the answer to this issue or not: https://github.com/p4lang/p4c/issues/675

DGeorgeNetro:
I quite like Option 2, but if this is the standard it means packets deemed bad by the parser will make it into ingress processing. There might well be some overhead in transitioning between parser and ingress which would could have been avoided for the case of the bad packet.

That said, I don't see any spec-conforming solution other than 2.

vgurevich:
@hanw , @cc10512 ,

I know I am on the hook to write this up, but I think this is actually a very small piece. The most important thing is to decide how errors will be exposed to the ingress control -- without that I can't explain how to process packets that had  parsing/checksum errors.

cc10512:
Please add what you need to the `psa_ingress_input_metadata_t` struct.

vgurevich:
@hanw and other interested parties...

We need to nail down the error handling otherwise ingress (or egress) control will not be able to do much. This consists of the following:
[...]

PR #534, opened 2017-12-20
Remove unnecessary psa_parser_output_metadata_t struct

jafingerhut:
The only reason it was there was to carry the parser error out of the
ingress and egress parsers, but the PSA architecture can just
transport those parser errors to the ingress and egress control blocks
'somehow' by however the architecture does it, without making it an
'out' parameter of the parsers.

Note that the example architecture in the P4_16 language specification
does the same thing.

### DIFF
--- a/p4-16/psa/Makefile
+++ b/p4-16/psa/Makefile
@@ -20,7 +20,8 @@ clean:
 # regularly occur with these example programs for the 'out
 # psa_parser_output_metadata_t ostd' parameter of the parsers.
 
-P4C=p4test --Wdisable=uninitialized_out_param
+P4C=p4test
+#P4C=p4test --Wdisable=uninitialized_out_param
 
 check:
 	${P4C} examples/psa-example-bridged-metadata.p4
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -358,8 +358,8 @@ packet contents and metadata when a packet begins ingress processing.
 |                     | For RESUB or RECIRC packets, the time the 'copy'    ||||
 |                     | began IngressParser, not the original.              ||||
 +--------------+-------------+---------------+--------------+-----+
-| `parser_error`      | From output of IngressParser.  Always `error.NoError` if there ||||
-|                     | was no parser error.                                           ||||
+| `parser_error`      | From IngressParser.  Always `error.NoError` if there ||||
+|                     | was no parser error.                                 ||||
 |--------------|-------------|---------------|--------------|-----|
 ~
 
@@ -765,8 +765,8 @@ packet contents and metadata when a packet begins egress processing.
 | `egress_timestamp` | Time that packet began processing in EgressParser.  Filled in ||||
 |                     | independently for each copy of a multicast-replicated packet. ||||
 +--------------+-------------+---------------+--------------+-----+
-| `parser_error`    | From output of EgressParser.  Always `error.NoError` if there ||||
-|                   | was no parser error.  See "Multicast copies" section.         ||||
+| `parser_error`    | From EgressParser.  Always `error.NoError` if there   ||||
+|                   | was no parser error.  See "Multicast copies" section. ||||
 |--------------|-------------|---------------|--------------|-----|
 ~
 
--- a/p4-16/psa/examples/psa-example-bridged-metadata.p4
+++ b/p4-16/psa/examples/psa-example-bridged-metadata.p4
@@ -171,8 +171,7 @@ parser IngressParserImpl(
     inout metadata meta,
     in psa_ingress_parser_input_metadata_t istd,
     in resubmit_metadata_t resubmit_meta,
-    in recirculate_metadata_t recirculate_meta,
-    out psa_parser_output_metadata_t ostd)
+    in recirculate_metadata_t recirculate_meta)
 {
     CommonParser() p;
 
@@ -208,8 +207,7 @@ parser EgressParserImpl(
     in psa_egress_parser_input_metadata_t istd,
     in metadata normal_meta,
     in clone_i2e_metadata_t clone_i2e_meta,
-    in clone_e2e_metadata_t clone_e2e_meta,
-    out psa_parser_output_metadata_t ostd)
+    in clone_e2e_metadata_t clone_e2e_meta)
 {
     CommonParser() p;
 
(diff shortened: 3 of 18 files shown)


jafingerhut:
I don't think it is a big deal whether we make this change or not, but @hanw was noticing all the warnings about this ingress/egress parser 'out' parameter not being assigned a value in the example programs, and asking whether the example programs should be assigning this output parameter a value explicitly in the code to avoid those warnings.

The P4_16 language spec example architecture has a parser_error input to the ingress control block, but it _doesn't_ make it an 'out' parameter of the parsers.  It just assumes that an architecture can access the error result of the parser somehow.  If this is reasonable to do, I think PSA should do it, too.

mbudiu-vmw:
In fact error should *not* be an out parameter. The reason is that the runtime writes to error behind the scenes, e.g., potentially in extract or verify calls, without passing error as an argument to these method calls. So the parser access to the error is very un-P4 like. Everywhere else we promise that a method cannot mutate any P4 dataplane value that is not an out argument.

jafingerhut:
@hanw Using verify() in P4_16 language spec, a user can assign whatever value to the parser error state that they wish.  They can even use a condition of 'false' if they wish to make it happen unconditionally.

jafingerhut:
@hanw There is one example already demonstrating one way to write a PSA P4 program that drops all packets that had a parser error, and counts how many packets experienced each kind of parser error.  It is in the file psa-example-parser-checksum.p4

@vgurevich has also said before that despite whatever the P4_16 language spec might say, he expects some implementations to leave the state of headers in an unpredictable state if a parser error occurs.  While that is not a faithful P4_16 implementation from what I can tell, I understand why Kangaroo parser implementations that can do multiple parser state transitions in a single clock cycle might make that difficult to avoid.

In such an implementation, the only reasonable ways I can see to handle errors in a P4 program are:

(1) drop them, optionally counting them, as we already have an example for.

(2) drop it, and create a clone to the CPU port (or some other port), with some kind of user-defined to-CPU header in front of it that the control plane software can use to distinguish this packet it is receiving as "dropped by the P4 program because of parser error X", e.g. with fields like switch_id and ingress_port, maybe a timestamp, etc.  This avoids the issue of unpredictable contents of parsed headers, because CLONE_I2E in PSA is defined to take the packet _as it arrived at the ingress parser_ and send that to the packet buffer.  As long as you can make the egress parser not repeat the same error, by adding some extra metadata that this packet experienced an ingress parser error, and stopping the egress parser before doing its regular thing (because otherwise it would end in the same unpredictable state the ingress parser did, and then you can't predict the contents of the packet leaving the egress deparser, either), you should be able to get predictable contents of the packet to the CPU.

If you try to do anything else with such a packet, you are probably asking for trouble on a P4_16 (subset) implementation that exhibits that kind of unpredictable behavior.

jafingerhut:
@cc10512 With Mihai's comment above, it seems that these changes are a bit more important than I originally thought to make in psa.p4 and the examples.  A user's program can assign a parser error using verify(), and they shouldn't really be allowed to cause a parser error in any other way, e.g. by trying to do a direct assignment statement to an 'out' parameter that we currently have, but don't need.

vgurevich:
I think it is reasonable, given the "exceptional" (pun intended) nature of parser_error to make it the input to the ingress/egress control while not being the explicit output from the parser. The spec certainly allows that.

In terms of which headers are valid after a parser error has been encountered, I am not sure I can find much in the spec. I think that a correctly written program should do exactly what you described: either drop such a packet or use the ingress_to_egress mirroring and explicitly avoid parsing the packet at the egress in order to pass the original packet onwards.

jafingerhut:
@vgurevich The P4_16 language spec has 0 occurrences of "unspecified" or "undefined" in the section describing the parser behavior.  If an error occurs, this implies that every header that has had extract() or setValid() called on it during parser execution up to that point should be valid, and every header that has not, should be invalid.  Every header for which extract() has been called that did not cause an error should be valid, and have defined values in all of its fields.

If an extract() call causes an error (e.g. PacketTooShort, HeaderTooLong), then the P4_16 language spec says that header into which you were trying to extract should not have its valid bit set to true.

I realize that some implementations might not implement this behavior as written, and that is par for the course.  I can't think of very much useful anyone would actually want to do with a packet that had a parser error, so it doesn't mess up any use cases I know about if headers end up with contents different than the spec currently requires.

jafingerhut:
I went back and found where @vgurevich had suggested this earlier as one possible option for PSA implementations (from this comment https://github.com/p4lang/p4-spec/issues/339#issuecomment-325722756 )

"Option 1: PSA-compliant programs should not assume that any of the headers are valid or invalid if parser_error is set to anything other than NoError. Note: this does not preclude reading the header fields: the results will be indeterminate, but it is safe to do. But writing the fields or doing the match will be a bad idea."

I realize that it was a suggestion, and it isn't in the PSA spec, but it stuck in my mind because _if_ we wanted to do that, it seems to me that it could not be done in PSA -- it would require changing the P4_16 language spec to allow such less-defined behavior.

mbudiu-vmw:
The way p4-16 is defined, almost anything can happen between two P4 controls/parsers, except the value of the user-defined data changing. All parameters whose type is given by a type variables are by construction invisible to the architecture, because the architecture has no way of knowing what they are. The architecture may never give control to the ingress, but if it does all the header fields that had a defined value when the last parser transition occurs should have the same defined value.

cc10512:
I would prefer to avoid opening the door for undefined reads for partially extracted headers. If a target wishes to implement such behavior in mapping a PSA program, that is their right and it becomes a relaxation of the PSA mandated behavior, which is to say that you can't read or write fields for headers that raised a parse error. I would like to advocate for that.

Handling the parser error directly as a metadata that materializes after parsing and does not need to be explicitly passed seems easy to support -- it being what we are currently doing. If @mbudiu-vmw is ok with it, I'm good. It does make parser error in some sense a "global", so we should restrict it's scope within the control immediately following the parser. Beside argument passing, we do not have any other such mechanism.

mbudiu-vmw:
The very simple switch in the P4-16 spec is structured this way.
Once you have entered the ingress and you have a handle to the error you can pass it manually to other pipelines, as user-defined metadata. So you don't need anything special - assuming you can serialize it.

jafingerhut:
@cc10512 If a P4 implementation does not implement what the P4_16 language spec says, this PR for the language spec has this 'blanket statement' that I think would apply to an implementation that didn't leave the headers in a state as defined by the spec:

"It is understood that some implementations may be unable to implement
the behavior defined here in all cases, or may provide options to
eliminate some safety guarantees in exchange for better performance or
handling larger programs.  They should document where they deviate
from this specification."

https://github.com/p4lang/p4-spec/pull/450

PR #538, opened 2017-12-21
PSA Update spec only to support truncation for cloned packets

jafingerhut:


p4-16/psa/PSA.mdk
```diff
@@ -472,20 +471,12 @@ assist P4 developers in debugging their code.
 
     psa_ingress_output_metadata_t ostd;
 
-    if (ostd.class_of_service value is not supported) {
-        ostd.class_of_service = 0;    // use default class 0 instead
-        // Recommended to log error about unsupported
-        // ostd.class_of_service value.
-    }
```
jafingerhut: The block of pseudocode above was not deleted, just moved a little bit later, so it was closer to the rest of the non-clone packet handling.

p4-16/psa/PSA.mdk
```diff
@@ -913,16 +897,15 @@ the contents of several metadata fields in the struct
         drop the packet
         return;   // Do not continue below.
     }
-    if (ostd.egress_port == PORT_RECIRCULATE) {
-        recirculate the packet, i.e. it will go back to starting with the
-            ingress parser;
-        return;   // Do not continue below.
-    }
-
```
jafingerhut: The code above was not deleted, just moved below the next comment, because I also corrected the occurrence of 'ostd.egress_port' in the original with what it should have been 'istd.egress_port'.

p4-16/psa/PSA.mdk
```diff
@@ -1013,16 +984,43 @@ The `clone` flag specifies whether a packet should be cloned.  If
...
+each clone session, the control plane may configure the following
+values that should be associated with packets cloned using that
+session.
+
+```
+PortId_t         egress_port;
+CloneSessionId_t class_of_service;
+bool             truncate;
+PacketLength_t   truncate_length_bytes;
+```
```
jafingerhut: @samar-abdi Take a look at these updated configuration parameters of a clone session to see if it looks reasonable.  egress_port and class_of_service were already there -- I've just made their PSA types explicit now.  truncate and truncate_length_bytes are new with this PR.

p4-16/psa/PSA.mdk
```diff
@@ -388,15 +388,14 @@ to raise an error.
 
 For RESUB packets, `packet_in` is the same as the pre-IngressParser
 contents of `packet_in`, for the packet that caused this resubmitted
-packet to occur (i.e. with NO modifications from any ingress processing).  Truncation is not supported for RESUB packets.
+packet to occur (i.e. with NO modifications from any ingress processing).
```
jafingerhut: Note that removing this sentence 'Truncation is not supported for RESUB packets.' does _not_ mean that truncation is supported in PSA for resub packets -- with this PR, truncation is _only mentioned as an option for cloned packets_.  If we don't say anything about truncation for other kinds of packets, then we are neither prohibiting it nor requiring it for those other kinds of packets.

p4-16/psa/PSA.mdk
samar-abdi: Looks good. Thanks! I would suggest s/truncate_length_bytes/packet_length_bytes;

cc10512: If we go with @samar-abdi's renaming (which makes sense because is not how much we're truncating, but rather how much is left), I suggest we add a comment explaining the link to truncation. Alternatively, we can rename it to `truncated_packet_length` (and the comment that it is in bytes).

p4-16/psa/PSA.mdk
```diff
@@ -1013,16 +984,43 @@ The `clone` flag specifies whether a packet should be cloned.  If
...
+PacketLength_t   truncate_length_bytes;
+```
 
 The `egress_port` may be any port that can be used for normal unicast
 packets, i.e. any normal port, `PORT_CPU`, or `PORT_RECIRCULATE`.  For
 the latter two values, the cloned packet will be sent to the CPU, or
 recirculated at the end of egress processing, as a normal unicast
 packet would at the end of egress processing.
 
+Truncation of cloned packets is supported as an optimization, to
```
cc10512: no comma?

p4-16/psa/PSA.mdk
jafingerhut: I renamed it to packet_length_bytes.  Also added a note that a PSA implementation is allowed to support a restricted subset of possible values for this property of a clone session, e.g. a PSA implementation might only support multiples of 32 bytes.

PR #541, opened 2017-12-22
PSA Add example program psa-example-parser-error-handling.p4

jafingerhut:


jafingerhut:
@vgurevich @hanw Just a reminder about this proposed program for sending packets that experienced parser errors to the control plane.

Writing this program made me realize that while the ingress-to-egress clone enables us to write a program that sends the exact packet that experienced a parser error in the ingress parser to the control plane, the PSA spec does not currently have a way to do an egress-to-egress clone that is exactly what the egress parser started with.  If it did, we could use that to send a clone to the control plane that was exactly the contents of the packet as received by the egress parser.  Without it, the best I can think of to do is as this program is written: send a packet to the control plane that is the result of getting an error in the egress parser, then goes through the egress deparser, emit'ing whatever the state of the headers happen to be, which Vladimir has suggested might be unpredictable in some implementations.

vgurevich:
@jafingerhut ,

You are correct. The ability to copy packets that experienced errors in the egress parser errors to the CPU would have been useful and currently there is no reliable way to do it in any direct manner.

However, there is currently **no way** in PSA for an arbitrary packet to get straight into the egress parser without passing through the ingress parser (and other components of the ingress pipeline). Even ingress-to-egress cloned packets get into the egress pipeline because the same packet has been first processed by the ingress pipeline that made cloning decision. Therefore, the fact that a program would experience an error in the egress parsing will probably indicate a bug in the program itself (such as lack of good error processing in the ingress parser or incorrect header manipulation in the ingress control or deparser). 

Thus, such a feature, while desirable and useful for debugging these kinds of errors is probably not as critical for the correctness /completeness/security as the ability to get the unmodified copy of the original ingress packet.

p4-16/psa/examples/psa-example-parser-error-handling.p4
```diff
@@ -0,0 +1,601 @@
...
+            meta.to_cpu_error_hdr.ingress = 0;
+            egress_drop(ostd);
+            ostd.clone = true;
+            ostd.clone_session_id = PSA_CLONE_SESSION_TO_CPU;
+            exit;
+        }
+    }
+}
+
+// BEGIN:Compute_New_IPv4_Checksum_Example
```
cc10512: copy & paste?

p4-16/psa/examples/psa-example-parser-error-handling.p4
```diff
@@ -0,0 +1,601 @@
...
+                clone_e2e_meta.to_cpu_error_hdr = meta.to_cpu_error_hdr;
+            }
+            // If you have other reasons to do CLONE_E2E with
+            // different metadata you want to carry with them, this is
+            // where to do the assignments you want.
+        }
+        cd.apply(packet, hdr);
+    }
+}
+// END:Compute_New_IPv4_Checksum_Example
```
cc10512: and here?

p4-16/psa/examples/psa-example-parser-error-handling.p4
jafingerhut: Removed the 2 unused obsolete comments you pointed out.  This new example program is currently not used anywhere in the PSA.mdk file.  Not sure whether we want to do so for this program, or perhaps merely mention its name and a brief description in the document?

Issue #537, opened 2017-12-20
P4_16 How is arithmetic left shift defined?


jafingerhut:
Discussion on a different issue led to this question: https://github.com/p4lang/p4-spec/pull/526

Basic question: Should a P4_16 left shift of a value with type `int<W>` always give the same bit-wise result as a logical shift left of the same bit pattern with type `bit<W>`?

Or, should an arithmetic left shift always have a result that has the same value of the most significant 'sign' bit as the input value?

Section 8.6.1. "A note about shifts" doesn't appear to mention this issue, only the behavior of right shifts.

PR #543, opened 2017-12-22
Fix for issue #537

mbudiu-vmw:


jnfoster:
LGTM but probably we should discuss at the January LDWG?

Issue #542, opened 2017-12-22
The spec does not define equality for tuples or structs.


mbudiu-vmw:
I think we should extend the spec rather than forbid these operations.

jamescoole-cisco:
Makes sense to me. We may also want to consider defining it for headers.

jamescoole-cisco:
The language for structs includes a requirement that their fields be recursively comparable

> Two structs can be compared for equality (==) or inequality (!=) only if they have the same type and all of their fields can be recursively compared for equality.

but this language is missing for headers

> Two headers can be compared for equality (==) or inequality (!=) only if they have the same type.

However, varbits can appear in headers but don't appear to be directly comparable https://github.com/p4lang/p4-spec/blob/54e425a99a387130f946d8a522bd36ac65af1e30/p4-16/spec/P4-16-spec.mdk#L2968-L2981

It seems like we should allow comparing headers that contain varbits, but if we're going to allow that we should probably also define == and != for varbits so they can be directly compared as well.

PR #544, opened 2017-12-22
Fix for issue #542

mbudiu-vmw:


jnfoster:
LGTM, but probably we should discuss briefly at the January P4 LDWG meeting?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3620,6 +3635,10 @@ parser Tcp_option_parser(packet_in b, out Tcp_option_stack vec) {
 }
 ~End P4Example
 
+Two header unions can be compared for equality (==) or inequality (!=)
+if they have the exact same type.  The unions are equal if and only if
+all their corresponding fields are equal.
```
jnfoster: Only one field in a union can be "hot" so the "all" seems potentially confusing.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3620,6 +3635,10 @@ parser Tcp_option_parser(packet_in b, out Tcp_option_stack vec) {
 }
 ~End P4Example
 
+Two header unions can be compared for equality (==) or inequality (!=)
+if they have the exact same type.  The unions are equal if and only if
```
jnfoster: Should we use the same phrasing as for structs here? (i.e., "same type and fields can be recursively...") I'm not sure why this is "exact same type."

jafingerhut: Perhaps this is opening a can of worms by asking this question, but it seems that most places in P4_16 use what I think @jnfoster called "structural typing", which I believe means that two types are considered "compatible" for purposes of comparison, assignment, and passing as a run-time parameter, if they have the same number of sub-elements, in the same relative order, and recursively those also have compatible types.  For example, the existing spec and open source tools treat any values typedef'd with different names, but with the same `bit<17>` underlying type, as compatible.

This is as opposed to "nominal typing", I think, where such an assignment between two typedef'd values with different typedef names, but the same underlying `bit<17>` type, would require an explicit cast or it would be a type error?

If I've got those definitions at least approximately correct, then it sounds like @jnfoster's comment here is that "exact same type" is misleading, if structural typing is allowed for determining whether two values can be compared via `==` or `!=`.  They don't have to be the exact same type, just structurally the same.

mbudiu-vmw: typedef just introduces another name for the same type, but the type is essentially the same.
This is like C.
Everywhere we require types to be the same, not structurally the same.

jafingerhut: So is there a reason that one of these proposed additions says "exact same types", but the others do not?

mbudiu-vmw:
This is ready for re-review.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3471,6 +3485,11 @@ void pop_front(int count) {
 }
 ~ End P4Pseudo
 
+Two header stacks can be compared for equality (==) or inequality (!=)
+only if they have the same element type and the same length.  Two
+stacks are equal if and only if all their corresponding elements are
+equal.
```
jafingerhut: Actually, one comment here.  Do we want to explicitly say that the 'nextIndex' hidden state that exists at least during parsing time is _not_ relevant for equality comparisons between header stacks?  I know that is implicit in the way it is worded here, but explicit might be better.

mbudiu-vmw: This makes me realize that the situation is pretty ugly: since assignment is supposed to preserve the nextIndex field, this means that this should field should be even preserved between architectural blocks. So the nextIndex left over in the ingress parser could be accessed in the egress parser! 

It probably should not be used for comparisons, although one could argue either way.

jafingerhut: In PSA, the only way to preserve an already-parsed header from ingress to egress would be to carry the header in user-defined metadata.  Kind of an odd choice, but I guess not currently disallowed by the architecture spec.

jafingerhut: Have you considered the idea of allowing constructs that are not in the base P4_16 language spec in the intermediate representations that you can spit out in the compiler?  That would allow you to explicitly define a my_hdr_stack_variable.nextIndex field in that slightly extended language, even if it is not available to P4_16 developers.

Or just go for it and make such a field part of the language explicitly, available everywhere, but not for P4_16 programmer convenience, but because it is helpful in some compilers?

PR #545, opened 2017-12-27
fix action_profile example in psa text

hanw:


p4-16/psa/PSA.mdk
```diff
@@ -1852,7 +1852,7 @@ control Ctrl(inout H hdr, inout M meta) {
 
   action foo() { meta.foo = 1; }
 
-  action_profile ap(32w128);
+  ActionProfile(32w128) ap;
```
jafingerhut: Out of curiosity, do these examples compile without error if you leave off the "32w" prefix on the constants?  Those prefixes are unnecessary in most contexts in a P4_16 program, and I personally prefer to leave them off unless they are necessary for some reason, or actually serve as some kind of aid to human understanding if bit widths are subtle or tricky somehow.  This is not a case where I would consider the bit widths of the constants to be an aid to human understanding.

hanw: The '32w' prefix can be removed without causing any error.

PR #546, opened 2018-01-03
Update telemetry specs

jklr:
- Remove telemetry from gh-pages branch docs folder
- Updated the pointers to telemetry specs


jklr:
@cc10512 sorry, didn't notice that. Fixed now.

PR #552, opened 2018-01-18
PSA Add Appendix on packet ordering recommendations in PSA devices

jafingerhut:


p4-16/psa/PSA.mdk
```diff
@@ -2572,3 +2543,131 @@ Certainly if there are long durations of time when egress decides to
...
+parallelism that can be taken advantage of in a variety of ways, that
+can lead to cheaper implementations if these recommendations are not
+followed.  We recommend that developers selecting P4 devices ask their
+designers about these issues, if those developers consider the issues
+important for their purposes.
+
+Recommendation 1: Packets that arrive on the same input port should
+begin ingress processing in the same relative order as they arrived.
+
+Recommendation 2: Packets that go out on the same output port should
```
vgurevich: Maybe the other way around, i.e. they should be transmitted in the same order they begin egress processing, but ultimately both are equivalent. For me it uis just easier to think from left to right :)

p4-16/psa/PSA.mdk
```diff
@@ -2572,3 +2543,131 @@ Certainly if there are long durations of time when egress decides to
...
+`multicast_group` value, that it may be especially difficult to
+satisfy Recommendation 4.  That recommendation is intended to apply
+only when the set of copies to be made for the multicast group has
+remained unchanged for a period of time.
+
+If an implementation satisfies recommendations 1, 2, and 4, then
+multicast traffic assigned to the same class of service will maintain
+its relative order through the device, when multicast group membership
+has been stable for long enough.
+
```
vgurevich: I might also mention that no specific assumptions should be made with regards to the relative order between multicast and unicast packets. In most modern PREs they are queued and processed somewhat separate from each other, mostly because it is understood that a multicast packet generally spends more time inside the PRE while it is being replicated N times.

p4-16/psa/PSA.mdk
jafingerhut: I have rephrased Recommendation 2 for 'left to right' ordering :)

p4-16/psa/PSA.mdk
jafingerhut: I added a few sentences mentioning that there is no such recommendation for relative ordering of unicast vs. multicast packets in PRE, with a couple of reasons to motivate why.

p4-16/psa/PSA.mdk
```diff
@@ -2572,3 +2543,139 @@ Certainly if there are long durations of time when egress decides to
...
+its relative order through the device.
+
+Recommendation 4: PRE multicast packets (i.e. those that follow the
+"Make 0 or more copies" path in the pseudocode of section
+[#sec-after-ingress]) that arrived on the same ingress port, did
+ingress processing once, were sent to the PRE with the same pair of
+values (`class_of_service`, `multicast_group`), for copies destined to
+the same egress port, and with the same pair of values for
+(`egress_port`, `instance`), should begin egress processing in the
+same relative order as they began ingress processing.
```
cc10512: Aren't the copies "spawned" in the PRE? Therefore, they have not gone through ingress processing, so I suppose the relative order refers only to the source packet for multicast. If the copies are to be ordered, then they need to inherit the ordering from the source.

p4-16/psa/PSA.mdk
```diff
@@ -2572,3 +2543,139 @@ Certainly if there are long durations of time when egress decides to
...
+1. Expectations of hosts
+
+While the Internet Protocol does not have strong ordering requirements
+for sequences of packets sent by one host to another, there are still
+widely deployed implementations of TCP that lead to significantly
+degraded throughput when the network frequently delivers packets to
+the receiver in a different order than they were transmitted.  While
+significant research and development has been done to mitigate this
+issue, e.g. later Linux TCP implementations (since 2011 when version
+2.6.35 was released) are much more resistant to this problem than
```
cc10512: Nit: s/resistant/resilient

p4-16/psa/PSA.mdk
```diff
@@ -2572,3 +2543,139 @@ Certainly if there are long durations of time when egress decides to
...
+This reason is much less significant than the previous one.  We
+mention it here primarily so implementers of networking protocols are
+aware of the issue.  This issue is only relevant for a relatively
+small fraction of network protocols.
+
+Some networking protocols add sequence numbers to packets, and require
+either dropping packets that arrive out of sequence number order at a
+later network point (e.g. GRE with sequence numbers enabled), or with
+a looser check that allows some amount of network reordering to occur
+without dropping the packets (e.g. IPsec).  If a device implementing
```
cc10512: Nit: s/If/When/ 
and adjust the rest of the sentence?
"When a device implements these protocols and does the ...

p4-16/psa/PSA.mdk
jafingerhut: I have tried rewording Recommendation 4 a little in order to make the original packet vs. copy distinction more explicitly.

p4-16/psa/PSA.mdk
jafingerhut: done.  Resilient is definitely a better word here.

p4-16/psa/PSA.mdk
jafingerhut: Done

jamescoole-cisco:
Since a subset of applications will care about ordering, I think it would be good to recommend or require that implementors provide a way for users to request that the compiler preserve ordering per-application. A simple way to expose this would be a compiler switch, but a `#pragma` like used for [OpenCL extensions](https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/EXTENSION.html) might be better so that the requirement travels with the application source.

This seems like an improvement over "asking the vendor" for a couple of reasons.

First, without being able to specify this as a requirement in the source, portability becomes a function of whether the application compiles on a given target _and_ whatever response the user got if/when they asked the switch vendor about ordering guarantees.

Also, for some targets whether these recommendations are met might not come directly from the device architecture, and might instead be possible with some tradeoffs. For example, a high-level synthesis tool may be able to increase parallelism when ordering isn't required by the program, but could support ordering when required by linearizing the pipeline, at the cost of latency and/or area.

jafingerhut:
@jamescoole-cisco James, definitely reasonable suggestions there.  The only part I am hesitating on is a good way to define and name such annotations/pragmas in the source code.  I could imagine starting with a separate PSA-standard name for each of the 4 recommendations described in the text here, but suspect that there might be variations on them that will end up being useful.

Maybe we just stick a stake in the sand and give PSA-standard names to these 4 recommendations for now, and handle future ones as they become clearer?

jafingerhut:
@jamescoole-cisco I will create a separate issue for your questions, in hopes of merging in this PR as it is.  Happy to continue the discussion there.

PR #554, opened 2018-01-19
PSA Add section on atomicity of control plane operations

jafingerhut:


jafingerhut:
Not sure who else might be interested in reviewing this @cc10512 but feel free to add others, or suggest others if Github does not permit you to add them yourself.

p4-16/psa/PSA.mdk
```diff
@@ -2249,6 +2249,64 @@ The control plane should provide a
 
 ```
 
+
+# Atomicity of control plane API operations
+
+All table add, delete, and modify operations must be atomic relative
```
vgurevich: Since this is a recommendation, perhaps we should replace the word must with "should". There are so many situations I've encountered in the real life, where this is "kinda" the rule, but there are exceptions. And in many cases they are not fatal -- all the control plane needs to do is to make sure that this entry that is being changed in some non-atomic fashion doesn't get hit during that process. There are many ways to do it.

jafingerhut: Are these only recommendations?  I thought the intent is that these were going to be requirements for PSA implementations.

cc10512: Indeed, my recollection is also that we are going to require atomic semantics for the table operations. I do not think we should water down this requirement.

p4-16/psa/PSA.mdk
```diff
@@ -2249,6 +2249,64 @@ The control plane should provide a
...
+adding an empty group, deleting and empty group, or modifying the
+action parameters of an action added earlier to a group, should all be
+atomic.
+
+Also, a control plane read, or write, of a single element of a
+Register array should be atomic, and behave as if it occurred before
+or after (but not during) any P4 program's section of code labeled
+with the `@atomic` annotation.  There is _no_ control plane operation
+on a Register that can atomically read, then modify the value, then
+write the modified value.  If you desire such a capability, you should
```
cc10512: I would split this last sentence in the paragraph into a separate note (programmer note?).

jafingerhut:
@cc10512 I would recommend that we hold of merging in this PR until after some of the questions in this issue have been resolved: https://github.com/p4lang/p4-spec/issues/556

cc10512:
I was hoping that I addressed at least one :)

p4-16/psa/PSA.mdk
jafingerhut: I'm not sure if this is worth mentioning in the PSA spec, but realize that a PSA implementation _could_ be done where "simple enough" tables were automatically atomic w.r.t. single control plane operations as proposed here, but "compex cases" (whatever that means for an implementation, e.g. too many bits of action parameters, or keys have too many bits, etc.) would by default be rejected as non-atomic by the compiler, but the implementation could also provide table annotations where the P4 programmer could explicitly say "I am OK if this table doesn't have atomic control plane operations, because my control plane code is written to handle that."

Because a control plane might need to be written in a significantly different way for a table that didn't provide this atomicity guarantee, vs. one that did, it seems much safer to require the programmer to "opt in" for the less safe behavior, than for the default to go the other way.

jafingerhut:
Do we maybe want to add an exception here for tables that have fields with the match_kind `range`?

At least one way such a thing is implemented with hardware TCAM implementations is to break up an arbitrary range, e.g. values in the range [1, 7] for a 16-bit field, into one or more power-of-2-aligned ranges, e.g. [1, 1], [2, 3], [4, 7] combine to make [1, 7], and each of those can be represented with a single TCAM value/mask.

I don't know any straightforward way that a PSA implementation doing something like this could ensure that control plane additions of a single entry to such a table would be atomic in the way the current text proposes.  It _could_ be done satisfying the weaker constraint of "every packet will match against the old contents, the new contents, or something in-between, where the something in-between has a subset of the full range specified".

jafingerhut:
I have created this issue for a related topic that came up during the 2018-Jan-31 PSA working group meeting: https://github.com/p4lang/p4-spec/issues/567

As stated there, my plan is _not_ to include any mention of annotations to mark some tables as "OK not to restrict control plane updates to be atomic for this one" in this PR.  I think it is useful to merge this PR before those questions are addressed, and they make take time, especially given that there may be P4 Runtime API implications there.

PR #558, opened 2018-01-22
PSA Editing of Digest extern details

jafingerhut:


p4-16/psa/PSA.mdk
```diff
@@ -2212,46 +2212,59 @@ program could discard the most significant bits of the timestamp so
...
+plane can distinguish each of these messages by the name of the digest
+instance in the P4 program.
+
+Since high speed PSA implementations are expected to be able to
+generate digests much faster than control software can consume them,
+it is expected that loss of such digest messages will occur if the
+data plane generates them too quickly.  It is recommended that PSA
+implementations maintain a count of digest messages that the data
+plane requests, but do not reach the control plane, independently for
+each digest instance.
```
cc10512: Does this imply that the control plane needs to acknowledge the request of a digest? Or did you mean that the PSA data plane keeps a counter of digests, and the control plane subtracts the ones it received? Is this a counter per instance? How many instances do we allow? Are these counters just like table counters or maintained by the hardware in the data plane?

p4-16/psa/examples/psa-example-digest.p4
```diff
@@ -125,67 +119,72 @@ parser EgressParserImpl(packet_in buffer,
...
-        meta.digest_id = 0;
-        meta.mac_learn_digest.srcAddr = hdr.ethernet.srcAddr;
-        meta.mac_learn_digest.ingress_port = istd.ingress_port;
+    Digest<mac_learn_digest_t>() mac_learn_digest;
+
+    action unknown_source () {
+        mac_learn_digest_t tmp;
+        tmp.srcAddr = hdr.ethernet.srcAddr;
+        tmp.ingress_port = istd.ingress_port;
+        mac_learn_digest.pack(tmp);
```
cc10512: Is the name that the control plane uses then `mac_learn_digest`? Might be good to state explicitly.
I looked for the P4Runtime digest proto, it is only a proposal (https://github.com/p4lang/PI/pull/242), not sure whether it has been approved or not.

p4-16/psa/PSA.mdk
jafingerhut: Note: Providing these drop counters is a recommendation in the proposed text, not a requirement.  That said, I wouldn't mention them if nobody would ever find them useful, e.g. I expect they could be very useful for someone debugging a system, including control plane software, who is wondering why the system is misbehaving in some way, and trying to narrow down the cause of the problem.

I don't think such counters implies that the control plane needs to ack digest requests at all.

There are multiple ways to implement such loss counters, depending on the implementation, but for example, if you have an ASIC and local driver/agent software involved in getting digest messages to a controller, implementing a drop count in the ASIC, and a separate drop count in the driver/agent software, and adding up those two numbers, should do the job.

The text proposes "independently for each digest instance".

The number of Digest instances supported by an implementation I expect would be implementation-dependent, just as the number of tables, number of Counter, Meter, and Register instances, number of classes of service, number of ports, etc. is implementation-dependent.

These drop counters I would expect to be one of the cheapest parts of a Digest extern implementation.  Far more expensive I expect would be on-ASIC buffers to store multiple digest messages, and/or DMA hardware to get them to the local driver software at a reasonably high rate, without dropping too many when they might occur in short bursts.

p4-16/psa/examples/psa-example-digest.p4
jafingerhut: These sentences are part of the proposed PR right now.  Let me know if  you think it answers your question:

"A digest contains two pieces of information: a name, which is derived
from the name of the digest object instance in the P4 program, and the
digest message.  The digest message can contain any values from the
data plane."

I suppose that text might be interpreted to imply that a string of ASCII characters must accompany each digest message, but really all that is intended is that the controller be able to distinguish digest messages from different instances in a P4 program from each other.  At _some_ level of the controller software it should be able to get the name of the Digest instance, although probably only a small integer id would be carried with each Digest, or some other low-overhead mechanism.

p4-16/psa/PSA.mdk
cc10512: Thanks for the details. I'm still getting used to the fact that you consider the driver/agent part of the "hardware" and it is fair game to get it involved.

p4-16/psa/examples/psa-example-digest.p4
cc10512: The sentence you quoted is exactly what triggered my comment. I was wondering what is the derived name in this context. I would suggest that we replace "a name" in that sentence with "an identifier".

p4-16/psa/PSA.mdk
jafingerhut: I think of the driver/agent as a part of the system where digest notification messages might have to be dropped, and so if you want to know how many were dropped in the system between the source and the controller, it is best to count that reason, too.  A different kind of PSA implementation might only have one place where they could get dropped.

p4-16/psa/examples/psa-example-digest.p4
jafingerhut: I was thinking of the same instance name used to identify the instance by the P4 Info file and used by the controller to name the instance in the P4 program.  Tables have them, register instances, etc., so digest instances should, too.

jafingerhut: OK, attempted to address this comment, and thought of a couple paragraphs that may be worth saying something more about Digest extern behavior while I was at it.  Feel free to "uncheck" Approval status if those additions seem wrong to you.

jafingerhut:
@hanw @cc10512 From discussions during and just after the 2018-Jan-31 PSA work group meeting, it sounds like a reasonable proposal here is to change the PSA spec, and the example program, to only require that Digest instances can be instantiated in IngressDeparser and EgressDeparser control blocks.

Even with such a change, a PSA implementation could choose to also support instantiating them in Ingress and Egress control blocks (the PSA spec as written for all other externs only mentions places where it is required they be supported -- it explicitly allows implementations to go above and beyond the minimum requirements).

jafingerhut:
@hanw @cc10512 This PR has now been updated by requiring Digest instantiations in IngressParser and EgressParser controls only.  The example program has been updated to instantiate a digest in the IngressParser control, instead of Ingress.

cc10512:
LGTM! Thanks @jafingerhut.

PR #562, opened 2018-01-25
PSA Proposed description of PortId_t translation in PSA

jafingerhut:


jafingerhut:
Note: This is not the only way to achieve the goals described with these changes, of including PSA-specific field values in packet headers between controller and data plane, and handling PortId_t values.  Comments are welcome.  I just want to get this out there in its current form to see if it is anywhere near the ballpark of what we want in PSA.

jafingerhut:
@samar-abdi I am not sure if I am able to add you as a reviewer in Github -- didn't find your user name on the list of possible reviewers to add, so notifying you with this comment.  Please take a look and provide your comments.

p4-16/psa/PSA.mdk
```diff
@@ -309,6 +310,134 @@ Additional supported match_kind types
...
+A PSA data plane implementation that supports the P4 Runtime
+API[^P4RuntimeAPI] is expected to include software called a "P4
+Runtime API agent" that encapsulates details specific to the PSA
+implementation, enabling a P4 Runtime API controller to more easily
+control multiple devices with different PSA implementations.  We will
+refer to these pieces of software as an "agent" and "controller" for
+brevity.
+
+As mentioned in section [#sec-psa-type-definitions], different PSA
+implementations are expected to differ in the minimum number of bits
```
cc10512: Suggestion on rephrasing: different PSA implementations are expected to customize the size of the data types that refer directly to physical entities, such as ports.

p4-16/psa/PSA.mdk
```diff
@@ -309,6 +310,134 @@ Additional supported match_kind types
...
+
+Because of this translation, it is recommended that values of type
+`PortId_t` be treated similarly to values of an `enum` type.
+Comparing two values of this type for equality or inequality is
+reasonable, as well as assigning the values to other variables or
+parameters of the same type, but nearly any other operations are prone
+to error.  When matching values of type `PortId_t` as part of a table
+key, always match a complete value exactly, or wildcard every bit of
+the value (i.e. a `ternary` match kind with all bits wildcard, or an
+`lpm` match kind with prefix length 0).  If you attempt to do any of
```
cc10512: I believe P4Runtime disallows such behavior: all wildcards or LPM with prefix of 0. @antoninbas can confirm.

p4-16/psa/psa.p4
```diff
@@ -65,6 +65,28 @@ const   CloneSessionId_t PSA_CLONE_SESSION_TO_CPU = unspecified;
...
+ * fields may be used in a single P4 header definition, even on P4
+ * implementations that restrict headers to contain fields with a
+ * total length that is a multiple of 8 bits. */
+typedef bit<16> PortIdInHeader_t;
+typedef bit<32> MulticastGroupInHeader_t;
+typedef bit<16> CloneSessionIdInHeader_t;
+typedef bit<8>  ClassOfServiceInHeader_t;
+typedef bit<16> PacketLengthInHeader_t;
+typedef bit<16> EgressInstanceInHeader_t;
+typedef bit<64> TimestampInHeader_t;
```
cc10512: Should we explicitly state that we do not need translation annotations for these other types?

p4-16/psa/PSA.mdk
jafingerhut: @cc10512 I am pretty sure P4Runtime allows one to create table entries that completely wildcard ternary fields, or do length 0 prefix of lpm fields -- it is just that they represent it in a way where the field isn't even mentioned in the P4Runtime API message.

p4-16/psa/psa.p4
jafingerhut: @vgurevich @antoninbas @samar-abdi Any thoughts on whether any of these other types would need conversion of values?

Can the control plane API just use whatever multicast group numbers and egress instance numbers it wants?  I assume different implementations will have different numbers of these things that they support, and some might have more restricted numerical ranges than others.  Does a P4Runtime API controller have a way to discover such restrictions on these values for different PSA devices in a heterogenous network?

Do we expect ClassOfService_t values to be numerically consecutive in a range [0, N] for all PSA devices, and should say that in PSA?

Same question for CloneSessionId values.

I _think_ we are safe from such issues for PacketLength_t and Timestamp_t values, i.e. some implementations might have different maximum PacketLength_t values they actually support, but all of them will be [my_min, my_max] for packet length support for some my_min, my_max integers, and timestamps should be all X-bit-wide values for the value of X supported by the implementation.

p4-16/psa/PSA.mdk
jafingerhut: I have replaced it with this "are expected to customize the size of the data types that refer directly to those objects, i.e. ports, multicast group ids, etc.".  Something just seems a little off to me to refer to multicast group ids and instance ids as physical entities.

p4-16/psa/psa.p4
samar-abdi: Very likely class of service will need translation. It has the same issue as ports. Different controllers may have different notions of classes of service and their own numbering scheme for them. The switch config handler on the target will map the controller's class of service to the ASIC's. Other metadata values are either programmed by the controller (mcast_group/egress instance) or have universally agreed values (packet_length/timestamp) IMO.

jafingerhut: So we should explicitly document that type ClassOfService_t fields will be translated between controller and P4 data plane programs, as PortId_t fields will be, but _no others will_ ?

Do we expect, and should we perhaps document, that for all of the following types, the data plane and the control plane will both support _all_ numerical values in the range [0, N] for whatever maximum value of N is supported by the PSA device (which can vary from one PSA implementation to another)?

```
typedef bit<unspecified> MulticastGroup_t;
typedef bit<unspecified> CloneSessionId_t;
typedef bit<unspecified> EgressInstance_t;
```

p4-16/psa/PSA.mdk
```diff
@@ -309,6 +310,134 @@ Additional supported match_kind types
 ```
 
 
+## Data plane vs. control plane data representations
+
+A PSA data plane implementation that supports the P4 Runtime
+API[^P4RuntimeAPI] is expected to include software called a "P4
+Runtime API agent" that encapsulates details specific to the PSA
```
samar-abdi: suggest: ...include software called a "P4 Runtime Agent" that enables runtime programming of the PSA device from a controller. A controller may control  multiple devices with different PSA implementations.

p4-16/psa/PSA.mdk
```diff
@@ -309,6 +310,134 @@ Additional supported match_kind types
...
+resources for things like table keys and action parameters if the data
+plane stores only the fewest number of bits required for values of
+these types, for that implementation.  For example, an implementation
+that defines `PortId_t` as `bit<6>` instead of `bit<16>`, and can take
+advantage of this difference, could save 10 Mbits of storage in a
+table with a million entries[^CacheCost].
+
+[^CacheCost]: While 10 Mbits sounds tiny to one accustomed to computers with hundreds of gigabytes of DRAM, the highest speed PSA implementations are ASICs that must keep tables in on-chip memories, similar to caches in general purpose CPUs.  The Intel i9-7980XE released in 2017 has 198 Mbits of on-chip L3 cache shared by its CPU cores.  Among Intel processors in Intel's 7th generation Core released in 2017 with at least 100 Mbits of L3 cache, they all cost close to $9 per Mbit of L3 cache.  https://en.wikipedia.org/wiki/List_of_Intel_microprocessors
+
+The P4 Runtime API is planning to use 64-bit quantities to hold values
```
samar-abdi: s/is planning to use/uses

p4-16/psa/PSA.mdk
```diff
@@ -309,6 +310,134 @@ Additional supported match_kind types
...
+of the types listed in section [#sec-psa-type-definitions].  For
```
samar-abdi: suggest: ...listed in section [#sec-psa-type-definitions] to simplify the manipulation of the metadata quantities in the agent software.

p4-16/psa/PSA.mdk
```diff
@@ -309,6 +310,134 @@ Additional supported match_kind types
...
++ Control plane operations on tables, where values of these types may
+  be included as part of the key, or as an action parameter.
++ Packets sent to the CPU ("packet in" from the controller's
+  perspective), or received from the CPU ("packet out" from the
+  controller's perspective).
++ Fields in a `Digest` extern notification message (section
+  [#sec-packet-digest]).
++ Fields in the data contents of a `Register` array (section
+  [#sec-registers]).
++ Match fields of an entry in a `ValueSet` extern instance (section
```
samar-abdi: Is this still applicable given that we were expecting to remove value_set from PSA 1.0?

p4-16/psa/PSA.mdk
```diff
@@ -309,6 +310,134 @@ Additional supported match_kind types
...
+many bits as `PortId_t` does.
+
+Because these types are guaranteed to
+be a multiple of 8 bits long, any combination of them may be included
+in a P4 header type definition, as long as the other fields in the
+header satisfy the multiple of 8 bits restriction.  The controller or
+P4 program generating packets with such headers should fill in any
+most significant "padding" bits with 0.  In a P4 program, this can be
+done with a normal assignment statement, where the value on the right
+hand side is cast to the wider `InHeader` type.
```
samar-abdi: suggest: Similarly, for an assignment of a PortIdInHeader_t field to a PortId_t field, there will be a truncation of the excess most significant bits as part of the cast.

p4-16/psa/psa.p4
samar-abdi: Yes for documenting translation of ClassOfService_t field values. The P4Runtime annotation will be `cos_translation("controller_bitwidth:32")`.

As for your second point, I suggest : "The values of multicast_group, clone_session and egress_instance are completely in the purview of the controller, and are limited only by the bitwidth of the respective types on the PSA device(s) being controlled by the controller."

I believe it is implicit that if a psa.p4 has bit<7> multicast_group_id, it will allow all values from 0 to 127.

jafingerhut: I know it is common for things in computers to have sizes that are powers of 2, but it sometimes happens that people implement SRAMs with a number of entries that aren't powers of 2 (or a collection of them whose total number of entries is not a power of 2, more likely, but have a common address space spanning all of them).

I would hate for a controller to have an easily avoided bug if this turns out to be the case, because it made such power-of-2 size assumption.

p4-16/psa/PSA.mdk
jafingerhut: I fall into the passive voice too easily by habit.  For that and other reasons, taking your suggested wording here.

p4-16/psa/PSA.mdk
jafingerhut: changed

p4-16/psa/PSA.mdk
jafingerhut: updated

p4-16/psa/PSA.mdk
jafingerhut: good catch.  Removing that from this PR.

p4-16/psa/PSA.mdk
jafingerhut: I added a sentence like this.  Thanks.

jafingerhut:
@samar-abdi 
You suggested this annotation for values of type ClassOfService_t: `cos_translation("controller_bitwidth:32")`
and this annotation is currently in the PSA spec for values of type PortId_t:
`@port_translation("controller_port_bitwidth : 32")`

Two questions for you:

1.

Is it intentional that one has "controller_bitwidth" but the other "controller_port_bitwdith"?  Would it be reasonable if they both used "controller_bitwidth"?  Either that, or perhaps cos_translation should use "controller_cos_bitwidth"?

2.

Why 32 bits for the bitwidths in your examples?  This seems a bit confusing, given that we state that the controller and agent will use 64-bit types for these in their code, and we currently propose these widths for the InHeader variant of the types in this PR:
```
typedef bit<16> PortIdInHeader_t;
typedef bit<8>  ClassOfServiceInHeader_t;
```
Should the bitwidths be 64, to match the controller/agent software?  Or perhaps they should match the 16, 8 numbers in the InHeader_t type definitions?

jafingerhut:
@cc10512 @samar-abdi In a few minutes I will push a commit to this PR saying that only PortId_t and ClassOfService_t (and their InHeader_t variants) require numerical translation between controller and data plane.

For all other types, I am adding a requirement that the data plane support consecutive values in the range [0, max], where max is PSA-implementation-specific, can be different for each of the types, and _need not be a power of 2_.  Take a look if any of that sounds wrong.

p4-16/psa/PSA.mdk
antoninbas: @jafingerhut is correct; we do allow wildcards for ternary & LPM but require a certain formatting for the messages.

jafingerhut:
@samar-abdi I have updated the one example P4 program that had an annotation on a PortIdInHeader_t field in a header to use the new simpler annotation you suggested in private email, i.e. `@p4runtime_translation("port")`.  I also added the `@p4runtime_translation("cos")` annotation you suggested for values of type ClassOfService_t (and its 'InHeader' variant).

jafingerhut:
@cc10512 @samar-abdi I am ready for this to be merged in.  The biggest remaining question for whether any of this changes for PSA v1.0 in my mind is whether people are comfortable with the bit widths proposed for the types that are intended to be big enough to handle all PSA implementations.

There is an open issue for that here: https://github.com/p4lang/p4-spec/issues/571 and any changes suggested there are a small edit to the existing document.

jafingerhut:
Going ahead and merging this in.  We haven't labeled anything version 1.0 yet, so further edits are still possible to this before then.

cc10512:
@jafingerhut  since we haven't heard any complaints on the sizes in a bit more than a week, I think we're good to go with this. I an pretty sure we'll revisit this once we have more experience with applications using the feature.

PR #569, opened 2018-02-01
PSA Move parser value sets to open issues

jafingerhut:


jafingerhut:
@cc10512 Please go ahead and merge this in if you think the note in open issues is reasonable.

PR #574, opened 2018-02-09
PSA Implementations may limit hash divisor and random range to power …

jafingerhut:
…of 2


jafingerhut:
@cc10512 I am pretty sure this PR aligns with a brief discussion or two we have had on this topic on at least one PSA work group meeting.

PR #575, opened 2018-02-10
PSA Explicitly forbid empty ActionSelector groups to be used by a tab…

jafingerhut:
…le entry

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -2031,7 +2031,7 @@ In this example, the table has a match key consisting of an LPM on header field
 `h.f`. A second match type `selector` is used to define the fields that are
 used to look up the action specification from the selector at runtime.
 
-A table with an action action selector implementation consists of entries that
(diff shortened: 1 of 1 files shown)


jafingerhut:
@antoninbas We can leave this unmerged until we know more.  If you are having such discussions, I would also be curious if the P4 Runtime API, and PSA, can say that you cannot "remove a member" while it is in active use.  That restriction would apply for both ActionProfile and ActionSelector externs, I believe.

jafingerhut:
I would prefer if PR https://github.com/p4lang/p4-spec/pull/662 was merged _instead of_ this one.  It has a more positive description for control plane software that wishes to have empty action selector groups, including a suggestion of how to implement it (as far as the control software is concerned -- even if the data plane itself does not support it directly).

If that other one is merged, this one should definitely be closed.

jafingerhut:
Closing this PR, since the changes made with PR https://github.com/p4lang/p4-spec/pull/662 were preferred over these.

PR #578, opened 2018-02-10
PSA Fix some typos, clean up a few things, progress towards v1.0

jafingerhut:


jafingerhut:
@cc10512 I don't think there is anything controversial here.  Mostly cleanup editing.  Please do take a look at the diffs to see if you find anything questionable, and we can fix it.

PR #581, opened 2018-02-13
PSA Updates and additions to introduction

jafingerhut:


jafingerhut:
@cc10512 I don't often read the introductory test, but when starting from the beginning of the PSA, I found it missing some things that seem like it would be good to say near the front.  Let me know what you think of these proposed changes.

p4-16/psa/PSA.mdk
```diff
@@ -58,38 +58,52 @@ interface ports.
 # Target Architecture Model
 
 
+As an analogy, the PSA is to the P4~16~ language as the C standard
+library is to the C programming language.  PSA defines a library of
+types, P4~16~ externs for frequently used constructs such as counters,
```
cc10512: types and externs? We do define a set of data types too.

p4-16/psa/PSA.mdk
```diff
@@ -58,38 +58,52 @@ interface ports.
...
+meters, and registers, and a set of "packet paths" that enable you to
+write P4 programs that control the flow of packets in a packet switch
+that has multiple ports, e.g. dozens of Ethernet ports.  By following
```
cc10512: Are you equating the "packet paths" with the pipeline? I kind of see those a bit different. It is true that we said right in the beginning that: "P4 programs specify how the various programmable blocks of a target architecture are programmed and connected". However, it is really PSA that defines the structure of programmable blocks and how they are connected. Somehow we missed this in the different iterations.

I see that you expand on this below, but it is never quite clear that the architecture defines these blocks -- in P4 no less :)

p4-16/psa/PSA.mdk
```diff
@@ -2751,6 +2765,48 @@ scheduling algorithm implementations are all built with a finite
...
+Similarly for the ingress deparser.  By making this an explicit and
+separate control block, you now have full control over exactly what
+data about a packet is included when it is sent to the packet buffer,
+and what is not.  In P4~14~, it was implicit that "all packet metadata
+used somewhere in egress code" was carried along with each packet.
+This can still be done in P4~16~ programs for the PSA, but now you
+must be explicit in doing so.  With PSA, you now have a way to
+restrict how much data is carried with each packet, which can be
+important if the I/O bandwidth of the packet buffer becomes a
+bottleneck.
```
cc10512: I like what you've done here.

p4-16/psa/PSA.mdk
jafingerhut: It does say "PSA defines a library of types, ...".  Did you mean you think that "types" should not be mentioned?  If you meant that they should, and it isn't prominent enough, let me know.

p4-16/psa/PSA.mdk
jafingerhut: The packet paths are what we call packet paths later in the document, e.g. to/from ports, normal unicast/multicast packets through the packet buffer, resubmit, recirculate, and clone.  Those are fairly specific to PSA, and things that I suspect would look somewhat different if there were a "Portable NIC Architecture" document.  The PSA defines these packet paths, but mostly in English and pseudocode.

p4-16/psa/PSA.mdk
cc10512: I guess the comma between types and externs is throwing me off. It is a long sentence that takes a while to absorb. I would probably rephrase it as:
PSA defines a library of types, P4~16~ externs, and a set of "packet paths" that enable you to
write P4 programs that control the flow of packets in a packet switch that has multiple ports. Examples of externs include frequently used constructs such as counters, meters, and registers.

p4-16/psa/PSA.mdk
cc10512: Right, but also the role of PSA is to define the programmable blocks, and while we do discuss these later, it seems that we lead with defining most of the other things.

p4-16/psa/PSA.mdk
jafingerhut: Sorry, I merged this, perhaps crossing in flight with your comments.  Please do feel free to edit these additions for clarity.

PR #582, opened 2018-02-13
PSA Move extern-style parser value set examples to examples/historica…

jafingerhut:
…l directory


jafingerhut:
Pardon my unnecessary rush if you would prefer that these file moves _not_ be made, but I would like to be able to point people at the 'examples' directory for sample programs, and would rather not confuse them with these probably-soon-obsolete parser value set examples that used the older approach of P4_16 externs that people did not like the syntax/flow of.

PR #583, opened 2018-02-13
Psa more open issues

jafingerhut:


jafingerhut:
This may be more than we want in PSA v1.0 open issues section.  The thinking here is _not_ that we want to create an entry there for every open issue or future addition we might make.  I selected a few open issues that seemed to me like things that some people would look for ways to do in PSA, but we haven't defined yet.  As long as they know we are considering adding them in the future, they may be less likely to ask, and/or more likely to help define them!

p4-16/psa/PSA.mdk
```diff
@@ -2473,6 +2473,56 @@ a common P4 program.
...
+packet that would be useful inputs to multiple congestion control
+algorithms.  One possibility is a hash of the packet's "flow id",
+often implemented as a hash of packet header fields like IP source and
+destination address, IP protocol, and optionally TCP/UDP source and
+destination ports.  Given that P4 programmable devices can implement
+network protocols other than IP, including custom ones, a more general
+mechanism is desirable in PSA devices.
+
+
+## Observation and control of congestion
```
cc10512: Cut and paste error for the title? Do you want to rename this to "Observing packet ordering"?

jafingerhut: This sub-section only deals with observing whether a packet experienced congestion, e.g. for implementing ECN header modifications in egress, or potentially feeding inputs to a congestion control algorithm from ingress.  I don't see any connection to packet ordering here.

I know there is an open issue about potentially adding ways in the future to tell the compiler restrictions it should follow on packet ordering for a particular P4 program.  I did not attempt to create an open issue in the document for that, mainly because it seemed like something fewer people would care about (not 0, but fewer).

cc10512: You have the same heading for both ECN and timestamps.

jafingerhut: Woops, disregard my previous comment.  Yes, this sub-section title is definitely wrong.  Fixing...

jafingerhut:
Thanks for the quick review and catching that mistake.  Merging now.

PR #586, opened 2018-02-14
P4-16 Restrictions on new proposed `value_set` language construct

jafingerhut:


jafingerhut:
This comment is longer than the proposed diffs, but hopefully makes them easy to think about :-)

Proposed additions to Appendix D "Restrictions on compile time and run
time calls" of the P4_16 language specification, for the new proposed
`value_set` language construct.

Note: There is nothing in these proposed changes that actually defines
or explains what a `value_set` is.  Such additional text would need to
be written.


Rationale: `value_set`s as proposed have many similarities to P4
tables.  They are stateful elements, built into the language.  Each
instance has a name specified in the source code.  For both, their
contents can only be modified via control plane API operations.

Because the "return value" of a value_set is effectively a "set type"
(see Section 7.2.8.1 "Set types" of the P4_16 specification), where it
is stated that this type "cannot be named or declared by P4
programmers", by restricting the use of value_sets in the ways
described here, there is still no need to define ways to name or
declare set types.


Thus the proposed restrictions on what a P4_16 compiler must support
are very similar to those for tables.

+ Neither tables nor value_sets can be used as a constructor parameter, nor passed as a run-time parameter.

+ Tables can only be instantiated within controls.  value_sets can only be instantiated within parsers.

+ Tables can only be "called" at run-time (i.e. their `apply` method invoked) from within a control's `apply` block.  value_sets can only be "called" by using their name inside of a parser state's `select` expression as a key set expression.

Gory detail: The statements are phrased in a way that you can think
"oh, a P4 compiler _can't_ support these operations except where
stated".  If we think of the P4_16 specification as saying "these are
the _minimum_ ways that you must support the language constructs", you
should instead interpret those phrases as "a P4 compiler _need not_
support attempting this kind of operation, but it is _allowed_ to, if
one can imagine and implement reasonable semantics for it, which might
not be obvious."

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6591,8 +6594,10 @@ type, or no value at all (specified by a return type of `void`).
 
 The next table lists restrictions on what kinds of calls can be made
 from which places in a P4 program.  Calling a parser, control, or
-table means invoking its `apply()` method.  The row for `extern`
-describes where extern method calls can be made from.
+table means invoking its `apply()` method.  Calling a `value_set`
```
mbudiu-vmw: I would say that a value set expression evaluates at runtime to a set of values of the specified type. The evaluation reflects the latest updates made by the control-plane.

jafingerhut: I agree with that statement.  This comment is just before the table on restrictions of where different kinds of stateful objects can be "called" at run-time.  The closest thing that a value_set has to being called at run-time is to be evaluated when the enclosing select expression is evaluated.  I wasn't intending to define the behavior of a value_set here in this section.

jnfoster:
Closing as I believe this was superseded by #602

PR #601, opened 2018-03-25
adds support for saturating arithmetic

### LINKED ISSUE #600: P4_16: Support for saturating arithmetic
See https://github.com/p4lang/p4-spec/wiki/saturating.

cc10512:
Fixes #600

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1801,7 +1801,7 @@ for a table when generating the control-plane API:
 @name("acl") table t1 { ...}
 ~ End P4Example
 
-### Integers (signed and unsigned) { #sec-integers }
+### Integers (signed, unsigned, and saturating) { #sec-integers }
 
 P4 supports arbitrary-size integer values. The typing rules for the
 integer types are chosen according to the following principles:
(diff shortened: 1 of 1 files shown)


jafingerhut:
I think Calin mentioned the use case of saturating counters.  Being able to have packet and byte counters in the data portion of a Register extern might be a more motivating use case than TTL decrement?

mbudiu-vmw:
If the counters are not registers, then they autoincrement, so there is no p4 code involved.

cc10512:
The main use case is saturating counters. These are different from the counter registers. Consider a 'postcard' scenario, where you want a header field to play the role of a counter that gets incremented. For that you would likely want a counter operation that you invoke in the body of an action.

You also asked for examples of other architectures implementing saturating arithmetic. I provided the links for Intel x86 and x86_64, ARM AArch32 and, Qualcomm DSP in the proposal (https://github.com/p4lang/p4-spec/wiki/saturating#targets-supporting-saturating-arithmetic).

vgurevich:
I think that the main question is: do we need two separate types or can we simply use saturating operations on the existing types?

cc10512:
I believe a data type is cleaner and easier to understand. Also easier for the compiler to make sure that all uses of a saturating field are properly handled. And really, it doesn't cost us anything but a new type definition. And if a hardware target supports it, why bastardize a program and not support it properly?

cc10512:
BTW, @jafingerhut I didn't put in the reference implementation because we don't have a good place in the spec. However, if this proposal is approved, we should consider having appendices with programmer notes and reference implementations.

vgurevich:
I think that we need to look at the specific code examples and see. For example, would you define the ttl field in ipv4 header as "saturating_bit<8>"? Why or why not? How often do you actually need to perform casts from saturated to regular types and vice versa? 

Note, that except addition and subtraction all other operations are either exactly the same or they should not be defined. For example, how would you define shift on a saturated number? if you have saturated_bit<8> value of 200, does it mean that when you right_shift it you will get 255? I seriously doubt so. 

If you define a type, you need to take care of all these, but if you define two additional operations (you can even define special signs for them), you do not have to worry about the rest of stuff.

cc10512:
And so, rather than relying on two types, you want to teach programmers a different symbol for addition? How is that simpler?

Note that I have restricted the operations that can be executed on this data type. See lines 1957-1961.

vgurevich:
Both proposals have pros and cons. By discussing them we can find the best solution.

You can use a separate operation, you can use a special function -- it's about the language expressiveness. 

You can also use a separate type. So far, based on what I know about its usage in P4_14 (which has it, BTW), it is used ONLY to select a separate addition/subtraction (and nothing else) and is relatively confusing.

cc10512:
Fortunately, P4_16 has type checking, which P4_14 was lacking. In my view that will make the uses of these operations much easier because the compiler can really check. Indeed, you can also define a saturating operation and apply it to regular integer types, but then there is nothing for the compiler to check -- only to generate code. It will lead to confusion, because now the same field can be used in both saturating and non-saturating operations.

By defining a type, you guarantee that all uses of that field are consistent. If you want to convert, you have to put an explicit cast, and thus are guaranteed to pay more attention.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1931,6 +1931,202 @@ described below.
...
+~ Begin P4Example
+header ipv4_t {
+        bit<4>  version;
+        bit<4>  ihl;
+        bit<8>  diffserv;
+        bit<16> totalLen;
+        saturating_bit<16> identification;
+        bit<3>  flags : 3;
+        bit<13> fragOffset;
+        saturating_bit<8> ttl;
```
vgurevich: This is the primary example of how NOT to use saturating_bit<>. Essentially the definition of a standard IP header now depends on how the programmer wants to to operate on it in the program, where as one would assume it should not depend on these details.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1931,6 +1931,202 @@ described below.
...
+        bit<32> srcAddr;
+        bit<32> dstAddr;
+    }
+}
+
+control ingress(...) {
+    action a(in saturating_bit<16> value) {
+        // even if ttl == 0, following the decrement operation,
+        // ttl will continue to be 0
+        ttl = ttl - saturating_bit<8>(1);
```
vgurevich: It is not even clear where ttl and identification came from. Probably from ipv4 header, but that's not in the example.

I assume that this particular expression with an explicit cast of an int into saturating_bit<> is just for illustrative purposes -- 8sw1 might be easier to read, but not by much

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1931,6 +1931,202 @@ described below.
...
+8-bit binary signed arithmetic, when the correct answer is 130, it is
+considerably less surprising to get an answer of 127 from saturating
+arithmetic than to get an answer of −126 from modular arithmetic.
+Likewise, for 8-bit binary unsigned arithmetic, when the correct
+answer is 258, it is less surprising to get an answer of 255 from
+saturating arithmetic than to get an answer of 2 from modular
+arithmetic.
+
+P4 supports two data types for saturating arithmetic in P4_16,
+`saturating_bit<W>` and `saturating_int<W>`, for unsigned and signed
```
vgurevich: While bit<n> sounds OK and is, indeed representing the idea of a bit string first and its interpretation as an unsigned integer second, the words "saturating_bit" simply do not make a lot of sense to me, since bits do not generally saturate.

mbudiu-vmw:
I actually tend to agree with @vgurevich. If you make ttl saturated you will have to make all table keys that t matches on also saturated. Then you will have to describe how saturated works with the control-plane API.
Instead, if you make it just `bit<32>` and have a saturated addition function this is everything you need. I expect that you only need saturated addition in practice, and none of the other operators.
Once you add something to the language you can never take it out, so I think it is worthwhile to design new features conservatively based on actual needs.
If saturated addition turns out to be useful you could promote it into a new operator, e.g., a |+| b, or something like that.

cc10512:
Actually, I think that @mbudiu-vmw's comment "If the counters are not registers, then they autoincrement, so there is no p4 code involved." makes the best argument for a data type. Since you can not write P4 code to specify an operator, you would either have to define a new counter extern, that supports saturating, which is otherwise identical to the existing counters, _or_, specialize the existing extern with a saturating data type, which is already supported since our counters are parametrized on the type.

mbudiu-vmw:
Do you want to sketch the code thcounters?

mbudiu-vmw:
Sorry, the phone view of github is broken, so I closed this by mistake.

cc10512:
added an example with saturating counters.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2092,36 +2092,49 @@ else
...
+
+~ Begin P4Example
+#ifdef HAVE_SATURATING
+typedef saturating_bit<16> ByteCounterT;
+#else
+typedef bit<16> ByteCounterT;
+#endif
+
+control C(in psa_ingress_input_metadata_t  istd) {
+  Counter<ByteCounterT, bit<10>> (1023, PSA_CounterType_t.BYTES) byte_counter;
```
vgurevich: While it is totally OK to have an extern that counts something using saturating arithmetic, I do not see how useful it is to instantiate such an extern using a P4 type, since this counter can't be retrieved or anyhow accessed in the data plane and you would not even use the addition operation here. 

Are you saying that in PSA we should also instantiate counters using bit<n> to indicate the width of the underlying HW registers?

mbudiu-vmw: Indeed, the way I would do this is actually

```
extern Counter {
   Counter(bit<32> width, bit<32> size, bool saturated);
}
```

jafingerhut:
I'm probably not helping with yet another suggestion, but what if there were no new types introduced, but instead a new operator that performed saturating addition on the existing `bit<W>` and `int<W>` types?  I don't have a proposal on the best character sequence to use for this, but a concatenation of the `+` character with some other character might be appropriate.  We already have some 2-character operators, so another wouldn't be unheard of.  Maybe '+|' ?

That doesn't affect how externs work, of course, but I think the behavior of externs issue is best addressed via new extern names, and/or additional parameters to the extern constructors, personally.

mbudiu-vmw:
@jafingerhut : that's exactly what I proposed a few comments above with `|+|`.

vgurevich:
And here is the criteria that should allow us to decide, whether to use an extern function or introduce the operator:
 * If we believe that every target can reasonably and cheaply implement saturated addition/subtraction, then  we should consider the operator
* If we think that not every target can do it, then an extern function seems to be more appropriate

If we still want the new (standard) types, then we better be more than certain that everyone can easily implement them (although I still believe this is not a good idea to begin with)

PS: I do like "|+|" since it doesn't require the programmer to remember the order of "+" and "|" :)

jamescoole-cisco:
@jafingerhut +1

Swift provides special operators for a similar purpose. Overflow is a runtime error by default (not appropriate for P4, but maybe useful as a default/option in BMV2?), but you can opt-in with [special operators](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/AdvancedOperators.html) like `&+`.

jnfoster:
I'm also a fan of the add-new-operator approach over the add-new-types
approach.

The former is lighter weight and seems to solve the problem at hand.

-N

gbrebner:
Me too.  I like the operator approach much better.

cc10512:
It is indeed solving the immediate problem, and I will rewrite the proposal with operators. 

However, I will personally maintain the view that we will pay the cost in the long run - see the lack of vector types in C.

mbudiu-vmw:
@cc10512 : we can always decide to add the types to the language later if they prove to be useful in many circumstances. There's no harm having two ways to express some things.

cc10512:
Doing it different ways may lead to confusion. If we later decide to support types, I would advocate to deprecate the operators. For now, the people have spoken :)
Thank you all for the discussion and expressing you opinions.

PR #602, opened 2018-03-26
adopted and modified P4-14 description on parser value set to P4-16

hanw:
P4-14 spec has a section about parser value set that is still applicable to P4-16 with some modifications. I added that section in this PR


p4-16/spec/P4-16-spec.mdk
```diff
@@ -4378,6 +4380,26 @@ select (p.tcp.port) {
...
+a type `T` for the type parameter of the value set, the type of the value set
+is `set<T>`. The type of the value set must match to the type of all other
+`keysetExpression` in the same `select` expression. If there is a mismatch, the
+compiler must raise an error.
+
+For example, to allow control plane API to specify TCP reserved ports at
+runtime, one could write:
+
+~ Begin P4Example
+value_set<bit<16>>(4) pvs;
```
vgurevich: This definition is still not enough to generate a good API, because there is no way to name the fields, the values for which are going to be passed when value_set entries are going to be populated. This is especially important in case value_set is used to match against a tuple. Neither it is clear whether the values passed should be single or value/mask pair, range or anything else.

mbudiu-vmw: In this example there is just one field.
We can allow a value_set take a struct type, and in that case the fields can be named like the struct fields.

mbudiu-vmw: @vgurevich : this spec does not specify anything about the control-plane API. The way values are specified should be part of that other document.

mbudiu-vmw:
I think that the match_type is really very target-specific; as a proof, v1model.p4 contains the definition of `match_type` `range`. Probably this should really be an annotation:
```
@implement(ternary)
value_set<S> set;
```

vgurevich:
@mbudiu-bfn ,

I think that the significance of match_type in value_sets is no different than the one in tables and thus we need to use a very similar mechanism. It should be a part of the language, IMHO.

jafingerhut:
Regarding match_kind for value sets, do you think for multi-field value sets, e.g. that match on a struct, that it should be possible to specify a different match_kind for each field, like you can in a table?

If so, it seems we are inching our way towards a value set specified much like a P4 table, except with no list of actions.

hanw:
Attempted to write down what Vlad is saying, 
```
    value_set pvs {
        key = { p.data.da : exact; }
    }

    state start {
        b.extract(p.data);
        transition select(p.data.da) {
            pvs: accept;
            0x810: foo;
        }
    }
```

hanw:
Here is a link to the solution that the P4runtime working group came up to solve the naming problem for `field list` which is not part of P4-16. 
https://github.com/p4lang/PI/blob/master/proto/docs/p4runtime_type_serialization.md

The associated proto file for p4types.
https://github.com/p4lang/PI/blob/master/proto/p4/p4types.proto

vgurevich:
@hanw This might work, but will limit the ability to use the same value_set in multiple parser states. I would suggest using names that are not bound to anything, more like formal parameters for a function.

vgurevich:
@hanw ,
> Here is a link to the solution that the P4runtime working group came up to solve the naming problem for field list which is not part of P4-16. https://github.com/p4lang/PI/blob/master/proto/docs/p4runtime_type_serialization.md

I might be wrong but I do not see a solution for naming individual members of an unnamed field list there. If these are just numbers, this is not a good API solution for value_sets.

hanw:
I updated p4c frontend to allow struct type for the value set https://github.com/p4lang/p4c/pull/1271
to address the issue of the naming objects in the value set in runtime API. 

We (Vlad, Calin and I) had an offline discussion on the PR. To summarize the discussion, the proposal for value_set must satisfy the following constraints:

1. it must have a ‘size’.
2. it must have a ‘name’.
3. it must support simple datatypes, such as BitType and TupleType.
4. it must be able to generate meaningful API.
5. it must support more complex datatype, such as StructType(to be able to name the fields)
6. it must support sharing between parser states.
7. it must support multiple match_kind: exact, ternary, range.

The current proposal of 
```
value_set<T> name (size);
```
satisfies all 7 requirements, but needs an annotation on the value_set instance to specify the match_kind. An option was proposed to integrate the match_kind in the struct as:
```
struct vsk_t {
    bit<16> field : ternary;
}
```
But the discussion on this proposal can happen in a different thread as part of the modularity discussion. 

We suggest the working group to accept this PR.

mbudiu-vmw:
You should then use annotations on struct fields:
```
struct vsk_t {
    @match(ternary)
    bit<16> field;
}
```

hanw:
@mbudiu-vmw can we merge this PR?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4407,6 +4409,33 @@ select (p.tcp.port) {
 The expression `16w0 &&& 16w0xFC00` describes the set of 16-bit
 values whose most significant six bits are zero.
 
+Some targets may support parser value set, see Section [#sec-value-set]. Given
+a type `T` for the type parameter of the value set, the type of the value set
+is `set<T>`. The type of the value set must match to the type of all other
+`keysetExpression` in the same `select` expression. If there is a mismatch, the
+compiler must raise an error. A parser value set accepts bit type, tuple type
```
mbudiu-vmw: `The type of the values in the set must be one of bit<>, tuple or struct`

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4810,6 +4839,30 @@ the time budget allocated for parsing, the parser should transition to
...
+##  Parser Value Sets { #sec-value-set }
+
+In some cases, the values that determine the transition from one parser state
+to another need to be determined at run time. MPLS is one example where the
+value of the MPLS label field is used to determine what headers follow the MPLS
+tag and this mapping may change dynamically at run time. To support this
+functionality, P4 supports the notion of a Parser Value Set. This is a named
+set of values with a run time API to add and remove values from the set.
+
+Value sets are declared locally within a parser. They should be declared before
```
mbudiu-vmw: ... and can be used as a label in a select expression.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4810,6 +4839,30 @@ the time budget allocated for parsing, the parser should transition to
...
+being referenced in parser `keysetExpression`.
+
+Parser Value Sets support a `size` argument to provide hints to the compiler to
```
mbudiu-vmw: Maybe you can either cross-reference the previous section or give another example here.

mbudiu-vmw:
I would appreciate if @vgurevich and @jafingerhut would take a look as well.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4412,6 +4414,33 @@ select (p.tcp.port) {
...
+`keysetExpression` in the same `select` expression. If there is a mismatch, the
+compiler must raise an error. The type of the values in the set must be one of
+bit<>, tuple, and struct.
+
+For example, to allow control plane API to specify TCP reserved ports at
+runtime, one could write:
+
+~ Begin P4Example
+struct vsk_t {
+    @match(ternary)
```
jafingerhut: Would it be a problem for the compiler, and/or the language definition, if this annotation name were 'match_kind' instead of 'match', so it was the same as 'match_kind' used elsewhere in P4_16?  This is a low priority issue, in my mind.  'match' is fine if there are any problems using 'match_kind' as an annotation name.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4412,6 +4414,33 @@ select (p.tcp.port) {
...
+    bit<16> port;
+}
+value_set<vsk_t>(4) pvs;
+select (p.tcp.port) {
+    pvs: runtime_defined_port;
+    _: other_port;
+}
+~ End P4Example
+
+The above example allows the runtime API to populate upto 4 different
```
jafingerhut: recommend "up to" instead of "upto" for English prose.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4412,6 +4414,33 @@ select (p.tcp.port) {
...
+compiler must raise an error. The type of the values in the set must be one of
+bit<>, tuple, and struct.
+
+For example, to allow control plane API to specify TCP reserved ports at
+runtime, one could write:
+
+~ Begin P4Example
+struct vsk_t {
+    @match(ternary)
+    bit<16> port;
```
jafingerhut: What if there is no @match annotation on a field?

Two possibilites I can think of:
(a) It is an error
(b) It defaults to @match(exact)

(b) seems reasonable to me, but whatever the choice, it would be good to say in the spec.

Is this syntax expected to be allowed?  `value_set<@match(ternary) bit<16>>(4) pvs;`  Or should we simply say that if you want a single non-exact field, you must put that field into a struct by itself, with the desired @match annotation?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4412,6 +4414,33 @@ select (p.tcp.port) {
...
 values whose most significant six bits are zero.
 
+Some targets may support parser value set, see Section [#sec-value-set]. Given
+a type `T` for the type parameter of the value set, the type of the value set
+is `set<T>`. The type of the value set must match to the type of all other
+`keysetExpression` in the same `select` expression. If there is a mismatch, the
+compiler must raise an error. The type of the values in the set must be one of
+bit<>, tuple, and struct.
+
+For example, to allow control plane API to specify TCP reserved ports at
```
jafingerhut: recommend adding "the" here: "to allow the control plane API"

jafingerhut:
@mbudiu-vmw I have added my review comments.

jafingerhut:
@hanw It would also be good to edit Appendix D "Restrictions on compile time and run time calls" to mention value_set's and their proposed restrictions, everywhere that a P4 table is mentioned already.  value_set's might not have identical restrictions to tables, but they are a new stateful element that is very similar to them, and people might look to that section for ideas on where they need to support it, vs. need not.

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: match_kind is a keyword used to declare new match kinds. We may be able to tweak the grammar to make it work, but it may be confusing.

mbudiu-vmw:
@hanw : I think we are close to merging this; @jafingerhut makes some good points that maybe you can address.

p4-16/spec/P4-16-spec.mdk
hanw: Use match_kind triggers a frontend error. I will use match for now.
```
syntax error, unexpected MATCH_KIND
  @match_kind
   ^^^^^^^^^^
error: 1 errors encountered, aborting compilation```

File: p4-16/spec/P4-16-spec.mdk (hunk 7 above)
Comment (hanw) [reply p4-16/spec/P4-16-spec.mdk]: I modified the text to specify default is match(exact) and single non-exact field must be in a struct by itself.

**Comment by hanw** [comment]
I modified Appendix D to specify the restrictions on value-set, please take a look at to make sure the text is accurate.

File: p4-16/spec/P4-16-spec.mdk [hunk 9]
```diff
@@ -6680,6 +6752,7 @@ e.g. `bit<32> x = rand.get();`.
 | control     |  no       | yes     | no        | no      | no     |
 | extern      | yes       | yes     | yes       | yes     | no     |
 | table       |  no       | yes     | no        | no      | no     |
+| value-set   |  no       | no      | no        | no      | no     |
```
jafingerhut: I think it might be reasonable to put a "yes" in the "parser state" column for the value-set row, if you also say above after the sentence:

"Calling a parser, control, or table means invoking its `apply()` method."

something like this:

"Calling a value-set means using it in a `select` expression."

hanw:
make sense, I fixed that column.

PR #604, opened 2018-03-27
PSA Updates related to recent P4Runtime API discussions

jafingerhut:


jafingerhut:
@samar-abdi @cc10512 @antoninbas Calling your attention to these proposed changes to the PSA spec, to make some discussion of properties of the P4Runtime API related to the '7 custom-width types' more up to date given recent discussions in the P4-API working group.  Feel free to add others for their attention if you think they should review.

I would also be happy to find a way to move all of this text into a P4Runtime API specification document, in order to simply refer to that section of the P4Runtime spec from the PSA spec, but given that there are some close ties between PSA and P4Runtime in this area, some overlap may be better than none.

p4-16/psa/psa.p4
```diff
@@ -78,7 +78,7 @@ const   CloneSessionId_t PSA_CLONE_SESSION_TO_CPU = unspecified;
  * fields may be used in a single P4 header definition, even on P4
  * implementations that restrict headers to contain fields with a
  * total length that is a multiple of 8 bits. */
-typedef bit<16> PortIdInHeader_t;
+typedef bit<32> PortIdInHeader_t;
```
jafingerhut: This change was suggested by Waqar Mohsin in a P4-API working group meeting, and several others agreed it was a good change to make.

p4-16/psa/PSA.mdk
```diff
@@ -401,7 +406,7 @@ port id values, for each of the channels of communication between the
 controller and data plane described above.
 
 Occurrences of values of type `PortId_t` or `PortIdInHeader_t` must
-include a `port_translation` annotation so the compiler can mark
+include a `p4runtime_translation` annotation so the compiler can mark
```
jafingerhut: This is basically fixing a typo that snuck past review in PSA v1.0.

p4-16/psa/PSA.mdk
```diff
@@ -424,6 +429,15 @@ translation performed may lead to functional errors in your program:
...
 + Compare port values with relational operators like `<` or `>`.
++ Compare port values to specific numeric literal values like 0 or
+  0xff.  It is recommended instead to compare their values by using
+  them as table key fields, or parser value set key fields, against
+  values installed by the control plane (which have been translated to
+  the corresponding device-specific value as determined by the
+  device's agent software).  It is also reasonable to compare port
+  values for equality against the symbolic constant values
+  `PSA_PORT_CPU` or `PSA_PORT_RECIRCULATE`, which have target-specific
+  numeric values.
```
jafingerhut: This addition isn't really a new restriction -- mainly calling out one that could have been described in PSA v1.0 but was not.  It generated some discussion in a P4-API working group, so I decided it would be best to explicitly describe it with the other recommended restrictions on use of such translated values in P4 programs.

p4-16/psa/PSA.mdk
```diff
@@ -452,6 +466,12 @@ supported value for each of these types.
 + `EgressInstance_t`
 + `Timestamp_t`
 
+~ TBD
+Values of type `Timestamp_t` are being considered for numerical
+translation in the agent software, between target-specific values, and
+a value with a common unit and 0 value reference across all targts.
+~
+
```
jafingerhut: This is something that Samar Abdi pointed out as perhaps something P4Runtime API server/agent software maybe should be doing.  The main challenge I can see is that there is a one-to-many relationship between controller timestamp values, and target-specific data plane timestamp values, because I would expect target-specific data plane timestamp values will wrap much more frequently than the larger values used in the controller.

p4-16/psa/PSA.mdk
```diff
@@ -293,12 +293,15 @@ few packets dropped for these reasons (perhaps only one).
...
+differ from one PSA implementation to another.
 
-The P4 Runtime API uses 64-bit values to hold values of these types.
-All PSA implementations must use sizes for these types no wider than
-the corresponding types with `InHeader_t` in their names shown below.
+For each one of these types, the P4 Runtime API uses values with the
+bit width of the corresponding `InHeader_t` type below to hold a value
+of that type, independent of the target.  All PSA implementations must
+use sizes for these types no wider than the corresponding types with
+`InHeader_t` in their names shown below.
```
jafingerhut: The original PSA v1.0 spec text here was written at least a month ago, when my understanding at the time from talking to Samar Abdi was the the controller/agent communication for all values of these types would be 64 bits wide.  Recent P4-API working group discussion makes it sound like values that are the same width as the corresponding `InHeader` types are perhaps preferable for this purpose.

jafingerhut:
@cc10512 @hanw @antoninbas @wmohsin @konne-google
It would be good to merge this in, if these changes meet with people's approval.  They are mostly minor corrections based upon things I learned about the P4Runtime API after March 1, 2018 when PSA v1.0 was published.

I would like to merge this in if it is good, and then on top of that make a pass of changes to the PSA spec to bring it up to date with the latest proposals for P4Runtime translation of types like PortId_t that were discussed several months back.

antoninbas:
@jafingerhut was there a plan to move to new type for `PortId_t`?

jafingerhut:
I have been trying out a few changes to PSA spec and its example programs related to using `type` for type `PortId_t` and `ClassOfService_t` over this past weekend.  I believe those two are the only ones that we definitely plane to use numeric translation between a P4Runtime client and server, although we can expand that list if desired.

I believe we want to use the new `type` definition for those 2 types, but also the 5 PSA-specific types below, even though we have no current plans to do server<->client numeric translation for those 5 types, primarily so that each PSA implementation can customize the data plane bit widths for each of these 5 types.

+ MulticastGroup_t
+ CloneSessionId_t
+ PacketLength_t
+ EgressInstance_t
+ Timestamp_t

I tried compiling a simple program using a type declared with `type` as the type of an action parameters, and the command `p4c --p4runtime-file p4rt.json --p4runtime-format json myprog.p4` gives the following error:

```
psa-portid-using-newtype2.p4(176): error: Action parameter next_port has a type which is not bit<> or int<> or bool
    action fwd(PortId_t next_port) {
                        ^^^^^^^^^
terminate called after throwing an instance of 'Util::CompilerBug'
  what():  In file: /home/jafinger/p4c/control-plane/p4RuntimeSerializer.cpp:791
Compiler Bug: psa-portid-using-newtype2.p4(85): Header field PortIdInHeader_t ingress_port has a type which is not bit<> or int<>
    PortIdInHeader_t ingress_port;
                     ^^^^^^^^^^^^
```

I believe one way to fix that would be to have code for generating the P4Info file that took some kind of architecture parameter, e.g. `psa`, and for the `psa` arch it would recognize these 7 type names specially, and emit the desired client<->server bit widths for values of those types, as well as some kind of extra data indicating that the value is of type `PortId_t`?  Were you thinking of saving the actual source code type name in the P4Info file?  It seems like a good idea to me.

mbudiu-vmw:
If you way `type bit<32> PortId_t` then you can find out the width by resolving the type. This should be as good as an annotation. This information is still available at the point the control-plane information is generated.

antoninbas:
@mbudiu-vmw  IIRC the `type` declaration gives you the data plane bitwidth, not the control plane bitwidth

mbudiu-vmw:
I see. Well, we could ask people to also add an annotation to such type declarations.

jafingerhut:
@antoninbas @mbudiu-vmw Not sure if this is a good idea, but we could have a version of the psa.p4 #include file specifically with all control plane widths for the `type` definitions in it.  That would mean having a separate compilation run with that version of psa.p4 to generate P4Info files with those widths, separate from the P4 compiler run to generate code for the specific target.

antoninbas:
I don't think it's good to require 2 compilations.
I think that the 2 solutions we have (annotation in psa.p4 on the type declaration or hard-code the bitwidth in the P4Info generation logic) are both acceptable. I do like the annotation solution better because it limits the amount of architecture dependent code in the P4Info generation code.

jafingerhut:
A new PR has been created with additional updates to the PSA spec for runtime translation of PortId_t types, here: https://github.com/p4lang/p4-spec/pull/678

It includes updating the psa.p4 include file to use P4_16 `type` instead of `typedef`.

PR #607, opened 2018-04-09
PSA add a more complex example for resubmit

hanw:


p4-16/psa/examples/psa-example-resubmit2.p4
```diff
@@ -0,0 +1,255 @@
...
+           PSA_PacketPath_t.RESUBMIT: copy_resubmit_meta;
+           PSA_PacketPath_t.NORMAL: packet_in_parsing;
+        }
+    }
+
+    state copy_resubmit_meta {
+        transition select(resub_meta.selector) {
+           8w0 : parse_resubmit_0;
+           8w1 : parse_resubmit_1;
+        }
```
jafingerhut: So if you are planning to use a header_union in an example, there shouldn't really be any need to have separate assignments for the cases where one member of the header_union is valid vs. a different member.  You can simply have a single assignment like `user_meta.h_u = resub_meta.h_u;`, where `user_meta.h_u` has the same header_union type as `resub_meta.h_u`.

I am not saying anything is functionally wrong with this code, but why write more code when something shorter will achieve the same effect?

PR #612, opened 2018-04-17
Remove PSA_SWITCH macro

hanw:


hanw:
ran 'make check', there are only warnings.

Issue #611, opened 2018-04-16
Serializable Enums


akeep:
Hey All,

The serializable working group is closing in on a resolution for the serializable enums feature request.

The current write-up of where we are is: https://github.com/akeep/p4-proposals/blob/master/enum.pdf part of the public repository: https://github.com/akeep/p4-proposals

The current potential options are described in the document, but to call out the list here, we have a few questions left to answer:

1. What should happen if a serializable enum is used in a header, but doesn't have a complete set of mappings from integers to symbols (as in the ``etherType_t`` enum example in the proposal), and we then extract from ``packet_in`` into the header.  We discussed a few options:

    1. Avoid the problem by disallowing enums in headers.
    1. Raise a parsing error when the numeric value does not have a symbolic representation in the enum
    1. Allow the operation with the possibility that the result will be undefined.

1. What should happen if a serializable enum is cast to from an integer type, and there is no mapping from the integer value to a symbolic one.  And we talked about a couple of actions.

    1. Avoid the problem by disallowing casts from ``bit<n>`` values to enums
    1. Allow the operation with the possibility that the result will be undefined.
    1. Provide a ``convert`` extern that takes a default value to be used in the case where the supplied ``bit<n>`` value is outside the symbolic list.

    (Note: we cannot necessarily raise an error here, since the cast might happen outside the parser.)

1. What happens when two symbols map to to the same underlying numeric representation?  We didn't have much discussion about this, but I would propose that we should either.

    1. Avoid the problem by disallowing more than one symbol to map to the same numeric value.
    1. Treat these two symbols as aliases for each other, effectively meaning these symbols are interchangeable.

I guess my personal bias is towards 1. iii., 2. ii., and 3. i.  My big concern with 1. i. and 2. i. is that this limits the usefulness of the enums so as to make them somewhat useless.  For instance, I then could not have an enum in the header for ``ether_t``, nor could I cast the ``bit<16>`` representing the ``etherType`` to an enum.  Others have brought up concerns about the expense of implementing 1.ii. and 2. iii..  One way to possibly mitigate this is to only allow these in cases where the symbol values are densely grouped within a numeric range, requiring only a less than or less than and greater than checks to determine if it is in range.

At the same time, there is no question that 1.iii. and 2.ii. do trade off a bit of safety, because after an extract or cast, we cannot assume the enum will have a value that matches one of the symbolic values.  Arguably, this is also true when we choose to simply represent the same information as a ``bit<n>`` type.

[...]

PR #615, opened 2018-05-02
Adding serializable enums (or enums with backing types and specified representations).

akeep:
Hey All,

Following up on the discussion in #611, this pull request attempts to capture the proposal in the P4-16 specification document.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2099,8 +2099,10 @@ enum bit<8> FailingExample {
 }
 ~ End P4Example
 
-would result in an error because the fixed-width integer value associated with
+would raise an error because `300`, the value associated with
 `FailingExample.unrepresentable` cannot be represented as a `bit<8>` value.
+
+The `initializer` expression must be a compile-time known value.
  
 Annotations, represented by the non-terminal `optAnnotations`, are
 described in Section [#sec-annotations].
@@ -2789,15 +2791,26 @@ enum bit<8> E {
 }
 ~ End P4Example
 
-Note that each symbolic value in the enum must map to a unique fixed-width integer
-representation, for example the `enum` in the following example would result in
-error:
+More than one symbolic value in an `enum` may map to the same fixed-with
+integer value.
 
 ~ Begin P4Example
-enum bit<8> Bad {
+enum bit<8> NonUnique {
   b1 = 0,
-  b2 = 1,  // Error: b2 and b3 both map to 1
-  b3 = 1,  // one must be changed.
+  b2 = 1,  // Note, both b2 and b3 map to the same value.
+  b3 = 1,  
+  b4 = 2
+}
+~ End P4Example
+
+Additionally, within an `enum` declaration, earlier symbols may be referred to in later
+symbol declarations, so `NonUnique` could also be defined as follows.
+
+~ Begin P4Example
+enum bit<8> NonUnique {
+  b1 = 0,
+  b2 = 1,  // Note, both b2 and b3 map to the same value.
+  b3 = b2,  
   b4 = 2
 }
 ~ End P4Example
@@ -2828,17 +2841,7 @@ E e = (E) x; // sets e to an unnamed value
 ~ End P4Example
 
 sets `e` to an unnamed value, since there is no symbol corresponding to the
-fixed-width integer value `5`.  While the value of `e` is unspecified, it must not
-correspond to any symbol defined within enum `E`.  This is because casting from
-a fixed-width integer type to an `enum` type is required to obey the following semantics:
-
-Given an enum type $E$ with underlying type `bit<W>` and the set of mappings
-of $(s, n)$ where $S$ is the set of symbols ($s$) and $N$ is the set of fixed-width integer
-representations ($n$), the following properties must hold:
-
-1. if an integer $x \notin N$ then $\forall e \in S$, `(E)x != e`
-1. $\forall x \in [0, 2^W]$, `(bit<W>)(E)x == x`
-1. $\forall e \in S$, `(E)(bit<W>)e == e`
+fixed-width integer value `5`.  
 
 For example, in the following code, the `else` clause of the `if/else if/else`
 block can be reached even though the matches on `x` are complete with respect
@@ -2895,7 +2898,7 @@ parser my_parser(...) {
 }
 ~ End P4Example
 
-Any `enum` that contains an unspecifed value, whether as the result of a cast
+Any `enum` that contains an unnamed value, whether as the result of a cast
 to an `enum` with an underlying type, parse into the field of an `enum` with an
 underlying type, or simply the declaration of any `enum` without a specified
 initial value might not be equal to any of the values defined for that
@@ -2917,21 +2920,6 @@ representation.
 For `enum` values with an underlying type and representations, the compiler should
 use the specified underlying type as the serialization data type and
 representation.
-
-In addition to these operations a P4 implementation may choose to implement
-externs for explicitly checking that a fixed-width integer value corresponds to
-a symbolic representation within an `enum` or provide a checked conversion
-extern.
-
-~ Begin P4Example
-extern bool is_valid<Enum, Int>(Int x);
-extern Enum convert<Enum, Int>(Int x, Enum default);
-~ End P4Example
-
-The `is_valid` extern returns `true` for integer values that correspond to
-symbols in the supplied `Enum` type parameter.  The `convert` function returns
-either the symbolic value associated with `x` in `Enum` when one exists, or the
-provided `default`.
 
 ## Expressions on Booleans { #sec-bool-exprs }
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -5,7 +5,7 @@ Author: The P4 Language Consortium
 Heading depth: 5
 Pdf Latex: xelatex
 Document Class: [11pt]article
-Package: [top=1in, bottom=1.25in, left=1in, right=1in]{geometry}
+Package: [top=1in, bottom=1.25in, left=1in, right=1in]geometry
 Package: fancyhdr
 
 Tex Header:
@@ -2058,8 +2058,8 @@ specified, so their "size" in bits is not specified (it is
 target-specific).
 
 It is also possible to specify an `enum` with an underlying representation.
-This requires the programmer provide both the numeric type and an associated
-numeric value for each symbolic entry in the enumeration.  For example, the
+This requires the programmer provide both the fixed-width integer type and an associated
+fixed-width integer value for each symbolic entry in the enumeration.  For example, the
 declaration
 
 ~ Begin P4Example
@@ -2074,10 +2074,10 @@ enum bit<16> EtherType {
 ~ End P4Example
 
 introduces a new enumeration type, which contains five constants---e.g.,
-`EtherType.IPV4`.  This `enum` declaration specifies the numeric representation
-for each entry in the `enum` and provides a backing type: `bit<16>`.
-Implementations are expected to raise an error if the numeric representation
-for an enumeration entry falls outside the representation range of the backing
+`EtherType.IPV4`.  This `enum` declaration specifies the fixed-width integer representation
+for each entry in the `enum` and provides an underlying type: `bit<16>`.
+Compiler implementations are expected to raise an error if the fixed-width integer representation
+for an enumeration entry falls outside the representation range of the underlying
 type.
 
 For example, the declaration
@@ -2091,7 +2091,7 @@ enum bit<8> FailingExample {
 }
 ~ End P4Example
 
-would result in an error because the numeric value associated with
+would result in an error because the fixed-width integer value associated with
 `FailingExample.unrepresentable` cannot be represented as a `bit<8>` value.
  
 Annotations, represented by the non-terminal `optAnnotations`, are
@@ -2334,7 +2334,7 @@ infinite-precision integer, without a width specified.
 |----------------|-----------|---------|----------|
 
 [^enum_header]: an `enum` type used as a field in a `header` must specify a
-    backing type and representation for `enum` elements.
+    underlying type and representation for `enum` elements.
 
 Rationale: `int` does not have precise storage requirements,
 unlike `bit<>` or `int<>` types. `match_kind`
@@ -2767,11 +2767,11 @@ X.v1  // reference to v1
 v1    // error - v1 is not in the top-level namespace
 ~ End P4Example
 
-Similar to errors, `enum` expressions without a specified backing type only support equality (`==`)
-and inequality (`!=`) comparisons. Expressions whose type is an `enum` without a specified backing type
+Similar to errors, `enum` expressions without a specified underlying type only support equality (`==`)
+and inequality (`!=`) comparisons. Expressions whose type is an `enum` without a specified underlying type
 cannot be cast to or from any other type.
 
-An `enum` that specifies a backing type, such as the following:
+An `enum` may also specify an underlying type, such as the following:
 
 ~ Begin P4Example
 enum bit<8> E {
@@ -2781,24 +2781,37 @@ enum bit<8> E {
 }
 ~ End P4Example
 
-also supports explicit casts to and from the backing type.  For instance, the
-following code:
+Note that each symbolic value in the enum must map to a unique fixed-width integer
+representation, for example the `enum` in the following example would result in
+error:
+
+~ Begin P4Example
+enum bit<8> Bad {
+  b1 = 0,
+  b2 = 1,  // Error: b2 and b3 both map to 1
+  b3 = 1,  // one must be changed.
+  b4 = 2
+}
+~ End P4Example
+
+An `enum` with an underlying type also supports explicit casts to and from the
+underlying type.  For instance, the following code:
 
 ~ Begin P4Example
 bit<8> x;
-E e1 = E.e2;
-E e2;
+E a = E.e2;
+E b;
 
-x  = (bit<8>) e1; // sets x to 1
-e2 = (E) x;       // sets e to E.e2
+x = (bit<8>) a; // sets x to 1
+b = (E) x;      // sets e to E.e2
 ~ End P4Example
 
-casts `e1`, which was initialized to `E.e2` to a `bit<8>`, using the specified
-numeric representation for `E.e2`, `1`.  The variable `e2` is then set to the
-symbolic value `E.e2`, which corresponds to the numeric value `1`.
+casts `a`, which was initialized to `E.e2` to a `bit<8>`, using the specified
+fixed-width integer representation for `E.e2`, `1`.  The variable `b` is then set to the
+symbolic value `E.e2`, which corresponds to the fixed-width integer value `1`.
 
-Note that while it is always safe to cast from an `enum` to its numeric type,
-it may not be th case that every numeric value specified by the backing type
+Note that while it is always safe to cast from an `enum` to its fixed-width integer type,
+it may not be the case that every fixed-width integer value specified by the underlying type
 has a related symbol in the `enum`.  For instance, the following code:
 
 ~ Begin P4Example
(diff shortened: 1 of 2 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2767,64 @@ X.v1  // reference to v1
...
+
+also supports explicit casts to and from the backing type.  For instance, the
+following code:
+
+~ Begin P4Example
+bit<8> x;
+E e1 = E.e2;
+E e2;
+
+x  = (bit<8>) e1; // sets x to 1
```
jafingerhut: Should that be "sets x to 0"?

jafingerhut: I may be wrong in my comment, but note that it is fairly confusing to have variables with the exact same names as the enum member names, e.g. e1, e2 are used as both in this example code.

akeep: No, e1 = E.e2, which is 1.  I should, however, chose different names for variables e1 and e2 (that was a really confusing choice on my part!)

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2767,64 @@ X.v1  // reference to v1
...
+correspond to any symbol defined within enum `E`.  This is because casting from
+a numeric type to an `enum` type is required obey the following semantics:
+
+Given an enum type $E$ with underlying type `bit<W>` and the set of mappings
+of $(s, n)$ where $S$ is the set of symbols ($s$) and $N$ is the set of numeric
+representations ($n$), the following properties must hold:
+
+1. if an integer $x \notin N$ then $\forall e \in S$, `(E)x != e`
+1. $\forall x \in [0, 2^W]$, `(bit<W>)(E)x == x`
+1. $\forall e \in S$, `(E)(bit<W>)e == e`
```
jafingerhut: So if I roughly translate this into English, if an enum type with a backing type `bit<W>`, where the numeric values defined by the enum definition do _not_ include all possible `2^W` values of the backing type, then it is guaranteed that variables of that enum type might not be equal to any of the named values in the enum type?

```
enum bit<2> MyPartialEnum_t {
    VALUE_A = 2w0,
    VALUE_B = 2w1,
    VALUE_C = 2w2
}

MyPartialEnum_t x;

if (x == MyPartialEnum_t.VALUE_A) {
    // some code here
} else if (x == MyPartialEnum_t.VALUE_B) {
    // some code here
} else if (x == MyPartialEnum_t.VALUE_C) {
    // some code here
} else {
    // A P4 compiler MUST ASSUME that this branch can be executed
    // some code here
}
```

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2043,10 +2053,47 @@ enum Suits { Clubs, Diamonds, Hearths, Spades }
...
+  // ...
+}
+~ End P4Example
+
+introduces a new enumeration type, which contains five constants---e.g.,
+`EtherType.IPV4`.  This `enum` declaration specifies the numeric representation
+for each entry in the `enum` and provides a backing type: `bit<16>`.
+Implementations are expected to raise an error if the numeric representation
+for an enumeration entry falls outside the representation range of the backing
+type.
```
jamescoole-cisco: It seems like the issue of multiple entries having the same representation should be discussed somewhere in here. For example, whether this is allowed:

```
enum bit<16> SomeEnum {
  Case1 = 0x0001,
  Case2 = 0x0001, // related, can I say instead "= Case1"?
  Case3 = 0x0002
}
```

jafingerhut: Good point.  Given the contents of the rest of the proposal, it seems that this should not be allowed.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2767,64 @@ X.v1  // reference to v1
...
+
+~ Begin P4Example
+bit<8> x = 5;
+E e = (E) x; // sets e to an unspecified value
+~ End P4Example
+
+sets `e` to an unknown value, since there is no symbol corresponding to the
+numeric value `5`.  While the value of `e` is unspecified, it must not
+correspond to any symbol defined within enum `E`.  This is because casting from
+a numeric type to an `enum` type is required obey the following semantics:
```
jamescoole-cisco: Some of this has been discussed before and was in the proposal, but I'm not sure where it ended up.

It seems that providing something like `bool in_range<Enum, Back>(Back b)` (or `bool in_range<Enum>(Enum e)`), either as part of the language or an extern, would be pretty valuable.

If a user wants to check that an unsafe cast will (or did) succeed, they're still going to even if these aren't provided, but the code they write will need to be updated whenever the enum definition changes.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2043,10 +2053,47 @@ enum Suits { Clubs, Diamonds, Hearths, Spades }
 introduces a new enumeration type, which contains four
 constants---e.g., `Suits.Clubs`. An `enum` declaration
 introduces a new identifier in the current scope for naming the
-created type. The underlying representation of such values is not
+created type.  The underlying representation of the `Suits` enum is not
 specified, so their "size" in bits is not specified (it is
 target-specific).
 
+It is also possible to specify an `enum` with an underlying representation.
+This requires the programmer provide both the numeric type and an associated
```
jamescoole-cisco: Should "the numeric type" be "a fixed-width integer type"?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2767,64 @@ X.v1  // reference to v1
...
+numeric value `5`.  While the value of `e` is unspecified, it must not
+correspond to any symbol defined within enum `E`.  This is because casting from
+a numeric type to an `enum` type is required obey the following semantics:
+
+Given an enum type $E$ with underlying type `bit<W>` and the set of mappings
+of $(s, n)$ where $S$ is the set of symbols ($s$) and $N$ is the set of numeric
+representations ($n$), the following properties must hold:
+
+1. if an integer $x \notin N$ then $\forall e \in S$, `(E)x != e`
+1. $\forall x \in [0, 2^W]$, `(bit<W>)(E)x == x`
```
mbudiu-vmw: I don' t know why you want this property to hold. Who cares about it?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2767,64 @@ X.v1  // reference to v1
...
+it may not be th case that every numeric value specified by the backing type
+has a related symbol in the `enum`.  For instance, the following code:
+
+~ Begin P4Example
+bit<8> x = 5;
+E e = (E) x; // sets e to an unspecified value
+~ End P4Example
+
+sets `e` to an unknown value, since there is no symbol corresponding to the
+numeric value `5`.  While the value of `e` is unspecified, it must not
```
mbudiu-vmw: I think it is simpler to keep this to be unspecified, and make no such promises.

p4-16/spec/P4-16-spec.mdk
antoninbas: Do you mean to say the property holds by definition of an enum with underlying type and doesn't need to be spelled out explicitly?

p4-16/spec/P4-16-spec.mdk
antoninbas: Strongly disagree. I believe that would greatly limit how these enums can be used. Please see the parser example I gave below in a different comment.

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: If that is the model, then we should say so explicitly in the spec.
This will preclude other implementations, and will also make it easier for people to understand what is going on. Do not leave freedom to the compiler writers if you think you will be sorry about it later. Like people designing C did with respect to integer overflow.

p4-16/spec/P4-16-spec.mdk
akeep: Should one (or both) of these examples be added as explanation to the math?

p4-16/spec/P4-16-spec.mdk
akeep: This was requested in the issue #611 related to this and I basically pulled the text from there (with a lite-rewording & turning some text in math).  Where do we stand on this?

Personally, I don't feel strongly about it one way or the other.

p4-16/spec/P4-16-spec.mdk
akeep: This property is important for programmers expectations. Without it the example that @antoninbas showed in the comments above is not something that a programmer could rely on working across P4 implementations.

p4-16/spec/P4-16-spec.mdk
jamescoole-cisco: It's not obvious that the first example is unique to enums with backing types. If uninitialized enum variables are intended to be "undefined" only within valid entries of the enum, it might be good to state that explicitly.

However, if that's the case, (compiler) implementations will already need to initialize non-backed enum variables in most cases, so why not require the same for enums with backing types?

mbudiu-vmw:
My proposal is to actually rephrase this to make it look more like a bit type with a bunch of constant values. Then the compiler does not have a choice: it will always implement such an enum with the underlying type. Also, the semantics of values that are not one of the constants is much simpler to describe. And you get all the properties that you are asking for for free.

So `enum bit<32> X { A = 1, B = 2 }` is in fact just a `bit<32>`, with a bunch of named constants X.A and X.B. Then we don't even have the requirement for A and B to be distinct, they can have any values you want. The compiler will treat them just as constants. Unlike a `bit<32>`, but much like a `newtype` (at least as it is today) this enum will have no arithmetic operations except tests for equality, assignments and casts from/to `bit<32>`. The most important gain is that now the semantics of a cast is very simple to describe: the same underlying bit representation is kept.

What do you think? So I think that the syntax is unchanged, you get all the properties you wanted to get, but it is much simpler for a programmer to understand what such an enum is doing. All the business about undefined values is gone.

antoninbas:
@mbudiu-vmw that works for me

p4-16/spec/P4-16-spec.mdk
jafingerhut: @jamescoole-cisco I actually wrote a change to the P4_16 language spec, committed after v1.0.0 was published, that says something about exactly this case, but had forgotten I wrote that part.  See https://github.com/p4lang/p4-spec/issues/611#issuecomment-386702605 for details.  Unless that text is changed further in the spec, P4_16 compilers are already free to make uninitialized enum and error values unequal to any of their named members.

jafingerhut: Given my previous comment, I am not against having example code like these in the spec, but if so, they should be clear that the examples apply for enum's with _or without_ underlying types.

I think having no code examples would also be fine, but adding a sentence that mentions that _all_ values with type enum can be not-equal to any of their named members, with a cross-reference to the section "## Operations on headers { #sec-ops-on-hdrs }" (Madoko syntax copied from the .mdk file)

jamescoole-cisco: I agree it would be good to copy or move the discussion of how undefined both types of enums can be into section, since it applies in situations besides header fields. But IDK if the example is necessary to get the point across.

akeep:
@antoninbas, @jamescoole-cisco, @jafingerhut I've added the examples with slight modification and moved the commentary around undefined values up into the operations section.

@mbudiu-vmw I've tried to add some sentences to the `enum` definition section to talk about it more like a `newtype` based on the `bit<W>` underlying type.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2832,6 +2840,77 @@ representations ($n$), the following properties must hold:
...
+      EtherType.IPV4 : parse_ipv4;
+      EtherType.IPV6: parse_ipv6;
+      default: reject;
+  }
+}
+~ End P4Example
+
+Any `enum` that contains an unspecifed value, whether as the result of a cast
+to an `enum` with an underlying type, parse into the field of an `enum` with an
+underlying type, or simply the declaration of any `enum` without a specified
```
mbudiu-vmw: I think this explanation can now be also simplified by reducing it to the case of bitstrings.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2076,6 +2076,14 @@ enum bit<16> EtherType {
 introduces a new enumeration type, which contains five constants---e.g.,
 `EtherType.IPV4`.  This `enum` declaration specifies the fixed-width integer representation
 for each entry in the `enum` and provides an underlying type: `bit<16>`.
+This type of `enum` declaration can be thought of as declaring a new `bit<16>`
```
mbudiu-vmw: Do we allow signed enum values as well?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2775,163 @@ X.v1  // reference to v1
...
+An `enum` with an underlying type also supports explicit casts to and from the
+underlying type.  For instance, the following code:
+
+~ Begin P4Example
+bit<8> x;
+E a = E.e2;
+E b;
+
+x = (bit<8>) a; // sets x to 1
+b = (E) x;      // sets e to E.e2
```
jafingerhut: typo: "sets e" -> "sets b"

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2775,163 @@ X.v1  // reference to v1
...
+
+~ Begin P4Example
+bit<8> x = 5;
+E e = (E) x; // sets e to an unspecified value
+~ End P4Example
+
+sets `e` to an unknown value, since there is no symbol corresponding to the
+fixed-width integer value `5`.  While the value of `e` is unspecified, it must not
+correspond to any symbol defined within enum `E`.  This is because casting from
+a fixed-width integer type to an `enum` type is required to obey the following semantics:
```
jafingerhut: I know you have the later condition that (bit<W>)(E)x == x for all possible `2^W` values of x of type `bit<W>`.  Assuming we keep that condition (seems reasonable to me), it isn't really true that the assignment to `e` in the example code above assigns `e` an "unknown" value?  I mean, it must be a value not equal to any named member of type `E`, I agree, but it must also be able to be cast back to a `bit<8>` type value of 5, and no other, so we know a lot more about it than that it is simply unknown.

I mention this simply because uninitialized values in P4_16 are, I believe, "more unknown" than the value of `e` in this sample code.

p4-16/spec/P4-16-spec.mdk
akeep: I'm not sure what you mean by this.

p4-16/spec/P4-16-spec.mdk
akeep: I've used unsigned in all of the examples, but I don't think there is any particular reason to restrict it to this.  Do you have an opinion?  (I'm happy to add a case for it.)

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2775,163 @@ X.v1  // reference to v1
...
+
+~ Begin P4Example
+enum bit<8> E {
+  e1 = 0,
+  e2 = 1,
+  e3 = 2
+}
+~ End P4Example
+
+Note that each symbolic value in the enum must map to a unique fixed-width integer
```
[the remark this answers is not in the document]
jamescoole-cisco: Can case names be used while defining the enum?

This is particularly useful if aliases are allowed:

``` p4
enum bit<32> Authors {
  Clemens = 0,
  Twain = Clemens // is Clemens defined here? is Authors.Clemens?
}
```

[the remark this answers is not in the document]
mbudiu-vmw: Not clear, a declaration is available only after it has been "completed". Not clear whether the Clemens declaration ends with the comma or with the closing brace.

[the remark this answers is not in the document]
jamescoole-cisco: I'd argue that since it's useful, it should be allowed. But there should be some language to say one way or another.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2775,163 @@ X.v1  // reference to v1
...
+  a table entry where the field has a prefix length of 0.
+
+Note that if an `enum` value lacking an underlying type appears in the control-plane API, the
 compiler must select a suitable serialization data type and
 representation.
+For `enum` values with an underlying type and representations, the compiler should
+use the specified underlying type as the serialization data type and
+representation.
+
+In addition to these operations a P4 implementation may choose to implement
```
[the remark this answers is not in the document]
akeep: I went ahead and removed these, I agree having them as part of an architecture description is probably right.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2775,163 @@ X.v1  // reference to v1
...
+bit<8> x = 5;
+E e = (E) x; // sets e to an unnamed value
+~ End P4Example
+
+sets `e` to an unnamed value, since there is no symbol corresponding to the
+fixed-width integer value `5`.  While the value of `e` is unspecified, it must not
+correspond to any symbol defined within enum `E`.  This is because casting from
+a fixed-width integer type to an `enum` type is required to obey the following semantics:
+
+Given an enum type $E$ with underlying type `bit<W>` and the set of mappings
```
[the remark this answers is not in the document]
akeep: I removed all of the language around this.

p4-16/spec/P4-16-spec.mdk
akeep: Ah, good catch, I missed the unspecified in this one.  That said, the `While the value ...` has now been removed.

akeep:
Hey All,

I think I've addressed everything with the exception of whether an enum value should be matchable in a select as the underlying representation.  As mentioned in a comment above, I'm not a fan of that, but I've not added any text to address it yet.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2777,149 @@ X.v1  // reference to v1
...
+~ Begin P4Example
+enum bit<8> NonUnique {
+  b1 = 0,
+  b2 = 1,  // Note, both b2 and b3 map to the same value.
+  b3 = 1,  
+  b4 = 2
+}
+~ End P4Example
+
+Additionally, within an `enum` declaration, earlier symbols may be referred to in later
```
mbudiu-vmw: Can we leave this out of the spec for now?
If anyone asks for this feature we should clarify.
But it will be easier to implement without it.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2777,149 @@ X.v1  // reference to v1
...
+
+x = (bit<8>) a; // sets x to 1
+b = (E) x;      // sets b to E.e2
+~ End P4Example
+
+casts `a`, which was initialized to `E.e2` to a `bit<8>`, using the specified
+fixed-width integer representation for `E.e2`, `1`.  The variable `b` is then set to the
+symbolic value `E.e2`, which corresponds to the fixed-width integer value `1`.
+
+Note that while it is always safe to cast from an `enum` to its fixed-width integer type,
```
mbudiu-vmw: Maybe "safe" is not the right word here; with our current definition both casts are actually "safe", but the second one does not create a "named" value.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2777,149 @@ X.v1  // reference to v1
...
+    transition select(hdr.ethernet.etherType) {
+      EtherType.VLAN : parse_vlan;
+      EtherType.IPV4 : parse_ipv4;
+      EtherType.IPV6: parse_ipv6;
+      default: reject;
+  }
+}
+~ End P4Example
+
+Any `enum` that contains an unnamed value, whether as the result of a cast
```
mbudiu-vmw: We should say here "variable with an enum type". The "enum" itself is the type declaration.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2777,149 @@ X.v1  // reference to v1
...
+to an `enum` with an underlying type, parse into the field of an `enum` with an
+underlying type, or simply the declaration of any `enum` without a specified
+initial value might not be equal to any of the values defined for that
```
mbudiu-vmw: Actually I think that "might" is too weak: it "will not" be equal for sure.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2717,13 +2777,149 @@ X.v1  // reference to v1
...
+type.  Such an unspecified value should still lead to predictable
```
mbudiu-vmw: unspecified -> unnamed

p4-16/spec/grammar.mdk
```diff
@@ -334,6 +334,7 @@ structField
 
 enumDeclaration
     : optAnnotations ENUM name '{' identifierList '}'
+    | optAnnotations ENUM typeRef name '{' specifiedIdentifierList '}'
```
mbudiu-vmw: I have implemented this grammar change in the compiler and it will work out (using "bit" in fact) - there are no ambiguities.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2026,12 +2026,22 @@ An enumeration type is defined using the following syntax:
 ~ Begin P4Grammar
 enumDeclaration
     : optAnnotations ENUM name '{' identifierList '}'
+    | optAnnotations ENUM typeRef name '{' specifiedIdentifierList '}'
```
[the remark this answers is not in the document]
akeep: You noted this here, but then noted later that the `typeRef` here worked, do you have a preference?

p4-16/spec/P4-16-spec.mdk
akeep: Done.

p4-16/spec/P4-16-spec.mdk
akeep: Done.

p4-16/spec/P4-16-spec.mdk
akeep: Done.

p4-16/spec/P4-16-spec.mdk
akeep: I just changed "might" to "will", I think that is accurate.

p4-16/spec/P4-16-spec.mdk
akeep: Done.

p4-16/spec/grammar.mdk
akeep: I've left this alone for now, if you prefer bit, let me know.

akeep:
Added `unsigned` to the description of the type you can use as the base type for a serializable `enum` type and updated the grammar to reflect this restriction.

akeep:
Hey All,

I'm not planning on making any additional changes, since I think I've addressed @mbudiu-vmw  requests and the requests that came up during the call on Monday.  If there are needed changes please let me know soon, I'll be traveling over the weekend and my opportunities to handle changes are going to be limited.

PR #624, opened 2018-05-17
Added specification for function declarations

mihaibudiu:
It turns out this is quite easy to write in the spec. The implementation will take a bit of work, but not too much.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4145,9 +4145,9 @@ bool b2 = x == y;  // no cast needed
 
 # Function declarations { #sec-functions }
 
-Functions can be declared only at the top-level.  P4 functions are
-modeled after functions as found in most other programming languages,
-however, P4 does not permit recursive functions.
+Functions can only be declared at the top-level and all parameters must have a direction. 
+P4 functions are modeled after functions as found in most other programming languages,
+however, the language does not permit recursive functions.
 
 ~ Begin P4Grammar
 functionDeclaration


p4-16/spec/P4-16-spec.mdk
```diff
@@ -3819,6 +3820,33 @@ table tbl {
...
+   if (left > right)
+      return left;
+   return right;
+}
+~ End P4Example
+
+A function returns a value using the `return` statement.  A function
+that returns `void` can simply use the `return` statement with no
+arguments.  A function with a non-void return type must return a value
+of the suitable type on all possible execution paths.
```
cc10512: Should we also discuss that it is only visible to the data plane, and therefore all arguments must be directional?

mbudiu-vmw: Actually non-directional arguments would work: they have to be bound to compile-time constants. We may want to prohibit them, because you can use them to pass externs or controls as arguments.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3819,6 +3820,33 @@ table tbl {
...
+~ Begin P4Grammar
+functionDeclaration
+    : functionPrototype blockStatement
+    ;
+~ End P4Grammar
+
+Here is an example of a function that returns the maximum of two 32-bit values:
+
+~ Begin P4Example
+bit<32> max(in bit<32> left, in bit<32> right) {
```
hanw: Is overloading allowed for function? Similarly, is optional argument allowed?

mbudiu-vmw: yes, but that is part of the other spec PR, with the named arguments.
That will clarify the overloading rules.
And currently there are no optional arguments at all in the spec.

p4-16/spec/P4-16-spec.mdk
jnfoster: In the interest of converging, I'd recommend that we stick with the conservative choice for now...

jnfoster:
@mbudiu-vmw Approving this pending the suggested change above (data-plane values only).

PR #626, opened 2018-05-22
Specification for named arguments

mbudiu-vmw:
Unfortunately I have realized that allowing default values for controls in a package will require fresh type variables - since in this case the type of the data coming out of a parser is no longer the same as the type of data going through the control. This will be fixable if we reintroduce generic controls later.

```
struct Empty {}
control nothing(inout Empty h, inout Empty m) {
   apply {}
}

parser parserProto<H, M>(packet_in p, out H h, inout M m);
control controlProto<H, M>(inout H h, inout M m);

package pack<HP, MP, HC, MC>(@optional parserProto<HP, MP> _parser,  // optional parameter
                             controlProto<HC, MC> _control = nothing()); // default parameter value

pack() main;   // No value for _parser, _control is an instance of nothing()
```

Ideally this would be written as follows:

```
parser parserProto<H, M>(packet_in p, out H h, inout M m);
control controlProto<H, M>(inout H h, inout M m);

control nothing<H, M>(inout H h, inout M m) {
   apply {}
}

package pack<H, M>(@optional parserProto<H, M> _parser,  // optional parameter
                                  controlProto<H, M> _control = nothing<H, M>()); // default parameter value

pack() main;   // No value for _parser, _control is an instance of nothing()
```

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1648,7 +1648,7 @@ directions:
 ### Optional parameters
 
 A parameter that is annotated with the `@optional` annotation is
-optional: the user may supply no value for that parameter in an
+optional: the user may omit the value for that parameter in an
 invocation.  Optional parameters can only appear for arguments of:
 packages, extern functions, extern methods, and extern object
 constructors.  Optional parameters cannot have default values.  If a
@@ -1660,19 +1660,20 @@ list.
 The implementation of such objects is not expressed in P4, so the
 meaning and implementation of optional parameters should be specified
 by the target architecture.  For example, we can imagine a two-stage
-switch architecture where the second stage is optional.  This could be
+switch architecture where the second stage is optional. This could be
 declared as a package with an optional parameter:
 
 ~Begin P4Example
 package pipeline(...);
 package switch(pipeline first, @optional pipeline second);
-
 ...
 pipeline(...) ingress;
 switch(ingress) main;   // a switch with a single-stage pipeline
 ~End P4Example
 
-The following example shows optional parameters and parameters with
+Here the target architecture could implement the elided optional argument using an empty pipeline.
+
+The following example shows optional parameters and parameters with 
 default values.
 
 ~Begin P4Example
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1688,8 +1688,8 @@ control nothing(inout Empty h, inout Empty m) {
    apply {}
 }
 
-parser parserProto<H, M>(packet_in p, out Header h, inout Meta m);
-control controlProto<H, M>(inout Header h, inout Meta m);
+parser parserProto<H, M>(packet_in p, out H h, inout M m);
+control controlProto<H, M>(inout H h, inout M m);
 
 package pack<HP, MP, HC, MC>(@optional parserProto<HP, MP> _parser,  // optional parameter
                              controlProto<HC, MC> _control = nothing()); // default parameter value
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1518,17 +1518,19 @@ the same result):
 
 1.	Arguments are evaluated from left to right as they appear in the
       function call expression.
-2.	For each `out` and `inout` argument the corresponding
+2.      If a parameter has a default value and no corresponding argument is
+      supplied, the default value is used as an argument.
+3.	For each `out` and `inout` argument the corresponding
       l-value is saved (so it cannot be changed by the evaluation of
       the following arguments). This is important if the argument
       contains indexing operations into a header stack.
-3.	The value of each argument is saved into a temporary.
-4.	The function is invoked with the temporaries as arguments. We are
+4.	The value of each argument is saved into a temporary.
+5.	The function is invoked with the temporaries as arguments. We are
       guaranteed that the temporaries that are passed as arguments are
       never aliased to each other, so this "generated" function call
       can be implemented using call-by-reference if supported by the
       architecture.
-5.	On function return, the temporaries that correspond to `out`
+6.	On function return, the temporaries that correspond to `out`
       or `inout` arguments are copied in order from left to right
       into the l-values saved in step 2.
 
@@ -1608,6 +1610,7 @@ nonEmptyParameterList
 
 parameter
     : optAnnotations direction typeRef name
+    | optAnnotations direction typeRef name '=' expression
     ;
 
 direction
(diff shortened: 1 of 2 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -3770,12 +3830,32 @@ typeArgumentList
...
 
-Function arguments are evaluated in order, left to right, before the
-function invocation takes place. The calling convention is
-copy-in/copy-out (Section [#sec-calling-convention]). For generic functions the type
-arguments can be explicitly specified in the function call. The compiler does not
-insert implicit casts for the arguments to methods or functions---the argument types
-must match the parameter types exactly.
+A function call or method invocation can optionally specify for each
+argument the corresponding parameter name.  It is illegal to use names
+only for some arguments: either all or no arguments should specifie
```
hanw: typo: specify

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1639,6 +1642,60 @@ directions:
...
+  these values must evaluate to compile-time constants.
+
+### Optional parameters
+
+A parameter that is annotated with the `@optional` annotation is
+optional: the user may supply no value for that parameter in an
+invocation.  Optional parameters can only appear for arguments of:
+packages, extern functions, extern methods, and extern object
+constructors.  Optional parameters cannot have default values.  If a
+function type has both optional parameters and default values then it
```
hanw: It is unclear to me if the term `function type` here is referring to all of the four types mentioned in the last few sentences, or it is referring to the extern `function`, or it is referring to some other `function`?

p4-16/spec/grammar.mdk
```diff
@@ -57,6 +57,7 @@ nonEmptyParameterList
 
 parameter
     : optAnnotations direction typeRef name
+    | optAnnotations direction typeRef name '=' expression
```
hanw: The grammar seems to imply `@optional in bit<16> x = 0` is a valid syntax. But it is actually not according to the current spec.

mbudiu-vmw: Watch out: optAnnotation does not mean "the annotation `@optional`", but it means "with some optional annotations", which could be e.g., `@hidden`.

hanw: Yes, I agree, but if `optional` is a keyword, then we won't have this ambiguity.

mbudiu-vmw: At this point optional is not a keyword. We could attempt to promote it into a keyword. Maybe it is a good idea.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1639,6 +1642,60 @@ directions:
...
+
+The implementation of such objects is not expressed in P4, so the
+meaning and implementation of optional parameters should be specified
+by the target architecture.  For example, we can imagine a two-stage
+switch architecture where the second stage is optional.  This could be
+declared as a package with an optional parameter:
+
+~Begin P4Example
+package pipeline(...);
+package switch(pipeline first, @optional pipeline second);
```
hanw: In this example, if both pipelines are optional: 
`package switch(@optional pipeline first, @optional pipeline second);` 

and if user writes:
`pipeline (...) egress;`
`switch(egress) main;`

There is an ambiguity on which pipeline argument is defined by the program. The spec did not say anything about using the named argument to disambiguate.

mbudiu-vmw: Overload resolution will not be very easy to implement. 
Also support for default parameter values is not yet implemented.

hanw: If overload resolution is implemented, do we still need the `optional` annotation? If I understand correctly, the original decision to have `@optional` was to close the gap while overload resolution is being implemented. But now we are promoting optional annotation to the language. We will end up with two mechanisms to implement the variable arguments feature.

mbudiu-vmw: Overloads are useful even if we don't have optional arguments.
Also, n optional arguments can express 2^n overloads, so they are still useful.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3954,12 +4014,32 @@ typeArgumentList
...
+they appear, left to right, before the function invocation takes
+place.
+
+~ Begin P4Example
+extern void f(in bit<32> x, out bit<16> y);
+bit<32> xa = 0;
+bit<16> ya;
+f(xa, ya);  // match arguments by position
+f(x = xa, y = ya);  // match arguments by name
+f(y = ya, x = xa);  // match arguments by name in any order
```
hanw: Is `ya` evaluated before `xa` in this case? The OrderArgument pass reorders the argument to the original order of x, y. As a result, backend only sees x, then y. Will the evaluation order be `xa`, then `ya` then?

mbudiu-vmw: The  evaluation order is handled by the frontend. It only matters if evaluation has side-effects. The text indicates that evaluation is done in argument order and not parameter order. At the backend stage the argument expressions will never have side effects.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1639,6 +1642,60 @@ directions:
   including values for the directionless parameters. The directionless
   parameters in this case behave like `in` parameters.
   See Section [#sec-invoke-actions] for further details.
+- Default parameter values are only allowed for 'in' or direction-less parameters;
+  these values must evaluate to compile-time constants.
+
+### Optional parameters
+
+A parameter that is annotated with the `@optional` annotation is
+optional: the user may supply no value for that parameter in an
```
jnfoster: "May supply no value for the parameter" --> "May omit the value for that parameter"?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1639,6 +1642,60 @@ directions:
...
+constructors.  Optional parameters cannot have default values.  If a
+procedure-like construct has both optional parameters and default values then it
+can only be called using named arguments.  It is recommended, but not
+mandatory, for all optional parameters to be at the end of a parameter
+list.
+
+The implementation of such objects is not expressed in P4, so the
+meaning and implementation of optional parameters should be specified
+by the target architecture.  For example, we can imagine a two-stage
+switch architecture where the second stage is optional.  This could be
```
jnfoster: To finish the thought experiment, can we also explain precisely how this hypothetical architecture defines the meaning of the optional parameter?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2493,9 +2550,11 @@ control d(packet_out b, in Hdr h) {
 
 Functions and methods are the only P4 constructs that support
 overloading: there can exist multiple methods with the same name in
-the same scope. Even so, two functions (or methods of an `extern`
-object) can have the same name only if they have a different number of
-parameters.
+the same scope.  Whe overloading is used, the compiler must be able to
+disambiguate at compile-time which method or function is being called,
+either by the number of arguments or by the names of the arguments,
+when calls are specifying argument names.  Argument type information
```
jnfoster: I'm slightly uneasy leaving this vague and implementation dependent.

jnfoster: But let's not hold it up for this...

jnfoster:
@mbudiu-vmw Approved modulo small fixes suggested above.

p4-16/spec/grammar.mdk
jnfoster: Since it is an annotation, it seems safe to consider it not a keyword for now.

PR #632, opened 2018-05-31
Correct syntax error in the declaration of `enum PSA_MeterColor_t`

SantiagoBautista:
This PR contains 3 minor corrections, and follows @jafingerhut 's advise on PR#1311 of p4lang/p4c.
+ There was an unexpected semicolon at the end of declaration of `enum PSA_MeterColor_t` , removed it.
+ Wrapped psa.p4 file in `#ifndef __PSA_P4__` etc...
+ Added an `#include<core.p4>` directive, since `packet_in` type is used.


jnfoster:
These look good to me, but should we just be copying `psa.p4` from the p4lang/p4-spec repo instead?

Issue #633, opened 2018-05-31
Let annotations take an argumentList rather than `expressionList | keyValueList `


SantiagoBautista:
Currently in the P4-16 grammar, annotations can take either an expression list or a keyValuePair List for their arguments and this allows arguments to annotations to be named:
```
annotation
    : '@' name
    | '@' name '(' expressionList ')'
    | '@' name '(' keyValueList ')'
    ;
```
But this makes the grammar ambiguous, because there is no way of telling whether the empty list is an expressionList or a keyValueList.

There is a non-ambiguous and nicer way of achieving exactly the same behavior:
there is a non-terminal in P4-16 grammar called argumentList , and each argument can either be an expression or a `name '=' expression` (i.e. a keyValuePair).

So if annotations took an argumentList, instead of either an expressionList or a keyValueList, we would have exactly the same behavior, but without introducing ambiguity to the grammar.

I will open a Pull Request implementing this solution.

jnfoster:
I guess the one difference is "_".

But generally, yes, this seems like a nice simplification.

On Thu, May 31, 2018 at 4:40 PM, Santiago Bautista <notifications@github.com

SantiagoBautista:
To be precise, there is another difference: if annotations take argumentList, then one single annotation can take both key-value pairs and expressions; whereas with the current grammar one annotation can either only take expressions or only take key-value pairs.

But this extra flexibility can, I guess, be useful, and AFAIK it doesn't introduce any problems.

SantiagoBautista:
For information:
This issue is related to #605 and #625 .

chrispsommers:
@SantiagoBautista I like this, good catch!

SantiagoBautista:
Thank you @chrispsommers !

mbudiu-vmw:
This is no longer relevant, since we now have free-form annotations.

PR #634, opened 2018-05-31
Application of suggestion #633

SantiagoBautista:
This Pull request offers a solution for issue #633 .

If annotations take argumentList instead
of either expressionList or keyValueList,
the exact same behavior is acheivable,
but without introducing ambiguity into the grammar.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6737,18 +6737,23 @@ annotations
 
 annotation
     : '@' name
-    | '@' name '(' expressionList ')'
-    | '@' name '(' keyValueList ')'
+    | '@' name '(' argumentList ')'
     ;
 ...
-keyValuePair
-    : IDENTIFIER '=' expression
+argumentList
+    : /* empty */
+    | nonEmptyArgList
     ;
 
-keyValueList
-    : /* empty */
-    | keyValuePair
-    | keyValueList ',' keyValuePair
+nonEmptyArgList
+    : argument
+    | nonEmptyArgList ',' argument
+    ;
+
+argument
+    : expression          /* positional argument to the annotation */
+    | name '=' expression /* key-value pair for the annotation */
+    | DONTCARE
     ;
 ~ End P4Grammar
 
--- a/p4-16/spec/grammar.mdk
+++ b/p4-16/spec/grammar.mdk
@@ -546,16 +546,6 @@ expressionList
     | expressionList ',' expression
     ;
 
-keyValuePair
-    : IDENTIFIER '=' expression
-    ;
-
-keyValueList
-    : /* empty */
-    | keyValuePair
-    | keyValueList ',' keyValuePair
-    ;
-
 member
     : name
     ;
--- a/p4-16/spec/grammar.mdk
+++ b/p4-16/spec/grammar.mdk
@@ -43,8 +43,7 @@ annotations
 
 annotation
     : '@' name
-    | '@' name '(' expressionList ')'
-    | '@' name '(' keyValueList ')'
+    | '@' name '(' argumentList ')'
     ;
 
 parameterList


ChrisDodd:
The grammar in the compiler does not allow an empty keyValueList, so there's no ambiguity.  Why the grammar in the spec was different, I don't know.

mbudiu-vmw:
Annotations never had a clear semantics in the spec, but they have an internal compiler api to access them, which will need to change if we take this pr.

mbudiu-vmw:
Although this does simplify the grammar, it changes the set of programs accepted (as you point out) and possibly it may require a different representation for the Annotation IR class in the compiler. I believe that this fix makes things more complicated. The right fix is to not accept an empty key-value list.

SantiagoBautista:
Accepting an empty key-value list (as it is currently the case in the p4-spec, but not in the p4 compiler) would introduce ambiguity.
If we want the spec to reflect what is currently done in the compiler, the fastest/less-disturbing solution is _not_ to accept empty key-value lists in the grammar.

SantiagoBautista:
That being said, the compiler and the spec already differ in many ways:
+ In the compiler there is a distinction between realTypeArgumentList and typeArgumentList that is not present in the grammar
+ In the compiler abstract methods and abstract method instantiations are supported for externs, and it is not the case in the grammar
etc

If this pr simplifies the grammar, is it really a problem if it temporarily adds another difference between the compiler and the grammar?

hanw:
The abstract method is an experimental feature and has not been accepted to the spec yet.

mbudiu-vmw:
1. The bison grammar should have clear comments on all features that are still experimental. If not, that's a bug.

2. I personally prefer to have first an implementation and then a spec change - because this way we are sure that the proposal works. Sometimes when we do it the other way round we discover that we proposed things that are very hard or impossible to implement, and then we are in a really bad position. 

3. We try to keep the bison grammar and the spec in sync, but it's not easy, since they cannot be generated automatically, especially the fragments in the spec. If you find a difference you should file a bug.

4. Value sets will appear in the next version of the spec - the draft in the repo should have that text already.

5. The problem is that the change you proposed changes the set of programs that are accepted. Your grammar will accept annotations that have both key-value pairs and lists of expressions. Today the implementation accepts one of the two, but not both. Implementing this will require some work - the classes we have today in the IR are not sufficient if you want to preserve the order of annotations on an object. 

This also changes the (not explicitly specified) contract between the front-end and the back-end: today we assume that a back-end can check for annotations in one of the two formats, but with your change back-ends have to be ready to accept annotations that have a mix of formats. 

6. In general, we should try to optimize for the language users, not for the language implementers. I will prefer a more complicated grammar if it makes the language easier to use, describe and understand.

SantiagoBautista:
I agree with @mbudiu-vmw that not accepting the empty keyValueList is a solution that ambiguity that does not require changing the compiler in any way (because it actually correspond to putting the compiler's current behavior into the grammar).

Since the proposition in this PR seems to need further discussion, I will do a PR implementing @mbudiu-vmw proposition, that shouldn't raise any problem, and we will see for this PR later. What do you think?

mbudiu-vmw:
The problem is that we will release a new specification this week, so whatever choice we make will become official. If this needs further discussion we won't be able to release this part of the spec. 
Perhaps there is an escape, since your grammar accept a strict superset of the programs accepted by the simpler proposal, we can make such a change later.

SantiagoBautista:
Ok. So I suggest that for the spec release we implement your proposition: requiring in the grammar that the keyValueList s are non-empty.
And I do think that doing so (changing the grammar, in one way or another) is necessary, because the current grammar is ambiguous about that point (and does not currently match the implementation, that does require the keyValueList to be non-empty).

SantiagoBautista:
I will also make sure to open a pull request that updates the grammar concerning value_set in parsers, because there are currently some rules missing in the grammar (and present in the implementation, so it is a bug in the grammar).

mbudiu-vmw:
I appreciate.
If they are not marked experimental it's a bug, either in the grammar or in the spec.

jnfoster:
Closing this for now...

Issue #635, opened 2018-05-31
Let optional type parameters be names rather than nonTypeNames


SantiagoBautista:
The current P4-16 grammar requires all optional type parameters to be non-type names.
But one could want to shadow locally a name that is already a type name with a type parameter.
Examples of files doing so can be found at:
[nested-tuple.p4](https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/nested-tuple.p4)
and
[type-shadow.p4](https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/type-shadow.p4) .
I suggest allowing optional type parameters to be any name.

In fact, it already is the case that the P4-16 [compiler](https://github.com/p4lang/p4c/blob/master/frontends/parsers/p4/p4parser.ypp) allows optional type parameters to be any name and not just non-type names.

I will open a PR that modifies the grammar accordingly to this suggestion.

SantiagoBautista:
Pull request #636 introduces this suggestion into the grammar.
There is nothing to change in the compiler, as it already works this way.

SantiagoBautista:
Associated Pull Request #636 was merged.

PR #636, opened 2018-05-31
Allow optional type parameters to be any name, and not just nonTypeName

SantiagoBautista:
This introduces into the grammar the suggestion #635 .
In the compiler, it is already the case that suggestion #635 is implemented.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2528,8 +2528,8 @@ optTypeParameters
     ;
 
 typeParameterList
-    : nonTypeName
-    | typeParameterList ',' nonTypeName
+    : name
+    | typeParameterList ',' name
     ;
 ~ End P4Grammar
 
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
Before you submit this you should make aure it works in the implementation.

SantiagoBautista:
The implementation already works this way.
In addition, I am implementing a P4 parser in Ocaml, and it works with this modification.

SantiagoBautista:
If you look at the implementation of the compiler, it already uses name instead of nonTypeName for optional type parameters.
```
optTypeParameters
    : /* empty */               { $$ = new IR::TypeParameters(); }
    | "<" typeParameterList ">" { $$ = new IR::TypeParameters(@1+@3, *$2); }
    ;

typeParameterList
    : name                           { $$ = new IR::IndexedVector<IR::Type_Var>();
                                       $$->push_back(new IR::Type_Var(@1, *$1)); }
    | typeParameterList "," name     { ($$=$1)->push_back(new IR::Type_Var(@3, *$3)); }
;
```

PR #638, opened 2018-06-01
Fix: add value sets to grammar

SantiagoBautista:
+ Added into the grammar the rule for valueSetDeclaration (was missing).
+ Added to the spec text a quote of the valueSetDeclaration rule (optional).


SantiagoBautista:
As promised, @mbudiu-vmw

SantiagoBautista:
You are right @mbudiu-vmw . Changed to single quotes in #639 . Thanks for noticing.

PR #640, opened 2018-06-01
Small clarifications

mbudiu-vmw:


mbudiu-vmw:
@jnfoster : I think we should merge this in the final 1.1 spec.

PR #643, opened 2018-06-12
Corrected token involved in newtype declarations in the spec description

SantiagoBautista:
This is a bugfix.
In the spec, in the description of newtype declarations, the grammar extract was wrong, as it did not use the correct token ( `TYPE_IDENTIFIER` was used instead of `TYPE` ).

The corresponding section in the grammar has not the same bug.


SantiagoBautista:
@mbudiu-vmw , @jnfoster 
As this fixes a bug in the spec text, I think it should be part of version 1.1 of the spec.

PR #648, opened 2018-06-12
Changed the place where TYPE token can be a name, to avoid conflicts

SantiagoBautista:
The fix proposed for issue#1336 of p4c

@mbudiu-vmw


mbudiu-vmw:
Since this is a bug fix I will merge this too.

PR #649, opened 2018-06-13
Use dotPrefix non terminal for toplevel paths in expressions

SantiagoBautista:
This was already the case for all other toplevel paths, so this modification adds consistency.

As dotPrefix is, in the grammar, an alias of '.' , this doesn't change at all the set of syntactically correct programs.


SantiagoBautista:
@mbudiu-vmw This reflects yesterday's change of the parser into the grammar formulation.
(It doesn't changes at all the set of accepted programs; but it increases consistency in the way the grammar is formulated).

PR #655, opened 2018-06-27
Added @deprecated annotation

mbudiu-vmw:


liujed:
Do you think it's worth mentioning `@deprecated` in §18.2 as a predefined annotation?

mbudiu-vmw:
The last paragraph could move to that section. I will do it.

PR #656, opened 2018-06-28
Draft proposal for structure overlays

mihaibudiu:
This is an initial draft for a proposal for adding structure overlays to P4.

The design is more general than the other proposals we have heard. The main difference is that an overlay is a separate function-like object that can is applied to a structure instance to produce a different structure instance. This enables people to create overlays for structs that are defined in library files, without requiring them to modify the libraries. 

Please comment on this proposal.

I believe that overlays can be implemented easily just in the compiler front-end; the compiler will just rewrite all field accesses to overlays as field accesses to the underlying structure. So they are just some nice syntactic sugar.

They have a very clear semantics, and are a much cleaner way to implement much of the functionality that is provided by C-style unions. We have debated in the past whether to have C-style unions, but I believe they won't be necessary at all with this proposal.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2553,9 +2553,8 @@ overlay IPv4 expand(IPv4_aligned s) {
 }
 ~ End P4Example
 
-An overlay must define all fields of the destination structure.  The
-"function" computing an overlay can only be used as an initializer in
-the declaration of a variable of the suitable type.  In an overlay
+The "function" computing an overlay can only be used as an initializer
+in the declaration of a variable of the suitable type.  In an overlay
 each bit from a source field must be used at most once.  This
 restriction is necessary to compute how the fields of the source
 structure are changed when the destination structure is updated.
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2541,8 +2541,11 @@ a `header_union` as an argument to an overlay..  A struct produced by
 an overlay can only contain fields of the base types: `bit<>` and
 `int<>`.  The only statement allowed within an overlay function is an
 assignment statement that assigns to a field of the destination.  The
-only expressions allowed within an overlay are field access, bit
-slicing and concatenation.
+only expressions allowed within an overlay are:
+- field access
+- bit slicing
+- concatenation
+- casts that convert in either direction between bit<W> and int<W> (but not casts that change the width)
 
 Under some restrictions overlays can be left-values:
 - they must be obtained by applying an overlay function to a left-value;
@@ -2555,30 +2558,67 @@ an overlay which is a left value will update the original structure or
 header.  If the input of an overlay is a header the rules about
 accessing uninitialized header fields are valid for the source header.
 
+The following example shows an use of overlays
+
 ~ Begin P4Example
-struct Addresses {
-    bit<32> source;
-    bit<32> destination;
+// In this header all fields are byte-aligned
+header IPv4_aligned {
+    bit<8>       version_ihl;
+    bit<8>       diffserv;
+    bit<16>      totalLen;
+    bit<16>      identification;
+    bit<16>      flags_fragOffset;
+    bit<8>       ttl;
+    bit<8>       protocol;
+    bit<16>      hdrChecksum;
+    bit<32>      srcAddr;
+    bit<32>      dstAddr;
 }
 
-struct Bytes {
-    bit<8> q0;
-    bit<8> q1;
-    bit<8> q2;
-    bit<8> q3;
+// Overlay of IPv4_aligned that exposes all fields
+struct IPv4 {
+    bit<4>       version;
+    bit<4>       ihl;
+    bit<8>       diffserv;
+    bit<16>      totalLen;
+    bit<16>      identification;
+    bit<3>       flags;
+    bit<13>      fragOffset;
+    bit<8>       ttl;
+    bit<8>       protocol;
+    bit<16>      hdrChecksum;
+    bit<32>      srcAddr;
+    bit<32>      dstAddr;
 }
 
-overlay Bytes sourceQuads(Addresses add) {
-   q0 = add.source[31:24];
-   q1 = add.source[23:16];
-   q2 = add.source[15:8];
-   q3 = add.source[7:0];
+overlay IPv4 expand(IPv4_aligned s) {
+    version = s.version_ihl[7:4];
+    ihl = s.version_ihl[3:0];
+    diffserv = s.diffserv;
+    totalLen = s.totalLen;
+    identification = s.identification;
+    flags = s.flags_fragOffset[15:13];
+    fragOffset = s.flags_fragOffset[12:0];
+    ttl = s.ttl;
+    protocol = s.protocol;
+    hdrChecksum = s.hdrChecksum;
+    srcAddr = s.srcAddr;
+    dstAddr = s.dstAddr;
 }
 
-Addresses a = { 0x01234567, 0x89ABCDEF };
-Fields    q = sourceQuads(a);  // q is { 0x01, 0x23, 0x45, 0x67 };
-q.q0 = 0; // q is { 0x0, 0x23, 0x45, 0x67 }, a is { 0x00234567, 0x89ABCDEF }
+parser P(...) {
+   IPv4_aligned header;
+   IPv4 fields = expand(header);
+   ...
+   state ipv4 {
+      packet.extract(header);
+      verify(fields.version == 5, error.UnsupportedVersion);
+   }
+}
 ~ End P4Example
+
+It is expected that such coding style will make it easier to generate
+efficient implementations on some architectures..
 
 ~ Begin P4Grammar
 overlayDeclaration
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2577,7 +2577,7 @@ overlay Bytes sourceQuads(Addresses add) {
 
 Addresses a = { 0x01234567, 0x89ABCDEF };
 Fields    q = sourceQuads(a);  // q is { 0x01, 0x23, 0x45, 0x67 };
-q.q0 = 0; // q is { 0x0, 0x23, 0x45, 0x67 }, a is { 0x00234556, 0x89ABCDEF }
+q.q0 = 0; // q is { 0x0, 0x23, 0x45, 0x67 }, a is { 0x00234567, 0x89ABCDEF }
 ~ End P4Example
 
 ~ Begin P4Grammar
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -7954,14 +7954,18 @@ The P4 compiler should provide:
 * Allow 1-bit signed values
 * Define the type of bit slices from signed and unsigned values to be unsigned.
 * Constrain `default` label position for `switch` statements.
-* Allow empty tuples.
-* Added `@deprecated` annotation.
-* Relaxed the structure of annotation bodies.
-* Removed the `@pkginfo` annotation, which is now defined by the P4Runtime specification.
-* Added `int` type (Section [#sec-infinite-precision-integers]).
-* Added error `ParserInvalidArgument` (Sections [#sec-packet-extract-two], [#sec-skip-bits]).
-* Clarified the significance of order of entries in `const entries` (Section [#sec-entries]).
+* Allow empty tuples (Seciton [#sec-tuple-types]).
+* Added `@deprecated` annotation (Section [#sec-deprecated-anno]).
+* Relaxed the structure of annotation bodies (Section [#sec-annotations]).
+* Removed the `@pkginfo` annotation. This is defined by the P4Runtime
+  specification.
+* Added `int` type (Section [#sec-infinite-precision-integers].)
+* Added error `ParserInvalidArgument` (Sections
+  [#sec-packet-extract-two], [#sec-skip-bits]).
+* Clarify the significance of order of entries in `const entries`
+  (Section [#sec-entries]).
 * Added methods to calculate header size (Section [#sec-ops-on-hdrs]).
+* Added structure overlays (Section [#sec-overlays]).
 
 ## Summary of changes made in version 1.1.0
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2517,11 +2517,11 @@ struct Fields {
 }
 
 overlay Fields sourceQuads(Addresses add) {
-   q0 = source.source[31:24];
-   q1 = source.source[23:16];
-   q2 = source.source[15:8];
-   q3 = source.source[7:0];
-   all = source.source ++ source.destination;
+   q0 = add.source[31:24];
+   q1 = add.source[23:16];
+   q2 = add.source[15:8];
+   q3 = add.source[7:0];
+   all = add.source ++ add.destination;
 }
 
 Addresses a = { 0x01234567, 0x89ABCDEF };
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2617,6 +2617,7 @@ parser P(...) {
 }
 ~ End P4Example
 
+In this example the `expand` overlay produces a left-value.
 It is expected that such coding style will make it easier to generate
 efficient implementations on some architectures..
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2530,7 +2530,8 @@ q.q0 = 0;   // q is { 0x00, 0x23, 0x45, 0x67, 0x0023456789ABCDEF }, a is { 0x002
 q.q1 = a.destination[7:0]; // q is { 0x00, 0xEF, 0x45, 0x67, 0x00EF456789ABCDEF }, a is { 0x00EF4567, 0x89ABCDEF }
 ~ End P4Example
 
-Applying an overlay to a left-value produces a left-value.
+Applying an overlay to a left-value produces a left-value.  An overlay
+must define all fields of the destination structure.
 
 Overlays can only be applied to `struct` and `header` objects, and
 they can only produce `struct` objects.  A struct produced by an
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1368,6 +1368,7 @@ declaration
     | errorDeclaration
     | matchKindDeclaration
     | functionDeclaration
+    | overlayDeclaration
     ;
 ~ End P4Grammar
 
@@ -2531,15 +2532,17 @@ a.source[7:0] = a.destination[7:0]; // q is { 0x00, 0x00, 0x00, 0xEF, 0x000000EF
 ~ End P4Example
 
 An overlay must define all fields of the destination structure.
+The "function" computing an overlay can only be used as an initializer
+in the declaration of a variable of the suitable type.
 
 Overlays can only be applied to `struct` and `header` objects, and
-they can only produce `struct` objects.  A struct produced by an
-overlay can only contain fields of the base types: `bit<>` and
+they can only produce `struct` objects, but one cannot use a member of
+a `header_union` as an argument to an overlay..  A struct produced by
+an overlay can only contain fields of the base types: `bit<>` and
 `int<>`.  The only statement allowed within an overlay function is an
 assignment statement that assigns to a field of the destination.  The
 only expressions allowed within an overlay are field access, bit
-slicing and concatenation.  Referring to `header_union` fields within
-an overlay is not allowed.
+slicing and concatenation.
 
 Under some restrictions overlays can be left-values:
 - they must be obtained by applying an overlay function to a left-value;
@@ -2548,9 +2551,9 @@ Under some restrictions overlays can be left-values:
 The second restriction is necessary to compute how the fields of the
 source structure are changed when the destination structure is
 updated.  In the previous example `q` is not a left value.  Updating
-an overlay which is a left value will update the original structure.
-If the input of an overlay is a header the rules about accesing
-uninitialized header fields are valid for the source header.
+an overlay which is a left value will update the original structure or
+header.  If the input of an overlay is a header the rules about
+accessing uninitialized header fields are valid for the source header.
 
 ~ Begin P4Example
 struct Addresses {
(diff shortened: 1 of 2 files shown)


chrispsommers:
This seems very clear and convenient.

gbrebner:
I agree that use of overlays is more transparent - and practical - than use of unions, so support this.

Compared to other overlay-style proposals, I like the function-like idea here since it decouples the overlay specification from the source struct/header specification, allowing separation of user overlays from library structs/headers, as Mihai points out.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2363,6 +2363,85 @@ struct Parsed_headers {
...
+struct Bits {
+   bit<2> b;
+}
+
+overlay Bits nonLeft(Addresses add) {
+   b = add.source[31] ++ add.source[31];
+}
+~ End P4Example
+
+an overaly built by applying `nonLeft` is never a left-value.  This is
```
jafingerhut: typo: "overaly"

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2363,6 +2363,85 @@ struct Parsed_headers {
...
+overlay Fields sourceQuads(Addresses add) {
+   q0 = add.source[31:24];
+   q1 = add.source[23:16];
+   q2 = add.source[15:8];
+   q3 = add.source[7:0];
+   all = add.source ++ add.destination;
+}
+
+Addresses a = { 0x01234567, 0x89ABCDEF };
+Fields    q = sourceQuads(a);  // q is { 0x01, 0x23, 0x45, 0x67, 0x0123456789ABCDEF };
```
jafingerhut: This appears to be not an assignment like most others in P4_16, in that by doing this it actually causes q to be overlayed with a.  Is that correct?

If so, is it allowed later to do another assignment to 'q' with a different overlay, or the same overlay with a different parameter other than 'a'?  If that is allowed, is it effectively changing what 'q' "points to" at that time?

jafingerhut: The example overlay sourceQuads does not meet the restrictions described later for the overlay being a left-value.  Should it still be allowed to assign values to fields of q, as is done immediately following in this example, if the overlay does not satisfy those restrictions?

Is it permitted to "stack" overlays?  That is, after defining q as shown above, can I define an overlay 'overlayA' that take type 'Fields' as an input parameter, and returns some other type T1, and then do 'T1 x = overlayA(q);', and now assignments to fields of x can modify fields of both q and a?  Is there any limit imposed by the feature on this?

What if I do the assignment to q above, then the one to 'x' in the previous paragraph, and then I assign 'a = newOverlay(x);'?

jafingerhut:
Is it allowed to do an if-then-else, or any other kind of branching in P4 program execution flow, where one branch assigns the return of an overlay to a variable like 'q' in the example, but a different overlay to 'q' in another branch?  What happens if one reads or writes 'q' after the if-then-else is done?  Should 'q' remember which other thing it is overlaid with from that point forward in the execution flow?

jafingerhut:
I suppose one possible restriction that would eliminate many of my questions would be: "You can only assign an overlay to a variable as an initialization expression, when the variable is defined."  But I do not know if you were thinking that was a restriction in this proposal.

antoninbas:
> I suppose one possible restriction that would eliminate many of my questions would be: "You can only assign an overlay to a variable as an initialization expression, when the variable is defined." But I do not know if you were thinking that was a restriction in this proposal.

I'm in favor of adding this restriction. Overlays are already a powerful construct but I don't think you should be able to "rebind" them (so a bit like C++ references).

mbudiu-vmw:
Yes, overlays must be initialized and cannot be changed.
I think you should be able to stack them.

jafingerhut:
I think it would be a good idea to mention the restriction that overlays must be initialized when the variable they are assigned to is declared, and then cannot be changed afterwards.

Another restriction on the body of overlays, that I think is intended but good to make explicit if it hasn't already been: it consists only of assignment statements.  All left-hand sides must be field names of the "returned" struct type, or slices of them.  All right-hand sides must be expressions with the already-described restriction on operators, and all "names" must be field references to a field of the parameter.

Antonin has an idea for generalizing this to overlays that can take multiple structs and/or headers as parameters.

I guess if a header is a parameter to an overlay, then all of the notes about reading and/or writing fields within an invalid header apply for the overlay as well, based upon the current validity of the overlaid header?

jafingerhut:
Mihai, if you were going to describe the 'type' of the expression after the '=' sign here:
```
Fields    q = sourceQuads(a);
```
one answer would be 'it is type Fields', but the assignment above is fundamentally different than the following sequence, right?
```
Fields a;
// code here to initialize fields of a
Field q = a;
```
The last assignment above is really allocating independent 'storage' for 'q', and then copying data from a's storage to q's storage.

Is another restriction you have in mind "expression of the form `overlay_name(parameter)` can _only_ appear as initialization expressions when defining a variable"?

If you think they should be allowed in other places in a program, where else?  For example, should `overlay_name(parameter)` be allowed as a term in an expression, like this?
```
// assume control foo has a first parameter, say direction 'in' but I do not think it matters much,
// with type sourceQuads

// is this call allowed?
foo.apply(sourceQuads(a));
```

I ask because I am trying to understand this proposal, which may generalize overlays a little bit from this original one: https://github.com/p4lang/p4-spec/issues/744#issuecomment-473160745

jafingerhut:
Perhaps it might be worth considering restricting overlays to the kind that cannot use the same bit more than once on a right-hand side expression in the overlay definition?  I do not know how useful people might consider it to allow that.  If that were disallowed, then there would be only "one kind" of overlay, the kind that could be used as an L-value.

antoninbas:
@jafingerhut That sounds like a reasonable restriction. Although, since I am planning on using overlays as a collection of field references, I wonder if in some cases it may be useful to allow repetitions. I think it was allowed for field lists in P4_14, but I don't recall anyone using this in practice. We can see how people feel about this at the next LDWG meeting.

jafingerhut:
Is the "left hand side" that is assigned to restricted to be a `struct` type?  If so, it would be good to say so.

If it can be other types, the next most "reasonable" one might be a `header` type.  But that raises questions, such as: is the header automatically made valid by an assignment like this?
```
my_header_t h1 = sourceQuads(a);
```
If `h1` isn't automatically made valid by such an assignment, then there are other parts of the spec that claim that assignments to fields values in an invalid header should not corrupt any state, and reads from its field values return an unspecified value.

Also, what if we later call `h1.setInvalid();`?  The cleanest answer seems to be "the original values are not affected, especially if they are members of structs".

The implementation gets a bit sticky, too, since then it seems you have introduced a new valid bit for `h1` that is unique to it, even though all of its fields are overlaid with other things.

We could start describing special cases for headers that are "overlay references" like this, but it seems to open a can of worms that we may want to put off, perhaps forever.

jafingerhut:
Note that if you allow `int<W>` fields in the overlay struct, can they overlay fields of type `bit<W>`?

Can one or both be slices?

If so, then making an overlay an L-value and making assignments to it seems to have a slight affect on the decision of whether the return type of `X[8:3]` where X is defined as `int<10> X;` should be of type `int<6>` or `bit<6>`.  Related discussion on that topic is on this issue: https://github.com/p4lang/p4-spec/pull/701

One possibility: avoid the issue entirely by restricting overlay fields to type `bit<W>`, disallowing `int<W>`.

Another: Allow `int<W>` in overlays, on left-hand or right-hand side, but only if there are no slices or concatenation involved.  Kind of weird restriction, though.

Another: Allow explicit casts between `int<W>` and `bit<W>` in the assignments of overlay definitions.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2406,39 +2406,88 @@ overlay Fields sourceQuads(Addresses add) {
...
+overlay Bytes sourceQuads(Addresses add) {
+   q0 = add.source[31:24];
+   q1 = add.source[23:16];
+   q2 = add.source[15:8];
+   q3 = add.source[7:0];
+}
+
+Addresses a = { 0x01234567, 0x89ABCDEF };
+Fields    q = sourceQuads(a);  // q is { 0x01, 0x23, 0x45, 0x67 };
+q.q0 = 0; // q is { 0x0, 0x23, 0x45, 0x67 }, a is { 0x00234556, 0x89ABCDEF }
```
jafingerhut: I think a's value here should begin with `{ 0x00234567`?

jafingerhut:
With the latest commit as of the time of this comment, Mihai has added this proposed restriction: "The "function" computing an overlay can only be used as an initializer in the declaration of a variable of the suitable type."

That does have the effect of simplifying the proposal in terms of implementation and cases to think about.

It also implies that one proposed use of overlays described in this comment on another issue: https://github.com/p4lang/p4-spec/issues/744#issuecomment-473160745

is no longer legal as it is written, in particular in the last line of this code snippet:
```
struct Hdr { } // my program headers
struct Meta { } // my program user-defined metadata

struct FieldList1_t {
  bit<8> a;
  bit<16> b;
}

struct FieldList2_t {
  bit<16> a;
  bit<16> b;
  bit<32> c;
}

struct FieldLists_t {
  FieldList1_t fl1;
  FieldList2_t fl2;
}

// extension to Mihai's proposal (the overlay can be built from multiple sources)
overlay FieldList1_t Overlay1 (Hdr hdrs, Meta meta) { 
  a = hdrs.ipv4.dstAddr[32:24];
  b = meta.f1;
}

overlay FieldList2_t Overlay2 (Meta meta) {
  a = meta.f1;
  b = meta.f5;
  c = meta.fb;
}

Hash<bit<16> /* output */> hash1;
hash1.get<FieldLists_t<_>>(Overlay1(hdrs, meta), Overlay2(meta));
```
because the overlay "function" calls Overlay1(...) and Overlay2(...) cannot appear at the specified point in a program.  Those lines could, I believe, be written with the meaning originally intended by that example as follows:
```
Hash<bit<16> /* output */> hash1;
FieldList1_t fl1 = Overlay1(hdrs, meta);
FieldList2_t fl2 = Overlay2(meta);
hash1.get<FieldLists_t<_>>({fl1, fl2});
```
Note: I added curly braces around the argument to the get() method call, which I think were missing from the original proposal.

@antoninbas @cc10512 Does the modified code above look reasonable to you?

If you find it objectionable in any way, note that it is still a kind of code that the compiler would likely transform the original code _into_ during some compilation phase.  At least, that is what I think the original code snippet you gave is equivalent to, give this proposal.

jafingerhut:
Another consequence of the restriction mentioned in my previous comment (and perhaps more generally), related to Calin and Antonin's proposal, motivated by a code snippet Antonin showed me.

Assume the context of all struct and overlay definitions in the code snippet in my previous comment.

```
// I think Mihai's proposed restriction makes the following line of code illegal, because
// the right-hand side initialization expression is not a _single_ overlay "function" call, but an
// expression that contains one or more such calls.
FieldLists_t val_to_hash = {Overlay1(hdrs, meta), Overlay2(meta)};
```
Mihai, would you say your intent was to make that line of code illegal?  I can't think of any problems if the answer is "yes".  Just wanted to be clear on that.

OK, another variation:
```
// code snippet (E)
FieldList1_t fl1 = Overlay1(hdrs, meta);
FieldList2_t fl2 = Overlay2(meta);
FieldList_t val_to_hash = {fl1, fl2};
hash1.get<FieldLists_t<_>>(val_to_hash);
```
From my reading and understanding, that code snippet above is legal, but it has a slightly different meaning than the snippet below:
```
// code snippet (F)
FieldList1_t fl1 = Overlay1(hdrs, meta);
FieldList2_t fl2 = Overlay2(meta);
hash1.get<FieldLists_t<_>>({fl1, fl2});
```
In (F), what is being passed to the get() method are aliases of the parts of the variables hdr and meta that are aliased by fl1 and fl2.  No "storage" outside of hdr and meta is implied for the compiler to implement this.

In (E), what is being passed to the get() method is val_to_hash, which _is_ declared as "storage" separate from, and can be modified independently of, the storage in hdr and meta aliased by fl1 and fl2.

If the get() parameter is direction 'in' only, then the same _values_ are passed in both (E) and (F) to get, so its calculated result should always be the same.

If the get() parameter was direction 'inout', then any modifications to the 'inout' parameter of code snippet (F) would cause those side effects on the values of the parts of hdr and meta that are aliased by fl1 and fl2, but for snippet (E) would cause those side effects on the values of val_to_hash, and _have no effect_ on any fields of hdr and meta, because they are not aliased by val_to_hash.

Corrections welcome if I have any of this wrong.  Just trying to make sure everyone is on the same page here with the meaning of overlays.

cc10512:
@jafingerhut thank you for your in-depth analysis. Declaring fl1 and fl2 separately to be initialized as Overlays is acceptable, but it does seem to add to the verbosity. Is there any reason for which Mihai's restriction to initializers does not apply to argument initializers? If they were, then the initial example will still be legal.

jafingerhut:
You would have to ask Mihai, but I suspect he may be open to the idea of this syntax being made a legal part of the proposal:
```
FieldLists_t val_to_hash = {Overlay1(hdrs, meta), Overlay2(meta)};
```
If that were legal, I suspect it would mean the same thing as the code below, and perhaps a compiler might make that transformation while compiling:
```
FieldList1_t compiler_temp1 = Overlay1(hdrs, meta);
FieldList2_t compiler_temp2 = Overlay2(meta);
hash1.get<FieldLists_t<_>>({compiler_temp1, compiler_temp2});
```

jafingerhut:
A little description of how this overlay proposal could let you define an arbitrary permutation of bits of one or more `bit<W>` fields, and what that means for correct vs. incorrect compiler transformations of expressions involving overlays: https://github.com/jafingerhut/p4-guide/tree/master/overlays#maybe-surprising-thing-that-overlays-can-do

jafingerhut:
Another note on implementation of fun.  While it seems possible to determine from the definition of a single overlay function, whether aliases created from it can be used as L-values or not, when you allow "stacking" of overlay functions, you can create aliases that cannot be used as L-values, even if none of the individual overlay functions prevents it.  This seems to me straightforward to determine for the compiler, or a persistent human who knows the rules: https://github.com/jafingerhut/p4-guide/tree/master/overlays#stacking-overlays-and-determining-whether-an-overlay-alias-can-be-an-l-value

mbudiu-vmw:
The ugliest part of P4-16 is the handling of side-effects, when you have an operation that may potentially update multiple left-values at once. If these left-values share the underlying storage you have to explain in which order the updates happen. What is nice about a struct is that the fields are known to be disjoint. That is not true about a P4-14 field-list in general. So using field-lists (or list expressions) introduces a bunch of problems that are quite ugly. However, if a left-value looks like a struct, i.e., all the fields are disjoint, then we can talk about updating it in a single statement, because the order that the fields are written does not matter.

Whatever we do with overlays tries to keep this constraint in mind. We can imagine lots of interesting extensions of the mechanism, but I would really like to avoid ordering-induced problems.

mbudiu-vmw:
Note that the overlay proposal, as written now, allows a single argument which must be a struct. So we know that the fields of the input structure cannot alias to each other, which simplifies the reasoning and the compiler analyses. If you allow multiple arguments to an overlay you may no longer have this property if the arguments are fields of the same union, or they are overlays themselves.

jafingerhut:
We can check with Calin and Antonin, but it seems that the ability to pass an overlay function at least two structs is, if not necessary, very convenient for their use case.

I agree this would make the compiler checking whether an overlay alias could be used as an L-value require a little more code, it certainly isn't a fancy algorithm or anything -- just follow the chain of aliases until you reach something that is not an alias.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2363,6 +2363,134 @@ struct Parsed_headers {
...
+
+Under some restrictions overlays can be left-values:
+- they must be obtained by applying an overlay function to a left-value;
+- each bit from a source field must be used at most once.
+
+The second restriction is necessary to compute how the fields of the
+source structure are changed when the destination structure is
+updated.  In the previous example `q` is not a left value.  Updating
+an overlay which is a left value will update the original structure.
+If the input of an overlay is a header the rules about accesing
```
jafingerhut: typo: "accesing"

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2363,6 +2363,134 @@ struct Parsed_headers {
...
+an overlay is not allowed.
+
+Under some restrictions overlays can be left-values:
+- they must be obtained by applying an overlay function to a left-value;
+- each bit from a source field must be used at most once.
+
+The second restriction is necessary to compute how the fields of the
+source structure are changed when the destination structure is
+updated.  In the previous example `q` is not a left value.  Updating
+an overlay which is a left value will update the original structure.
```
jafingerhut: "structure" -> "structure or header"?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2363,6 +2363,134 @@ struct Parsed_headers {
...
+   q3 = add.source[7:0];
+}
+
+Addresses a = { 0x01234567, 0x89ABCDEF };
+Fields    q = sourceQuads(a);  // q is { 0x01, 0x23, 0x45, 0x67 };
+q.q0 = 0; // q is { 0x0, 0x23, 0x45, 0x67 }, a is { 0x00234556, 0x89ABCDEF }
+~ End P4Example
+
+Overlays can be "stacked", as in the following example, where both `s`
+and `b` are left-values.
```
jafingerhut: Is the plan to simplify this proposal by disallowing "stacked" overlays?  If not, note that if both stacking and passing multiple parameters to an overlay function are allowed, it makes the compiler's job of determining whether an overlay can be used as an L-value a little bit more interesting: https://github.com/jafingerhut/p4-guide/tree/master/overlays#stacking-overlays-and-determining-whether-an-overlay-alias-can-be-an-l-value

mbudiu-vmw: I think the trouble is from the multiple arguments; consider the case of calling an overlay with the same argument twice: o(a,a). But I think we should start simple and grow, so I will remove the stacking for now as well until someone asks for it.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2377,6 +2378,108 @@ struct Parsed_headers {
...
+
+overlay Bytes sourceQuads(Addresses add) {
+   q0 = add.source[31:24];
+   q1 = add.source[23:16];
+   q2 = add.source[15:8];
+   q3 = add.source[7:0];
+}
+
+Addresses a = { 0x01234567, 0x89ABCDEF };
+Fields    q = sourceQuads(a);  // q is { 0x01, 0x23, 0x45, 0x67 };
```
mbudiu-vmw: Change Fields to Bytes

mbudiu-vmw:
BTW: I think I have a killer app for struct overlays. It really makes up for imperfect compilers, but it is likely to be very useful. Consider this way of writing a program:

```
header IPv4_aligned {
    bit<8>       version_ihl;
    bit<8>       diffserv;
    bit<16>      totalLen;
    bit<16>      identification;
    bit<16>      flags_fragOffset;
    bit<8>       ttl;
    bit<8>       protocol;
    bit<16>      hdrChecksum;
    bit<32>     srcAddr;
    bit<32>     dstAddr;
}

struct IPv4 {
    bit<4>       version;
    bit<4>       ihl;
    bit<8>       diffserv;
    bit<16>      totalLen;
    bit<16>      identification;
    bit<3>       flags;
    bit<13>      fragOffset;
    bit<8>       ttl;
    bit<8>       protocol;
    bit<16>      hdrChecksum;
    IPv4Address  srcAddr;
    IPv4Address  dstAddr;
}

overlay IPv4 expand(IPv4_aligned s) {
    version = s.version_ihl[7:4];
    ihl = s.version_ihl[3:0]
    diffserv = s.diffserv;
   ...
}
```

Now you can extract() the compact header, but access fields in the overlay in your program, making the program easier to read and write: you can access 'ihl' in the overlay without having to mask and shift in the original header.

At least the ebpf compiler will generate much more efficient code for this kind of implementation.

hanw:
Came across a use case that might need overlay to make the code more concise.

```
struct intrinsic_metadata {
    // more_than_10_fields_that_we_don_t_care;
    bit<32> flow_control;
    // another_dozen_of_fields_that_we_don_t_care;
} 
```

Can we support specifying overlay only on the 32 bit field within the struct, without writing down all fields in the struct in an overlay function?

jafingerhut:
@hanw I believe the answer is "yes".

An overlay function's "return value" probably needs to be "completely filled in".  That is your "??" type in your example.

The "input parameter" struct can be "partially used", i.e. only a small fraction of them might appear on right-hand sides of assignment statements in the overlay function.

mbudiu-vmw:
I have simplified the overlays in 3 ways:
- only allow left-value overlays (for now)
- the overlay function directly defines the structure; no need to use a separate struct declaration
- added a syntax to copy all of the fields of a structure `_ = _;`

jafingerhut:
I do not know if this is a minor side issue, fundamental issue, or somewhere in between, so I'll ask and let Mihai decide :-)

Right now if you have a control, parser, or function with a direction `in` parameter that is a struct or header, the body of that can read the entire value, or its fields, but cannot change them.

With this proposal for overlays, suppose inside of one of these control/parser/function with an `in` header, we create an overlay of that header parameter, and then try to assign a value to one of the overlay's fields.   Would/should the compiler reject that, as it does now if you try to make a direct assignment to a field of the `in` parameter?

mbudiu-vmw:
An overlay is a left-value only if it is applied to a left-value.

blp:
@mbudiu-vmw brought this proposal to my attention.

It's clever. I don't understand yet why the overlay syntax for `IPv4_aligned` and `IPv4` is better than, say:

```
header IPv4 {
    union {
        bit<8>       version_ihl;
        struct {
	    bit<4> version;
	    bit<4> ihl;
	};
    };
    bit<8>       diffserv;
    bit<16>      totalLen;
    bit<16>      identification;
    union {
        bit<16>      flags_fragOffset;
	struct {
	    bit<3> flags;
	    bit<13> fragOffset;
	};
    };
    bit<8>       ttl;
    bit<8>       protocol;
    bit<16>      hdrChecksum;
    bit<32>      srcAddr;
    bit<32>      dstAddr;
}
```

although the C-like syntax above (using anonymous struct and union members so that all the names are visible at the outer level) could no doubt be improved for use in P4.

mbudiu-vmw:
First, we don't have unions in P4 yet. 
Second, even when we'll have them they will be safe tagged unions probably. There is no tag here.
Third, the overlays as functions are a separate declaration from the original datatype. This allows you to import types from a library and declare your own overlays for them.

blp:
> First, we don't have unions in P4 yet.
> Second, even when we'll have them they will be safe tagged unions probably. There is no tag here.

I did not mean to propose the above as syntax; it would be a bad syntax for P4.

> Third, the overlays as functions are a separate declaration from the original datatype. This allows you to import types from a library and declare your own overlays for them.

Is that a commonly valuable use case? I'm more accustomed to the use case where I sometimes want to treat a bunch of bytes as a unit and sometimes as individual bits or bitwise subfields, sometimes changing from one line of code to another. That's usually obvious at the point of original declaration. It could feel a bit repetitive to add an extra `expand(...)` (or whatever) call at the head of each block that wanted to do that.

mbudiu-vmw:
We don't know if it's a valuable use case since the feature is not implemented yet.
Perhaps you are right, and we could have a additional qualifier for a field to indicate that it's an overlay of another field in the same struct.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2425,6 +2426,124 @@ struct Parsed_headers {
...
+overlay IPv4 expand(IPv4_aligned s) {
+    bit<4>  version = s.version_ihl[7:4];
+    bit<4>  ihl = s.version_ihl[3:0];
+    bit<3>  flags = s.flags_fragOffset[15:13];
+    bit<13> fragOffset = s.flags_fragOffset[12:0];
+    _ = _; // map all remaining fields unchanged
+}
+~ End P4Example
+
+An overlay must define all fields of the destination structure.  The
```
jfingerh, line 2498: I am not sure, but was the sentence "An overlay must define all fields of the destination structure." applicable when this PR was originally written, when the overlay required a separate struct definition for the overlay, separate from the struct or header that was overlaid?

Given that this has been removed, this sentence seems like it might be obsolete.  If it is not obsolete, what does the sentence mean?

jfingerh:
Can a struct or header that has struct members, that 'bottom out' in only types `bit<W>` and/or `int<W>`, have an overlay declared on it?  For example, could one declare an overlay for struct type `s2_t` declared below?
```
struct s1_t {
    bit<4> f1;
    bit<4> f2;
}
struct s2_t {
    s1_t x1;
    bit<8> f3;
}
```
That doesn't seem to go against anything said in the currently written PR, but might not have been intended, either.

If so, it seems that the proposed `_ = _` syntax would have two possibilities:
(a) only allow it if all of the remaining fields of the overlaid-upon header or struct have type `bit<W>` or `int<W>`
(b) allow it even if the overlaid-upon header or struct has remaining fields of a struct type, and allow sub-struct types to appear as members of the overlay struct type, too.

As currently written, it seems that this PR would imply alternative (a).

jnfoster:

PR #657, opened 2018-07-09
restrict digest extern to ingressDeparser

hanw:


jafingerhut:
@hanw Thanks for making the change here, if that is what was intended originally.  If we follow the idea of PSA specifying minimum levels of support, not prohibiting future extensions unless we think there is a very strong reason to prohibit something, it seems like it might be better to say something like "Digest need not be supported in the EgressDeparser" ?

PR #658, opened 2018-07-12
Bug fix:  '_' argument can be used for parameters with a default value

SantiagoBautista:


jnfoster:

PR #662, opened 2018-08-09
PSA - Add notes about empty groups for the ActionSelector extern

jafingerhut:


jafingerhut:
@konne-google I do not seem to be able to add you as a reviewer for this change, but if I recall correctly you were interested in this, so wanted to notify you of this proposed change in case you had comments.

konne-google:
This looks good. Thanks for the update!

p4-16/psa/PSA.mdk
```diff
@@ -2069,7 +2069,16 @@ specified as a parameter when instantiating the action selector.
 Action selector members may only specify action types defined
 in the `actions` attribute of the implemented table. All actions in a group must be
 of the same type. The action parameters for actions in the same group are allowed
-to differ, and the action of different groups in a selector may be different.
+to differ, and the action of different groups in a selector may be different.  More precisely, a PSA implementation is allowed to, but need not, support an action selector group where different members use different action types.
```
antoninbas: Line wrap?
Doesn't the precision contradict the statement above:

> All actions in a group must be of the same type.

p4-16/psa/PSA.mdk
```diff
@@ -2069,7 +2069,16 @@ specified as a parameter when instantiating the action selector.
...
+
+PSA implementations are allowed to support adding entries in a table
+that point at groups which currently have no members, but they are
+also allowed to have non-portable implementation specific behavior if
+this occurs in the data plane.  This limitation of a PSA data plane
```
antoninbas: double whitespace

p4-16/psa/PSA.mdk
```diff
@@ -2422,7 +2431,11 @@ In particular, ActionProfile and ActionSelector single operations,
 such as adding a member to a group, removing a member from a group,
 adding an empty group, deleting an empty group, or modifying the
 action parameters of an action added earlier to a group, should all be
-atomic.
+atomic.  One possible exception is a PSA implementation that uses the
```
antoninbas: double whitespace again it seems (and there are other occurrences as well). This is funny because I have seen it in a PR from Tom Everman as well; does it come from the text editor?

p4-16/psa/PSA.mdk
```diff
@@ -3128,3 +3141,165 @@ the performance of these protocols.
...
+            } else {
+                send_to_port(ostd_meta, user_meta.out_port);
+            }
+            // ... later ingress code goes here ...
+        }
+```
+
+In either case, an implementation might also support putting the `if`
+statement inside of the action `set_output_port`, but this is not
+required by PSA.
```
antoninbas: did you want to mention "fast-failover" as one possible reason for the group to become empty without action from the switch or controller software? That's a valid way to implement the P4Runtime "watch" attribute if the HW supports it. In this case, the behavior may be undefined until the software is notified that the group is empty and can take action.

p4-16/psa/PSA.mdk
```diff
@@ -3128,3 +3141,165 @@ the performance of these protocols.
...
+perspective is: issue the P4Runtime command to remove the last member
+from the group, and have a default action of "drop the packet" take
+effect for any packets applying the table and selecting the empty
+group.
+
+This behavior can be implemented via a small extra bit of logic in the
+P4Runtime server implementation (hereafter called the agent).  The
+basic idea is that first the controller specifies the action (with
+action parameter values) to be performed when the group is empty.  The
+details of the message are left for the P4Runtime API specification to
```
antoninbas: we can bring this up at the WG tomorrow

p4-16/psa/PSA.mdk
jafingerhut: I checked Wikipedia on this, out of curiosity.  Quoting from the first paragraph of the Wikipedia page below: "Many people prefer double sentence spacing for informal use because that was how they were taught to type."  <-- That is my reason.

https://en.wikipedia.org/wiki/Sentence_spacing

mbudiu-vmw: if you use emacs it has some very nice commands to operate on phrases.
phrases in emacs are delimited by a sentence terminator and two spaces.

p4-16/psa/PSA.mdk
jafingerhut: If the way it is written now is confusing, I would prefer to modify the sentence "All actions in a group must be of the same type."  Except in rare circumstances where we actually want to prevent something, I think PSA should specify "implementations should support _at least this_", and should usually not say "an implementation _cannot_ do X".

p4-16/psa/PSA.mdk
mbudiu-vmw: more precisely, that's the default regular expression for phrases. you can customize it. but it makes expressions such as i.e. or e.g. that and with a period unambiguous.

p4-16/psa/PSA.mdk
antoninbas: I think the first sentence should be changed. Right now IMO they strictly contradict each other.

p4-16/psa/PSA.mdk
antoninbas: @jafingerhut My bad I had no idea this was a thing. I should have Googled it first.

p4-16/psa/PSA.mdk
jafingerhut: Hashing this out in live conversation might be better, but I am also confused by this sentence of your comment: "In this case, the behavior may be undefined until the software is notified that the group is empty and can take action."  The reason for this new appendix is to _avoid_ undefined behavior, even if the data plane does not have predictable behavior for empty action selector groups.  The combined "data plane + P4Runtime agent" system can still have predictable portable behavior for times when the "action selector group is empty, as far as the controller knows".

p4-16/psa/PSA.mdk
jafingerhut: Eliminated the contradiction, by changing the whole paragraph.  Hopefully the meaning is still reasonably clear.

p4-16/psa/PSA.mdk
jafingerhut: @antoninbas Not bad on your part at all.  Until a month ago I did not even realize that a few people are vehemently anti-double-space-between-sentences.

antoninbas: Lol. Given that the using a single space after a period is called "French spacing" (https://en.wikipedia.org/wiki/History_of_sentence_spacing#French_and_English_spacing), I think my ignorance can be excused :)

jafingerhut:
@konne-google @cc10512 @vgurevich @wmohsin @hanw @antoninbas 

I have updated the text a bit based on discussion we had at the P4 API and architecture work group meetings held on 2018-Aug-15.

The basic changes are:

+ recommend (but do not require) that a PSA data plane table with action selector implementation, when matching a table entry that points at an empty group, executes the table's default action.  For PSA v1.1, a PSA data plane is allowed to have unpredictable packet processing behavior in that situation, but we may make the recommendation a requirement in some future PSA version.

+ the appendix is still there, but it explains why the workaround there doesn't really implement the recommendation.  I don't see a way to extend it so that it does, so I left it as an "almost does it" in case it helps implementers think of ways to achieve the goal for their implementation.

konne-google:
This looks good to me now.

antoninbas:
@jafingerhut my understanding was that we were considering adding a new table property to indicate which action to execute when the selector group is empty. Any reason why this has to be the same as the default action for a match miss?
On a related note, the current P4Runtime spec states that:

> In this P4Runtime release, we have decided to restrict the default entry for indirect tables - tables with an ActionProfile or ActionSelector implementation property - to a constant NoAction action entry, with the hope that it would simplify the implementation of the P4Runtime service.

See https://github.com/p4lang/PI/issues/306#issuecomment-364627953

jafingerhut:
@antoninbas No, no reason I can think of that the 'encountered an empty action selector group' action must be the same as the default action.  I probably thought of making the same while writing up the idea and thought it would simplify things if the default action for both were identical.

So if we make a new table property, e.g. one with a name like `psa_empty_group_action = <action_name>;`, it would only be correct to have such a property defined for a table with an action selector implementation.  Like a `default_action = <action_name>;`, it would specify the action initially assigned to the table for this purpose (i.e. execution if an empty action selector group is found during a match), and the P4Runtime API will provide a command by which a controller can change that action after loading the P4 program, like one can for non-const default_actions for tables?

jafingerhut:
@antoninbas @konne-google I have updated the PR to add a new table property `psa_empty_group_action`, specifying the action to be executed if an action selector table hits an empty group.

p4-16/psa/PSA.mdk
```diff
@@ -3128,3 +3170,192 @@ the performance of these protocols.
...
+Below we describe one way that nearly achieves the goals above, except
+that it requires the empty group action to have the same name as the
+non-empty groups of the action selector.  This may be too onerous of a
+restriction on system developers to be worth implementing.
+
+This behavior can be implemented via a small extra bit of logic in the
+P4Runtime server implementation (hereafter called the agent).  The
+basic idea is that first the controller specifies the action (with
+action parameter values) to be performed when the group is empty.  The
+details of the message are left for the P4Runtime API specification to
```
antoninbas: It is likely that now that we have the new table property, we will not add this to P4Runtime 1.0

p4-16/psa/PSA.mdk
```diff
@@ -2066,10 +2068,50 @@ to select a member from the group, and the action specification
...
+implementation specific behavior if this occurs in the data plane.  In
+some cases, it may be possible for the combination of a PSA data plane
+plus its P4Runtime server software to achieve this desired behavior,
+as far as the P4Runtime client controller software can observe.  See
+Appendix [#appendix-empty-action-selector-groups].
+
+The `psa_empty_group_action` property of a table is similar to the
+`default_action` property in the following ways:
+
++ They both have actions as their values.
```
antoninbas: For P4Runtime 1.0 - to avoid making the TableEntry message too complex - it is likely that we will require the "empty group action" to be const.

p4-16/psa/PSA.mdk
jafingerhut: Understood.  Do you think that this appendix is clear enough that it is not required behavior, but one possible idea that might help PSA implementers understand and/or create their own implementation of the empty group action behavior?

p4-16/psa/PSA.mdk
antoninbas: Yes we definitely to mention this in the spec. I have opened an issue to keep track of this: https://github.com/p4lang/p4runtime/issues/43

p4-16/psa/PSA.mdk
antoninbas: I think it's clear but that specific paragraph is maybe a bit out-of-sync with the rest. Up until that paragraph we talk about the "empty group action" (which can be set through the new table property). But in this paragraph we have:

> since PSA allows different groups to use different action
types, it seems best to allow the controller to specify a different
empty group action per group.

I understand that it is because you are trying to take into account the constraint that some HW may be limited to using the same action in any given group, while making sure that the solution is generic enough to allow different groups to use different actions.

In my case the most likely implementation will be to create a new group and update the match rules pointing to that group, but it is not ideal if there is more than one match rule for any given group.

p4-16/psa/PSA.mdk
```diff
@@ -3128,3 +3170,192 @@ the performance of these protocols.
...
+
+Achieving all of these requirements seems not to be possible with a
+PSA data plane implementation that meets only the minimum requirements
+for action selectors, i.e. one that restricts members of a group to
+all have the same action name, and that does not natively support
+predictable behavior if a group is empty in the data plane.
+
+Below we describe one way that nearly achieves the goals above, except
+that it requires the empty group action to have the same name as the
+non-empty groups of the action selector.  This may be too onerous of a
```
antoninbas: maybe it would be clearer to repeat the word "action" here.

jafingerhut: If you meant replacing "the same name as" with "the same action name as", then I have made that change.

If  you meant something else, let me know.

antoninbas: I may be misunderstanding the sentence, but I meant:
"it requires the empty group action to be the same as the action for non-empty groups"

antoninbas: I think we're on the same page, you should feel free to merge
I think replacing "name" with "action name" like you did in your latest commit made the sentence clearer

jafingerhut:
@cc10512 I think Antonin and Konstantin are OK with these changes now.  I have read through them one last time and done one hopefully final typo fix.  If you could give it a read, I'll be happy to merge it once it is in a state where you approve, too.

Here is a summary of the changes:

+ Recommend (but do not require) that a PSA data plane table with action selector implementation, when matching a table entry that points at an empty group, executes the action assigned to a new table property called `psa_empty_group_action`, similar to the existing `default_action` table property.  For PSA v1.1, a PSA data plane is allowed to have unpredictable packet processing behavior in that situation, but we may make the recommendation a requirement in some future PSA version.

+ A new appendix explains another way to achieve the desired empty group action effect, but only if the PSA implementation supports multiple different action names within the same action selector group at the same time (which is not required by PSA).  I would guess that such a technique might be useful to some PSA implementers in supporting the recommended empty group behavior.

p4-16/psa/PSA.mdk
```diff
@@ -2066,10 +2068,53 @@ to select a member from the group, and the action specification
...
-to differ, and the action of different groups in a selector may be different.
+Action selector members may only specify action types defined in the
+`actions` attribute of the implemented table.  PSA implementations are
+allowed to restrict all actions in a group to have the same action
+name, but they may allow more general configurations, i.e. different
+action names among members of the same group.  If a PSA implementation
+does have this restriction, it must still allow different members of
+the same group to have different action parameter values.  PSA
+implementations must also allow the action names of different groups
+in a selector to be different.
```
cc10512: I think this paragraph may benefit from a bit of reorganization for clarity.
In particular, it would be good to first list the requirements: must have at lease one action in each group, allow different actions in different groups. And introduce the `psa_emtpy_group_action` property (it simply occurs later) and its semantics: provide the must have action when there are no entries in the group. Then discuss the possible extensions and restrictions: may allow different parameter values, may restrict to one action.

p4-16/psa/PSA.mdk
```diff
@@ -2066,10 +2068,53 @@ to select a member from the group, and the action specification
...
+
+It is recommended, but not required, that a PSA data plane
+implementation perform the action specified as the value of the
+`psa_empty_group_action` table property if a packet applies the table
+and matches the entry `(K, G)`.  Note that the empty group action
+might have a different action name than `G` contained before, if
+earlier `G` was non-empty.  We will consider requiring such support in
+future versions of PSA.
+
+Until then, implementations are allowed to have non-portable
```
cc10512: Please specify what is supported today first, rather than base it on what will be required in the future. Then say this may be tightened in the future.

p4-16/psa/PSA.mdk
```diff
@@ -2066,10 +2068,53 @@ to select a member from the group, and the action specification
...
+implementation specific behavior if this occurs in the data plane.  In
+some cases, it may be possible for the combination of a PSA data plane
+plus its P4Runtime server software to achieve this desired behavior,
+as far as the P4Runtime client controller software can observe.  See
+Appendix [#appendix-empty-action-selector-groups].
+
+The `psa_empty_group_action` property of a table is similar to the
```
cc10512: Please move higher in the section. Have the spec first, then the example.

p4-16/psa/PSA.mdk
```diff
@@ -2066,10 +2068,53 @@ to select a member from the group, and the action specification
...
+`default_action` property in the following ways:
+
++ They both have actions as their values.
++ The P4 source code specifies the initial value.
++ If the table property is not given in the P4 source code, its value
```
cc10512: Is that the `psa_empty_group_action` property? Please be precise.

p4-16/psa/PSA.mdk
```diff
@@ -2066,10 +2068,53 @@ to select a member from the group, and the action specification
...
+  is `NoAction()`.
++ They may have a `const` modifier, indicating that control software
+  is not allowed to change this action.
++ In the absence of a `const` modifier, the control software is
+  allowed to change the action.  The P4 API working group has no plans
```
cc10512: Please remove the plans for the P4 API working group. It is not relevant, and some implementations of P4 runtime may already chose to implement such a message.

p4-16/psa/PSA.mdk
jafingerhut: Thanks for detailed review comments.  Take a look at the latest version when you get a chance.  I have attempted to address all of your comments, although I am only replying to this one.

cc10512:
Thank you for making the changes. Looks good to me!

jafingerhut:
Great.  I am going to hurry up and merge this now, before any more comments arise :)

PR #663, opened 2018-08-11
Remove PSA from P4-16 language spec open issues

jafingerhut:


jafingerhut:
Perhaps it is reasonable now to remove this section about PSA from the open issues in the language spec?

PR #675, opened 2018-09-27
Corrections to TeX markup for P4_14 v1.0.5 spec

jafingerhut:


jafingerhut:
I was trying to create a PR for P4_14 v1.0.5 spec, but found at least on my Ubuntu 18.04 machine that TeX was failing due to a few things, which at least in my testing these changes fix.

PR #676, opened 2018-09-27
Correction and clarifications to P4_14 spec on resubmit & recirculate…

jafingerhut:
… actions


jafingerhut:
See https://github.com/p4lang/p4c/issues/1516 for some background here.

This is not time-urgent by any means.  Like changes that were made in the P4_14 spec regarding how header stacks behave, several months ago, to reflect the behavior of actual P4_14 implementations, this is similar for resubmit and recirculate operations.  There are perhaps more clarifications that could be made for clone operations, but I wanted to start with these to verify whether these actually do reflect what P4_14 implementations do, and whether it was of interest to update the spec to match.

jafingerhut:
I messed up branching on my end when creating this PR.  Closing in favor of https://github.com/p4lang/p4-spec/pull/677

PR #678, opened 2018-09-28
PSA - updates and corrections to data plane to control plane numeric …

jafingerhut:
…translation

### DIFF
--- a/p4-16/psa/PSA.mdk
+++ b/p4-16/psa/PSA.mdk
@@ -326,11 +326,11 @@ Additional supported match_kind types
 ## Data plane vs. control plane data representations {#sec-data-plane-vs-control-plane-values}
 
 A PSA data plane implementation that supports the P4 Runtime
-API[^P4RuntimeAPI] includes software called a "P4 Runtime Agent" that
-enables runtime programming of the PSA device from a controller.  A
-controller may control multiple devices with different PSA
-implementations.  We will refer to these pieces of software as an
-"agent" and "controller" for brevity.
+API[^P4RuntimeAPI] includes software called a "P4 Runtime Server" that
+enables runtime programming of the PSA device from one or more "P4
+Runtime Clients".  We will call a P4 Runtime Server an "agent", and a
+P4 Runtime Client a "controller", here for brevity.  A controller may
+control multiple devices with different PSA implementations.
 
 As mentioned in section [#sec-psa-type-definitions], different PSA
 implementations are expected to customize the size of the data types
@@ -347,12 +347,15 @@ table with a million entries[^CacheCost].
 
 [^CacheCost]: While 10 Mbits sounds tiny to one accustomed to computers with hundreds of gigabytes of DRAM, the highest speed PSA implementations are ASICs that must keep tables in on-chip memories, similar to caches in general purpose CPUs.  The Intel i9-7980XE released in 2017 has 198 Mbits of on-chip L3 cache shared by its CPU cores.  Among Intel processors in Intel's 7th generation Core released in 2017 with at least 100 Mbits of L3 cache, they all cost close to $9 per Mbit of L3 cache.  <https://en.wikipedia.org/wiki/List_of_Intel_microprocessors>
 
-The P4 Runtime API uses quantities with bit widths independent of the target device to hold values of the types
+The P4 Runtime API uses quantities with bit widths independent
+of the target device to hold values of the types
 listed in section [#sec-psa-type-definitions], to simplify the
-manipulation of these values in the agent software.  For control plane
+manipulation of these values in the controller and agent software.
+For control plane
 operations on tables, any trimming or padding of values will be
-performed in the agent (trimming from controller to device direction,
-padding in device to controller direction).
+performed in the agent (usually trimming in the direction from
+controller to device, and padding in the direction from device
+to controller).
 
 There are multiple channels of communication over which such values
 might be carried between the controller and the data plane.  These
@@ -389,7 +392,7 @@ definition, as long as the other fields in the header satisfy the
 multiple of 8 bits restriction.  The controller or P4 program
 generating packets with such headers should fill in any most
 significant "padding" bits with 0.  You may do this with a normal
-assignment in statement in your P4 program, where the value on the
+assignment statement in your P4 program, where the value on the
 right hand side is cast to the wider `InHeader` type.  Similarly,
 casting a value of a wider type such as `PortIdInHeader_t` to the
 corresponding narrower type `PortId_t` truncates the excess most
@@ -405,11 +408,11 @@ numerical translation between controller port id values and data plane
 port id values, for each of the channels of communication between the
 controller and data plane described above.
 
-Occurrences of values of type `PortId_t` or `PortIdInHeader_t` must
-include a `p4runtime_translation` annotation so the compiler can mark
-occurrences of these values appropriately, so the agent can know which
-values need this translation.  See the example P4 header definition
-below.
+The `psa.p4` contains an annotation `p4runtime_translation` on the
+`type` definition of types `PortId_t` or `PortIdInHeader_t`.  This
+enables the compiler to mark all uses of values of these types that
+are accessible from the P4Runtime API, so the agent software knows to
+translate them.
 
 ```
 [INCLUDE=examples/psa-example-parser-error-handling.p4:PortId_Annotation_Example]
@@ -417,15 +420,16 @@ below.
 
 Because of this translation, we recommend that values of type
 `PortId_t` be treated similarly to values of an `enum` type.
-Comparing two values of this type for equality or inequality is
+Comparing two values of this type for equality or "not equal to" is
 reasonable, as well as assigning the values to other variables or
 parameters of the same type, but nearly any other operations are prone
 to error.  When matching values of type `PortId_t` as part of a table
 key, always match a complete value exactly, or wildcard every bit of
 the value (i.e. a `ternary` match kind with all bits wildcard, or an
 `lpm` match kind with prefix length 0).  If you attempt to do any of
-the following things on a value with type `PortId_t`, the numerical
-translation performed may lead to functional errors in your program:
+the following things on a value with type `PortId_t` or
+`PortIdInHeader_t`, the numerical translation performed may lead to
+functional errors in your program:
 
 + Do a table key match on a subset of the bits, or a range match.
 + Compare port values with relational operators like `<` or `>`.
@@ -450,8 +454,8 @@ translation occurs between the controller and the data plane.
 Below is a complete list of such types, and an example of the
 annotation to use in your P4 program.
 
-+ `PortId_t` or `PortIdInHeader_t` - annotation `@p4runtime_translation("port")`
-+ `ClassOfService_t` or `ClassOfServiceInHeader_t` - annotation `@p4runtime_translation("cos")`
++ `PortId_t` or `PortIdInHeader_t`
++ `ClassOfService_t` or `ClassOfServiceInHeader_t`
 
 For the types listed below, no numerical translation occurs, and the
 data plane must support all numerical values from 0 up to the maximum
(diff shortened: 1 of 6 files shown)


jafingerhut:
OK, I think I am done editing this now, except for handling any changes from reviewer comments.

p4-16/psa/PSA.mdk
```diff
@@ -293,15 +293,24 @@ few packets dropped for these reasons (perhaps only one).
...
+    for different targets would be annotations on externs, etc. that
+    the P4 compiler for that target needs to do its job.
+
+For each of these types, the P4 Runtime API[^P4RuntimeAPI] uses bit
+widths independent of the targets, large enough to hold a value for
+any target.  These widths are defined by the P4 Runtime API
+specification, but they are expected to be at least as large as the
+corresponding `InHeader_t` type below.  All PSA implementations must
+use data plane sizes for these types no wider than the corresponding
+types with `InHeader_t` in their names.
```
cc10512: It may be just me, but this reads a bit awkward. Suggestion:
```
For each of these types, the P4 Runtime API[^P4RuntimeAPI] uses bit
 widths independent of the targets.  These widths are defined by the P4 Runtime API
 specification, and they are expected to be at least as large as the
 corresponding `InHeader_t` type below, such that they hold a value for
 any target.  All PSA implementations must use data plane sizes for these
 types no wider than the corresponding `InHeader_t`-defined types.
```

p4-16/psa/PSA.mdk
```diff
@@ -405,27 +417,36 @@ numerical translation between controller port id values and data plane
...
-below.
-
-```
-[INCLUDE=examples/psa-example-parser-error-handling.p4:PortId_Annotation_Example]
-```
+The file `psa.p4` contains an annotation `p4runtime_translation` on
+the `type` definition of types `PortId_t` and `PortIdInHeader_t`.
+This enables the compiler to mark all uses of values of these types
+that are accessible from the P4Runtime API, so the agent software
+knows to translate them.  The benefit is that you do not need to put
```
cc10512: knows how to translate them?

p4-16/psa/PSA.mdk
```diff
@@ -405,27 +417,36 @@ numerical translation between controller port id values and data plane
...
+any special markings on your uses of values of these types throughout
+your P4 program.
+
+The cost of this approach is: if you want to do arithmetic on values
+of these types, you must explicitly cast them to a normal `bit<W>`
```
cc10512: normal is not defined. builtin type? core type?

p4-16/psa/PSA.mdk
```diff
@@ -405,27 +417,36 @@ numerical translation between controller port id values and data plane
...
+type.  The `psa.p4` include file defines `PortIdUint_t` as a `typedef`
+with exactly the same width in bits as `type` `PortId_t`, so you can
+cast values of type `PortId_t` to type `PortIdUint_t`, and then you
+can perform all P4 arithmetic operations on the value.  The result
+must be explicitly cast back to type `PortId_t` if you wish to assign
+it to a metadata field with that type.  Corresponding types with
+`Uint` in their name are defined for all PSA types.
```
cc10512: Can we also have a note to compiler implementers that mentions what should be the recommended behavior if this is not followed?

p4-16/psa/psa.p4
```diff
@@ -83,15 +135,64 @@ const   CloneSessionId_t PSA_CLONE_SESSION_TO_CPU = unspecified;
...
+#define PSA_EGRESS_INSTANCE_HEADER_TO_INT(x) ((EgressInstance_t) (EgressInstanceUint_t) (EgressInstanceInHeaderUint_t) (x))
+#define PSA_TIMESTAMP_HEADER_TO_INT(x) ((Timestamp_t) (TimestampUint_t) (TimestampInHeaderUint_t) (x))
+
+#define PSA_PORT_ID_INT_TO_HEADER(x) ((PortIdInHeader_t) ((PortIdInHeaderUint_t) ((PortIdUint_t) (x))))
+#define PSA_MULTICAST_GROUP_INT_TO_HEADER(x) ((MulticastGroupInHeader_t) (MulticastGroupInHeaderUint_t) (MulticastGroupUint_t) (x))
+#define PSA_CLONE_SESSION_INT_TO_HEADER(x) ((CloneSessionIdInHeader_t) (CloneSessionIdInHeaderUint_t) (CloneSessionIdUint_t) (x))
+#define PSA_CLASS_OF_SERVICE_INT_TO_HEADER(x) ((ClassOfServiceInHeader_t) (ClassOfServiceInHeaderUint_t) (ClassOfServiceUint_t) (x))
+#define PSA_PACKET_LENGTH_INT_TO_HEADER(x) ((PacketLengthInHeader_t) (PacketLengthInHeaderUint_t) (PacketLengthUint_t) (x))
+#define PSA_EGRESS_INSTANCE_INT_TO_HEADER(x) ((EgressInstanceInHeader_t) (EgressInstanceInHeaderUint_t) (EgressInstanceUint_t) (x))
+#define PSA_TIMESTAMP_INT_TO_HEADER(x) ((TimestampInHeader_t) (TimestampInHeaderUint_t) (TimestampUint_t) (x))
```
cc10512: Seriously? Is this the best we can do? I usually find that when we need to do this kind of stuff, something is wrong with the design.

jafingerhut: There may be something wrong with the design.  No argument there.  Frankly, right now I don't have any ideas to make it better, unless we consider it to be a better design if we force PSA program writers to add annotations to _every_ occurrence of values of these types requiring translation, and they get a buggy system if they forget.

These macros are only useful when dealing with headers in data-plane-to-controller and controller-to-data-plane packets, if that helps you swallow the nasty-tasting-pill any more easily.

p4-16/psa/PSA.mdk
jafingerhut: I do not understand what you mean by "if this is not followed".  The definition of `type` in the P4_16 language specification allows you to do nothing with values of those types except: assign them to other values with an identical type, compare them for == or != with values of an identical type, or cast them to the "underlying" type.  You can also cast values from the underlying type to the type.

Any P4_16 compiler that does anything else isn't a P4_16 compiler (or implements an extension not in the spec).

p4-16/psa/psa.p4
jafingerhut: I don't know if this helps with thinking about the problem, but if we want the compiler to be able to determine that a value of type `PortId_t` is actually `PortId_t`, and not some other type, imagine we tried to do it with typedefs, which allow arithmetic on them with other things that are the same bit width.

```
typedef bit<9> PortId_t;
typedef bit<9> ClassOfService_t;

// ... later in the program ...

PortId_t x;
ClassOfService_t y;

    table foo {
        key = {
            x + y : exact;
        }
        // ... rest of table defn here ...
    }
```

What is the type of that table key?  If you say PortId_t, why shouldn't it be ClassOfService_t?  What is a deterministic rule you are happy living with that all P4 compilers can use to find the answer?

And a program can have any number of typedefs with the same bit width, not just 2.

jafingerhut: I know @jnfoster believes there is a better way of approaching this whole area, but may not have free time to work on it.  I'm not trying to twist your arm on this, Nate, if that is still the case, but if you might be able to find 30 mins to brainstorm any ideas you have had on the topic with Calin and I, maybe we could take the ball and try running with it.

jnfoster:
I'd love to do this. Let's find a time off-GitHub.

-N

p4-16/psa/psa.p4
```diff
@@ -83,15 +135,64 @@ const   CloneSessionId_t PSA_CLONE_SESSION_TO_CPU = unspecified;
...
+type bit<16> CloneSessionIdInHeader_t;
+@p4runtime_translation("p4.org/psa/v1/ClassOfServiceInHeader_t", 8)
+type bit<8>  ClassOfServiceInHeader_t;
+type bit<16> PacketLengthInHeader_t;
+type bit<16> EgressInstanceInHeader_t;
+type bit<64> TimestampInHeader_t;
+
+/* See the comments near the definition of PortIdUint_t for why these
+ * typedef definitions exist. */
+typedef bit<32> PortIdInHeaderUint_t;
```
mbudiu-vmw: Maybe you want to move these before the `type`s and use them there.
`type PortIdInHeaqderUint_t PortIdInHeader_t;`
One less chance to get it wrong.

jafingerhut: Nice idea.  I tried it, but latest p4test/p4c give error when trying to use a value of such a `type` in a struct.  See https://github.com/p4lang/p4c/issues/1535

p4-16/psa/PSA.mdk
jafingerhut: Your suggestion looks like a reasonable replacement to me.  Copying yours.

p4-16/psa/PSA.mdk
jafingerhut: In next commit, I will try "knows that it must translate them, and what kind of translation to perform".

p4-16/psa/PSA.mdk
jafingerhut: Next commit will remove "normal", leaving it simply "must explicitly cast them to a `bit<W>` type"

p4-16/psa/PSA.mdk
jafingerhut: I am not changing anything related to this in the next commit, since it is not yet clear to me what should be changed.

p4-16/psa/psa.p4
jafingerhut: One of Calin or Nate proposed the idea that became this PR, which would reduce the number of casts needed from 3 down to 1: https://github.com/p4lang/p4-spec/issues/682.  My only plan for a change in this area in the next commit is to replace the macro definitions with P4_16 functions that do the same thing, since macros with arguments are not required by the P4_16 language spec.

p4-16/psa/psa.p4
jafingerhut: Leaving this unchanged for now, but will consider using the idea when the bug fix is merged in, probably soon-ish.

p4-16/psa/PSA.mdk
```diff
@@ -293,15 +293,24 @@ few packets dropped for these reasons (perhaps only one).
...
+from one PSA implementation to another[^PSAP4TargetSpecific].
+
+[^PSAP4TargetSpecific]: It is expected that `psa.p4` include files for
+    different targets will typically be nearly identical to each
+    other.  Besides the possibility of differing bit widths for these
+    PSA types, the only expected differences between `psa.p4` files
+    for different targets would be annotations on externs, etc. that
+    the P4 compiler for that target needs to do its job.
+
+For each of these types, the P4 Runtime API[^P4RuntimeAPI] uses bit
```
antoninbas: "may use" maybe?
I think the P4Runtime does say it is optional. In some cases maybe a controller does want direct access to the target-specific representation and numbering space.

p4-16/psa/PSA.mdk
```diff
@@ -447,20 +469,31 @@ The list above is not intended to be exhaustive.
...
 supported value for each of these types.
 
-+ `MulticastGroup_t`
+[^P4CNumericTranslationOptions]: The open source `p4c` P4 compiler is
+    planned to support an option to enable numerical translation for
+    additional types, without modifying your P4 program, nor the
+    `psa.p4` include file.  These additional types would be specified
+    by their name.
+
++ `MulticastGroup_t` - 0 is a special value that indicates no
```
antoninbas: do you know why there seems to be an asymmetry between multicast and cloning? IIRC there is a boolean flag to indicate that cloning is required.

p4-16/psa/PSA.mdk
jafingerhut: Sure.  The next commit will change "uses" to "may use" in that sentence.

p4-16/psa/PSA.mdk
jafingerhut: Antonin, why do you ask such darned reasonable questions? :-)

The multicast group specialness of 0 is kind of inherited from P4_14 and v1model.  There didn't seem to be much reason to change it.

Cloning being different, via using a boolean flag to indicate that cloning should be performed for a packet, is likely my doing if I recall correctly.  It was not a conscious decision to make them different, but obviously neither was there a conscious decision to make them similar.  I can't think of any strong reason to choose one approach over the other for cloning, although perhaps consistency would have been nice here (e.g. making clone session id 0 mean no cloning, and != 0 meaning do cloning, like the multicast group id).

p4-16/psa/PSA.mdk
cc10512: I missed that `type` is defined in the language as having a restricted set of operations, and thus any compiler should reject arithmetic ops. I was thinking of it more as a `typedef`. Looks good.

p4-16/psa/psa.p4
jafingerhut: I have made a commit after you approved, Calin, but it was only to follow this suggestion of Mihai's to define the typedef types first, then after that define the `type` types based on the `typedef` types, so they automatically have the same width as each other without having the number in the file twice.

I also updated a few comments in the psa.p4 file, I believe accurately, but you may want to take a quick look at just the changes I made in the most recent commit on 2018-Oct-05 to double check.

jafingerhut:
@cc10512 Do we think this is ready to be merged in?

cc10512:
Yes, certainly.

PR #681, opened 2018-10-02
P4-16 Add restrictions on use of functions to appendix tables

jafingerhut:


p4-16/spec/P4-16-spec.mdk
```diff
@@ -7196,11 +7198,16 @@ constructor parameters to other types.
 | parser      | yes     | yes    |  no     |  no    |
 | control     | yes     |  no    | yes     |  no    |
 | extern      | yes     | yes    | yes     | yes    |
+| function    |  no     |  no    |  no     |  no    |
 | table       |  no     |  no    |  no     |  no    |
 | value-set   |  no     |  no    |  no     |  no    |
 | value types | yes     | yes    | yes     | yes    |
 |-------------|---------|--------|---------|--------|
 
+Note: There is work under way that may add to P4~16~ the ability to
```
mbudiu-vmw: I don't know if we have to say that in the spec.

PR #685, opened 2018-10-20
PSA - Remove obsolete extern definition of ValueSet from psa.p4 inclu…

jafingerhut:
…de file

The tentative definition of the ValueSet extern was left in the psa.p4
include file even after all descriptions of it were removed from
PSA.mdk.  Now that parser value sets have been added to the P4_16
language spec, we should remove all traces of the old extern
definition from PSA files.  This includes removing some value set
example programs from Makefile.

Other updates to the Makefile:

+ I tested and found that the latest version of p4test can compile the
  psa-example-incremental-checksum2.p4 program without errors, so
  added it to the list of programs compiled via 'make check'.

+ I tested the latest p4test with psa-example-register1.p4, after
  making changes to that program that were recently made to
  psa-example-register2.p4, and it still gives a similar error as it
  did a year ago due to limitation in p4c where extern methods cannot
  return structs.  Updated the date of the latest version of p4test
  that gives error in Makefile comments to today.

### DIFF
--- a/p4-16/psa/Makefile
+++ b/p4-16/psa/Makefile
@@ -29,6 +29,7 @@ check:
 	${P4C} examples/psa-example-counters.p4
 	${P4C} examples/psa-example-digest.p4
 	${P4C} examples/psa-example-incremental-checksum.p4
+	${P4C} examples/psa-example-incremental-checksum2.p4
 	${P4C} examples/psa-example-parser-checksum.p4
 	${P4C} examples/psa-example-parser-error-handling.p4
 	${P4C} examples/psa-example-recirculate.p4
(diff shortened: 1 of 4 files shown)


jafingerhut:
@cc10512 If it makes it easier to review these changes, I can separate out the "remove obsolete ValueSet definition from psa.p4" change from the other stuff.  The other stuff is really just cleaning up one or two of the PSA example programs that without these changes give compilation errors, because they haven't been test-compiled and fixed since they were written.

p4-16/psa/Makefile
```diff
@@ -40,23 +41,10 @@ check:
...
-#	${P4C} examples/psa-example-incremental-checksum2.p4
-
 # psa-example-register1.p4 is almost identical to
 # psa-example-register2.p4, except that it attempts to return a struct
 # from a register read, and write a struct back to the register.  As
-# of 2017-Dec-04, the latest version of p4test gives an error if you
-# attempt this.
+# of 2018-Oct-19, the latest version of p4test gives an error if you
+# attempt this, I believe somewhere in one of the mid end passes.  It
+# is definitely not in the bmv2 back end code.
```
cc10512: We should track this down and fix. I'll take a look before the release.

mbudiu-vmw: I don't think bmv2 supports registers with structs.
We can implement this in the compiler by using bit slices and concatenations if it is necessary, but I am not sure how useful this will be in general. The main question is about field alignment in a register and what hardware can support in general.

jafingerhut: I think it would be nice to have, but realize there are other useful things to work on besides this.

The example program was the first one I tried to write using registers with multiple 'fields' inside of the register array elements, and when it did not work I left the example around as a reminder, but did not include it in the PSA spec document, including only the variant of the program that used explicit bit slices and concatentation instead, since that compiles with p4c.  I can create a separate issue for p4c for it, with Mihai's comment, in case that helps to track it.  It sounds like it may require enhancements in bmv2 as well.

jafingerhut: Created this issue just to record it somewhere closer to where changes would be needed to make the enhancement: https://github.com/p4lang/p4c/issues/1591

PR #688, opened 2018-11-05
adds idle_timeout specification

cc10512:


p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,59 @@ table.
...
+```
+enum PSA_IdleTimeout_t {
+  PSA_NO_TIMEOUT,
+  PSA_NOTIFY_CONTROL
+}
+
+table t {
+  key = {};
+  actions = {};
+  psa_idle_timeout = PSA_NOTIFY_CONTROL;
```
antoninbas: would be `PSA_IdleTimeout_t.PSA_NOTIFY_CONTROL`

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,59 @@ table.
 A PSA implementation must implement tables that have both a
 `psa_direct_counter` and `psa_direct_meter` property.
 
+### Table entry timeout notification {#sec-idle-timeout}
+
+PSA uses the `psa_idle_timeout` to enable a table implementation send
```
antoninbas: typo s/send/to send/

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,59 @@ table.
...
+notifications from the PSA device when a configurable time has passed since an
+entry was last matched. The property may take one of two values --
+`PSA_NO_TIMEOUT`, and `PSA_NOTIFY_CONTROL`. `PSA_NO_TIMEOUT` disables idle
```
antoninbas: maybe you should say that `PSA_NO_TIMEOUT` is the default when the property is omitted

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,59 @@ table.
...
+timeout support for the table. `PSA_NOTIFY_CONTROL` enables the notification. A
+PSA implementation will then generate an API for the control plane set
```
antoninbas: typo s/set/to set/

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,59 @@ table.
...
+time-to-live (TTL) values for table entries and if at any time during its
+lifetime, the table entry is not "hit" (&ie; not selected by any packet lookup)
+for a lapse of time greater or equal to its TTL, the device should generate a
+notification to the control plane.
```
antoninbas: Can we add a sentence stating that how these notifications are generated (buffering, batching, rate-limiting) is up to the exact control-plane API being used?

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,59 @@ table.
...
+  enabled the idle timeout property, the entry will not generate a
+  notification. The entry will participate in the aging process, with the
+  maximum timeout interval possible, i.e., its idle timeout counter will be set
+  to the maximum interval possible and the entry may be deleted when the
+  maximum interval elapses.
+
+- PSA does not require a timeout value for a default action entry. The reason
+  for not making this mandatory in the specification is that the default action
+  may not have an actual entry in the table and it is may be hit on the miss
+  table misses. Not sure what an TTL means for a miss path. The default action entry will not be aged out.
```
antoninbas: nit: wrap the text?

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,59 @@ table.
...
+
+- ActionSelectors and ActionProfiles will inherit the idle timeout property
```
antoninbas: I think this is ambiguous. The idle timeout process does not depend on the table's implementation (direct or indirect). But that only includes the match entry part, not the member / group the entry is pointing to. Even if a member is only used by a single match entry and that entry expires, I don't think the member should be removed.

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,59 @@ table.
...
+  bits to represent the values, and, since the values are programmed at
+  runtime, it is the responsability of the runtime (P4Runtime or other
+  controller software) to guarantee that the TTL values can be represented in
+  the device. This can be done by scaling the values to the number of bits
+  available on the platform, ensuring that the range of values between
+  different entries are representable. A PSA implementation should only enable
+  the programming of such tables, and return an error if the device does not
+  support the idle timeout at all.
+
+- If no value is programmed for a table entry, even though the table has
```
antoninbas: I'm not sure I understand this paragraph. If no notification is generated, then the entry will not be deleted since the control is in charge of removing the entry.

p4-16/psa/PSA.mdk
jafingerhut: It might be nice to have at least one key and action with example names in there.  I know it is just an example, but I don't think you can actually add table entries if there are no key fields, so for such a table it would be useless to specify an idle timeout.

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,59 @@ table.
...
+  actions = {};
+  psa_idle_timeout = PSA_NOTIFY_CONTROL;
+}
+```
+
+Restrictions on the TTL values and notifications:
+
+- It is likely that any hardware implementation will have a limited number of
+  bits to represent the values, and, since the values are programmed at
+  runtime, it is the responsability of the runtime (P4Runtime or other
```
jafingerhut: s/responsability/responsibility/

p4-16/psa/PSA.mdk
jafingerhut: Yes, the name `NOTIFY_CONTROL` is intended to signify that the only thing the PSA device does is notify the controller about these long-time-since-last-matched table entries.  It is up to the controller what to do about that, and the device _must not_ remove them on its own, without an explicit request from a controller to do so.

p4-16/psa/PSA.mdk
jafingerhut: @antoninbas @cc10512 Would you be willing to specify that for tables with psa_idle_timeout equal to NOTIFY_CONTROL, that they _must never_ generate a notification message for the 'default entry'?  Even if an implementation actually implemented `default_action` using some wildcard ternary table entry with lowest priority, for example, the P4Runtime server software could avoid generating any timeout notification messages to the controller if that entry went unmatched for a long time.

Does the P4Runtime API allow the controller to specify a timeout duration for a table's default entry?  I don't think it needs to, personally, but am open to other views on that detail.

p4-16/psa/PSA.mdk
jafingerhut: If it gets too confusing/weird, we could consider explicitly saying that PSA does not (yet) specify how the idle timeout feature works with tables that have action selector or action profile implementations, and implementations need not support it in order to be PSA v1.1. compliant.  I believe the PSA spec already says that implementations need not support direct meters or direct counters on tables that have action profiles/selectors, and timeouts are similar to those things in some ways.

p4-16/psa/PSA.mdk
antoninbas: P4Runtime does not make the distinction between the default entry & regular match entries in this case.
I feel like it does not harm to leave it as optional. I can add a note in the P4Runtime spec stating that idle timeout support for default entry is optional to keep things consistent. But I also do not see a use case for this specific feature, so I'm fine with prohibiting the use of idle timeout with default entries.

p4-16/psa/PSA.mdk
antoninbas: I'm fine with making support optional. But I think it's fairly easy to have idle timeout semantics for match tables regarding of the implementation, by limiting the idle timeout feature to the match "part" of the entry.

p4-16/psa/PSA.mdk
cc10512: actually enable an implementation send notifications is also correct.

antoninbas: my bad that's what I get for commenting on syntax as a non-native speaker :/

p4-16/psa/PSA.mdk
cc10512: done

p4-16/psa/PSA.mdk
cc10512: ok, removed the bit about deleting the entry. However, what do we want to say about the default value. Do we want to specify that by default is the maximum value? A different -- do not notify value? Or leave it to implementers?

p4-16/psa/PSA.mdk
cc10512: I'd prefer to leave it to implementers as well. Even though it may seem counter to portability, there will be situations where one might want timeouts on default entries, and then it is just a matter of the target supporting it or not. Unfortunately it is not something that the compiler can detect, so runtime API will have to have the means of knowing whether setting the timeout succeeded.

p4-16/psa/PSA.mdk
cc10512: restricted and left for future work.

cc10512:
@jafingerhut, @antoninbas Thank you for the comments. I hope I addressed your concerns, please take another look.

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,65 @@ table.
...
+  the programming of such tables, and return an error if the device does not
+  support the idle timeout at all.
+
+- If no value is programmed for a table entry, even though the table has
+  enabled the idle timeout property, the entry will not generate a
+  notification.
+
+- PSA does not require a timeout value for a default action entry. The reason
+  for not making this mandatory in the specification is that the default action
+  may not have an actual entry in the table and it is may be hit on the miss
```
antoninbas: "it is may be hit on the miss table misses"?

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,65 @@ table.
...
+  table misses. Not sure what an TTL means for a miss path. The default action
```
antoninbas: I would remove this sentence ("Not sure what an TTL means for a miss path"). If you don't remove it, you probably need to replace "an TTL" with "a TTL".

p4-16/psa/PSA.mdk
```diff
@@ -1497,6 +1499,65 @@ table.
...
+`psa_idle_timeout` properties on the same table.
+
+### Table entry timeout notification {#sec-idle-timeout}
+
+PSA uses the `psa_idle_timeout` to enable a table implementation send
+notifications from the PSA device when a configurable time has passed since an
+entry was last matched. The property may take one of two values --
+`NO_TIMEOUT`, and `NOTIFY_CONTROL`. `NO_TIMEOUT` disables idle timeout support
+for the table and it is the default value when the property is not present.
+`PSA_NOTIFY_CONTROL` enables the notification. A PSA implementation will then
```
jafingerhut: s/PSA_NOTIFY_CONTROL/NOTIFY_CONTROL/

p4-16/psa/PSA.mdk
cc10512: Replaced. Sounds a lot better :)

p4-16/psa/PSA.mdk
cc10512: +1

p4-16/psa/PSA.mdk
jafingerhut: @cc10512 Not sure if this little replacement has been made yet.  I think these changes are ready to commit now.  I will try to look at them with fresh eyes once we are in the 2 week per-release review period, and if there are any minor things like this we can fix them then.

cc10512: I did make it. Refresh?

jafingerhut:
This looks good to me.  I verbally approve this, because I can't figure out how to let Github allow me to mark it as approved in the normal way.   Grrr.  Please merge it in, too, or I can do it if you prefer.

PR #689, opened 2018-11-06
Add a bash script that installs Madoko on Ubuntu Linux systems

jafingerhut:


p4-16/spec/setup-for-ubuntu-linux.sh
```diff
@@ -0,0 +1,81 @@
...
+# used disk space, although temporarily went about 1 GB more than that
+# before 'sudo apt clean'.
+
+# On a freshly installed Ubuntu 18.04 system, added about 0.8G.
+
+# Retrieve and install fonts
+curl -fsSL -O https://raw.github.com/p4lang/p4-spec/gh-pages/fonts/UtopiaStd-Regular.otf
+curl -fsSL -O https://raw.github.com/p4lang/p4-spec/gh-pages/fonts/luximr.ttf
+mkdir -p "${FONT_INSTALL_DIR}"
+/bin/mv UtopiaStd-Regular.otf luximr.ttf "${FONT_INSTALL_DIR}"
```
cc10512: Why not make the directory and then use the `--output <file>` option to put the fonts directly in the directory?

p4-16/spec/setup-for-ubuntu-linux.sh
```diff
@@ -0,0 +1,81 @@
...
+sudo apt-get install git curl make nodejs npm texlive-xetex dvipng
+
+if [[ "${UBUNTU_RELEASE}" > "18" ]]
+then
+    # Only needed for Ubuntu 18.04
+    sudo apt-get install texlive-science
+else
+    # Only needed for Ubuntu 16.04
+    sudo apt-get install nodejs-legacy texlive-generic-extra texlive-math-extra
+fi
```
cc10512: I wish they would be more consistent in naming their packages ...

PR #690, opened 2018-11-07
One correction, and one slight enhancement, to recently added setup s…

jafingerhut:
…cript


jafingerhut:
Small bug -- I hadn't _actually_ tested it on an Ubuntu 18.04 system before, and it had a trivial failure when I did, fixed with this change.  Also added a small enhancement suggested by Calin on the font installation.

PR #696, opened 2018-11-20
PSA add link to example PSA programs in specification

jafingerhut:
Also change word "distinctive" to "unusual", since "distinctive" means
"characteristic of one person or thing", and it was being used to
refer to the numerical translation property of type PortId_t values,
which we expect will be used for multiple types, not only PortId_t.


jafingerhut:
No big changes here, Calin.  It just bugged me that the PSA spec didn't have a link to the PSA example programs in the Github repository.

Issue #668, opened 2018-09-08
bitslicing signed bitstrings


ChrisDodd:
I ran into an underspecified hole in the type system specification around slicing one bitstrings.  Section 8.5 talks about bitstring extraction operations `[m:l]` but only talks about the relations between the sizes of the of the source and resulting bitstring, and never mentions signedness (`bit` vs `int`).  Obviously an extraction from an unsigned `bit<n>` bitstring should be an unsigned `bit<k>` bitstring, but what about an extract from a signed `int<n>` bitstring?  There are arguments either way.

The p4c implementation preserves the signedness, so an extract from a signed `int<n>` bitstring is always a signed `int<k>` type, but this has an unfortunate corollary:  you can't extract a single bit from a signed bitstring as that would have the disallowed type `int<1>` (7.1.6.3).

We should clean up the hole, but there are a number of ways we could do it

1. Codify the p4c behvior of preserving the signedness, and note the odd effect that you cannot extract a single bit from a signed `int<n>` value

2. Codify a different behavior (and change p4c to match)

    1. bitstring extracts always result in unsigned `bit<k>` values

    2. single bit extracts result in `bit<1>` but multibit extracts from `int<n>` result in `int<k>`

    3. Extracts from `int<n>` result in `int<k>` values iff `m` is the msb and `l < m`, and `bit<k>` values otherwise (k = m - l + 1 as always)

Alternately we could also (or independently) relax the prohibition on `int<1>` -- this type is weird (a bare sign bit), but does have a reasonable meaning (can only contain the values 0 and -1)

cc10512:
From the alternatives you suggest, I prefer 2. Although supporting int<1> sounds quite reasonable, extracting a single bit from an int<n> and ending up with sign bit seems wrong. Getting a bit<1> is much more inline with what's likely to be the programmer's intention.

There is another alternative: prohibit slices on signed int altogether. Likely too big of a hammer, but I thought I would put it out there.

jafingerhut:
No strong opinion here, but out of alternatives mentioned, these sound most reasonable to me:

+ prohibit slices on `int<W>` types completely.  We could point out that `int<W>` can be type cast to `bit<W>` first (which does not change the sequence of bits at all), then bit slicing is available on the result.

+ allow slices of any width, even down to only 1 bit wide, on `int<W>` types, with the result being `int<X>` for X <= W.  Allow `int<1>` type.

mbudiu-vmw:
[...]

PR #701, opened 2018-11-29
Allow `int<1>`: fix for issue #668

mbudiu-vmw:
This allows the type `int<1>` and adds slices for signed values.
@jafingerhut : you have claimed this issue, but I wanted to discuss this next Monday so I took over.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3318,7 +3318,7 @@ type. The result always has the same width as the left operand.
 - Extraction of a set of contiguous bits, also known as a slice, denoted by `[m:l]`,
   where `m` and `l` must be positive integers
   that are compile-time known values, and `m >= l`. The result is
-  a bit-string of width `m - l + 1`, including the bits numbered
+  a signed bit-string of width `m - l + 1`, including the bits numbered
   from `l` (which becomes the least significant bit of the result) to `m` (the
   most significant bit of the result) from the source operand. The conditions `0 <= l < W`
   and `l <= m < W` are checked statically (where `W` is
@@ -3328,7 +3328,7 @@ type. The result always has the same width as the left operand.
   at compile time. Slices are also l-values, which means that P4 supports assigning to a slice: ` e[m:l] = x `.
   The effect of this statement is to set bits `m` to `l` of `e` to the
   bit-pattern represented by `x`, and leaves all other bits of `e`
-  unchanged.  A slice of a signed integer is treated like a signed integer as well.
+  unchanged.  A slice of a signed integer is treated like an unsigned integer.
 
 ### A note about shifts
 
@@ -7125,6 +7125,8 @@ The P4 compiler should provide:
 
 ## Summary of changes made in version 1.2.0
 
+* Allow 1-bit signed values
+* Define the type of bit slices from signed and unsigned values to be unsigned.
 * Constrain `default` label position for `switch` statements.
 * Allow empty tuples.
 * Added `@deprecated` annotation.
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1953,7 +1953,7 @@ described in Section [#sec-bit-ops].
 
 Signed integers are represented using two's complement. An integer with `W`
 bits is declared as: `int<W>`. `W` must be an expression that evaluates to
-a compile-time known value that is a positive integer greater than 1.
+a compile-time known value that is a positive integer.
 
 Bits within an integer are numbered from `0` to `W-1`. Bit `0`
 is the least significant, and bit `W-1` is the sign bit.
@@ -1969,6 +1969,8 @@ bit values).
 
 All operations that can be performed on signed integers are described
 in Section [#sec-int-ops].
+
+A signed integer with width 1 can only have two legal values: 0 and -1.
 
 #### Dynamically-sized bit-strings
 
@@ -2035,7 +2037,7 @@ types. For additional examples of literals see Section
 | `8s10`  | Type is `int<8>`, value is 10 |
 | `2s3`   | Type is `int<2>`, value is -1 (last 2 bits), overflow warning |
 | `1w10`  | Type is `bit<1>`, value is 0 (last bit), overflow warning |
-| `1s10`  | Error: 1-bit signed type is illegal |
+| `1s10`  | Type is `int<1>`, value is -1, overflow warning |
 |----------|--------------------|
 
 ## Derived types { #sec-derived-types }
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -2025,7 +2027,7 @@ types. For additional examples of literals see Section
 | `8s10`  | Type is `int<8>`, value is 10 |
 | `2s3`   | Type is `int<2>`, value is -1 (last 2 bits), overflow warning |
 | `1w10`  | Type is `bit<1>`, value is 0 (last bit), overflow warning |
-| `1s10`  | Error: 1-bit signed type is illegal |
+| `1s10`  | Type is `int<1>`, value is -1, overflow warning |
```
jafingerhut: This seems very odd to me.  If the example were

`1s1` Type is `int<1>`, value is -1

that would make sense to me.

Why should it be allowed if two bits are given after the 's'?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3220,6 +3222,20 @@ type. The result always has the same width as the left operand.
...
+  from `l` (which becomes the least significant bit of the result) to `m` (the
+  most significant bit of the result) from the source operand. The conditions `0 <= l < W`
+  and `l <= m < W` are checked statically (where `W` is
+  the length of the source bit-string). Note that both endpoints of
+  the extraction are inclusive. The bounds are required to be
+  compile-time known values so that the result width can be computed
+  at compile time. Slices are also l-values, which means that P4 supports assigning to a slice: ` e[m:l] = x `.
+  The effect of this statement is to set bits `m` to `l` of `e` to the
+  bit-pattern represented by `x`, and leaves all other bits of `e`
+  unchanged.  A slice of a signed integer is treated like a signed integer as well.
```
jafingerhut: Would it be reasonable to define a slice on type `int<W> e` to be equal to `(int<X>) ( ((bit<X>) e)[m:l] ), where `X=m-l+1` ?

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: I will change this, since it is wrong. But that's more than two bits, it's ten in base ten.

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: that's a way to define the value of this expression, but it does not work as an l-value, since casts are not l-values.

liujed: (Documenting and expanding on what I mentioned in the LDWG meeting on 3 Dec 2018.)

To me, bit-slicing is fundamentally an operation on bit strings, not integers. It seems strange, then, for the result of a bit slice to be signed. I can think of two cases where it makes sense to have a signed result:
1. the bit slice is being used as an r-value, and is truncating an `int<W>`, which is equivalent to casting to a shorter integer type; and
2. the bit slice includes the sign bit, which is equivalent to an arithmetic right shift, followed by a cast.

I'd argue that these use cases are an abuse of bit slicing (programmers should instead be doing the alternative equivalents mentioned above) and that bit slicing should always produce an unsigned result.

jnfoster: Writing without my LDWG co-chair hat on, I endorse this perspective. It is simple and clean. And adding the required casts doesn't seem like it would clutter up programs in practical scenarios.

jafingerhut: I am not arguing _for_ the result of a bit-slice on `int<W>` types to be signed, but note that if you consider cases 1 and 2 to make sense to have signed results, 1 could be considered simply a "synonym" for `my_int[W-1:0]`, and 2 as a synonym for `my_int[width-1:lsb]`.  If you think of the 'extreme cases' as reasonable to have signed results, then I'd say it makes just as sense for all of the intermediate bit slices to have signed results, too.

Again, I am perfectly happy to simply just say that the result of a bit slice on a signed integer is an unsigned result always.  Bits are bits, and none of this wrangling over the types changes the bit patterns of the results at all.

jnfoster:
I can be fine with this.

jafingerhut:
I do not have a strong preference for whether the results are of type `bit<W>` or `int<W>`, but it seems like too many angels dancing on the head of a pin if it were defined as sometimes `bit<W>`, sometimes `int<W>`, depending upon the bit range in the slice operator.  Please let us not do that.

liujed:
I agree with @jafingerhut's sentiment that having the result be sometimes `bit<W>` and sometimes int<W>` will just cause confusion. My preference for `bit<W>` over `int<W>` is a strong one, but I could be convinced of `int<W>` if there is a common use case for wanting `int<W>`.

p4-16/spec/P4-16-spec.mdk
jafingerhut: My understanding is that the LDWG's decision on the type of `a[m:l]` where `a` is a signed integer, is that `a[m:l]` is an unsigned integer.  If that is correct, this PR should be updated to reflect that.

I assume that means that when used as a left value `a[m:l] = expression;` is only legal if expression is an unsigned integer?

jafingerhut: Resolving this Github conversation, since Mihai answered it to my satisfaction.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3313,6 +3315,20 @@ type. The result always has the same width as the left operand.
   - all result bits are zero when shifting left
   - all result bits are zero when shifting right a positive value
   - all result bits are one when shifting right a negative value
+- Extraction of a set of contiguous bits, also known as a slice, denoted by `[m:l]`,
+  where `m` and `l` must be positive integers
+  that are compile-time known values, and `m >= l`. The result is
+  a signed bit-string of width `m - l + 1`, including the bits numbered
```
jafingerhut: Replace 'signed' with 'unsigned' on this line?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3313,6 +3315,20 @@ type. The result always has the same width as the left operand.
...
+  where `m` and `l` must be positive integers
+  that are compile-time known values, and `m >= l`. The result is
+  an unsigned bit-string of width `m - l + 1`, including the bits numbered
+  from `l` (which becomes the least significant bit of the result) to `m` (the
+  most significant bit of the result) from the source operand. The conditions `0 <= l < W`
+  and `l <= m < W` are checked statically (where `W` is
+  the length of the source bit-string). Note that both endpoints of
+  the extraction are inclusive. The bounds are required to be
+  compile-time known values so that the result width can be computed
+  at compile time. Slices are also l-values, which means that P4 supports assigning to a slice: ` e[m:l] = x `.
```
jafingerhut: If the intent is that in the assignment `e[m:l] = expression;` that `expression` must be an unsigned integer type is that worth saying explicitly?  That is, just like the assignment `x = y;` is a compile-time error if x is unsigned and y is signed, so should `e[m:l] = y;` if `e` and `y` are both signed, because `e[m:l]` is an unsigned l-value?

If the answer is yes, then not that this is even a compile-time error, but I don't see a strong reason to include that in the spec.  It probably should be a p4c test case:
```
int<8> signed_int1;
int<8> signed_int2;

signed_int1[7:0] = signed_int2;
```

mbudiu-vmw:
```
./p4test signed.p4 
signed.p4(5): [--Werror=type-error] error: AssignmentStatement: Cannot unify int<8> to bit<8>
        signed_int1[7:0] = signed_int2;
                         ^
```

mbudiu-vmw:
I will then go ahead and merge this change.

PR #703, opened 2018-11-30
Formatting fixes

mbudiu-vmw:
The pdf tables were rendered wrong without these fixes.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -7201,7 +7201,7 @@ above on recursion that apply while a parser or control is executing.
 The first table lists restrictions on what types can be passed as
 constructor parameters to other types.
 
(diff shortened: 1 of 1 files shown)


jnfoster:
How so? I'm not sure what @jafingerhut intended, but this looks okay to me.

p4-spec-table

jafingerhut:
@jnfoster I tried building with the current version, and with Mihai's proposed changes.  The PDF file generated looks the same for both, but the HTML looks better with these changes than without them. With them, the top border line of the table looks normal.  Without them, the top border line is incomplete.

jnfoster:
Ahh, I'd only checked the PDF based on the commit comment.

jafingerhut:
I don't know whether Mihai replaced "HTML" with "PDF" by accident in his comment, or perhaps he is getting different output than we are for the PDF.

Issue #282, opened 2017-06-02
Type of empty list expressions


ChrisDodd:
The P4_16 spec 2017-05-22 has a conflict between the specification of expression list and tuple type.

Secttion 8.11 gives the syntax of expression lists:

expression ...
    | '{' expressionList '}'

expressionList
    : /* empty */
    | expression
    | expressionList ',' expression
    ;

and says "The type of a list expression is a tuple type (Section 7.2.8)."  Note that according to the syntax it is possible to write an empty expression list:

    { }

Section 7.2.6 gives the syntax of a tuple type:

tupleType
    : TUPLE '<' typeArgumentList '>'
    ;

And then section 8 gives the syntax of a type argument list:

typeArgumentList
    : typeArg
    | typeArgumentList ',' typeArg
    ;

Note that according to the syntax it is NOT POSSIBLE to have a tuple type with NO typeArgs.

SO: What is the type of an empty expression list?  According to section 8.11 it is supposed to be a tuple type.  But according to sections 7.2.6 and 8, every tuple type contains one or more type args.  So the spec is in conflict on this point.

mbudiu-vmw:
The spec does not say you cannot have an empty tuple type, it just says you cannot write one in p4. This is no different from the set type.

But indeed it would be cleaner if we allowed empty tuple types. This would also allow declaring variables with an empty tuple type.

mbudiu-vmw:
Do we want to allow a type like `tuple<>`, with no fields?

PR #705, opened 2018-11-30
Fix for issue #282

mbudiu-vmw:


mbudiu-vmw:
As we have agreed during the design meeting, I think we can merge this PR. I have added a change to the changelist.

Issue #331, opened 2017-06-27
'int' type


ChrisDodd:
Section 7.1.6.5 talks about the type `int`

> 7.1.6.5. Infinite-precision integers
>
> The infinite-precision data type describes integers with an unlimited precision. This type is written as int.
>
> This type is reserved for integer literals and expressions that involve only literals. No P4 runtime value > can have an int type; at compile time the compiler will convert all int values that have a runtime > component to fixed-width types, according to the rules described below.

However, the grammar in the appendix does not recognize `int` as a type (only `int<` *integer* `>`).  P4_14 (v1.1) used `int` in declarations of extern functions where the argument needed to be a constant and had no particular bitsize.  Should something similar be allowed in P4_16?  What exactly does it mean -- an argument that must be a compile time constant (generally a literal or expression involving only literals?)

jnfoster:
@mbudiu-vmw should weigh in, but I believe this is intentional. `int` is not a type the programmer can write down, just an internal type used by the compiler. If that understanding is correct, the text should be tweaked to clarify.

We define compile-time known values precisely at the start of Section 16. Would it help to add a link or would that just add clutter?

ChrisDodd:
In that case, we should probably remove "This type is written as int", as that suggests `int` can be used to denote the type.  Its not clear what, if any, connection there is to section 16 compile-time known values.

ChrisDodd:
The place this comes up is things the 'index' arguments to counter/meter methods.  In v1model.p4, these are generally defined as `bit<32>`, which generally means that the user needs to explicitly cast the argument they use to `bit<32>` if it is of some other type (which it often is if it comes from an action parameter).  It would be nice if there was a way to define an extern that "just worked" without the need for extra casts and type specifiers.

tomrodeheffer:
And then there are constant declarations.

A constant declaration does not construct a P4 runtime value, but rather a P4 compile time value.  So there is no reason why you should not be able to write a constant declaration of type `int`, for example:

```P4
const int threshold = 150 + (6 / 2);
```

which would then permit the programmer to use `threshold` everywhere in the P4 program where formerly there occurred the magic constant `153`.

Except that the grammar does not permit it, since you cannot actually write `int` for a baseType.

jnfoster:
That's a good observation. Indeed, it'd be fine to allow `int`s as constant
declarations.

On the other hand, if one wanted to play devil's advocate, one could argue
against allowing this for the sake of uniformity -- in general we've tried
[...]

Issue #383, opened 2017-08-15
P4_16: Allow 'bit-vector structs' to be used everywhere that bit<W> type can


jafingerhut:
Or at least almost everywhere.

By a 'bit-vector struct' is meant the same thing as described in this issue: https://github.com/p4lang/p4-spec/issues/342

Motivation: structs and nested structs can make it noticeably more convenient to write code that wants to bundle together bit<W> fields that occur together into a unit.

The proposal is to allow bit-vector structs to be used in just about any place that a single bit<W> type value is permitted today.  Some of these may already be permitted by the existing P4_16 spec and open source tools, but an explicit mention in the spec might help ensure they gain wide adoption in P4-based systems.

Proposed that bit-vector structs should be supported:
+ as parameters of actions
+ as keys in tables
+ as parameters of extern methods (e.g. data parameter of Register.write)
+ as return values of extern methods (e.g. return value of Register.read)
+ allow them to be cast to the type `bit<W>` where W is the total width of all fields in the bit-vector struct (including nested structs)
+ allow `bit<W>` type to be cast to type of a bit-vector struct with the same total width in bits

mbudiu-vmw:
This is not trivial. It has important consequences for the control-plane api. Also, casting may be very expensive on some architectures (e.g. Software where p4 structs are mapped to c structs which have different allignment rules).

jnfoster:
I don't think there are necessarily complex run-time costs associated with this feature. The compiler can recognize these type equalities but use a canonical representation (e.g., flat bits) internally. (The one source of additional complexity I see is what you point out: ideally P4Runtime would allow any of the bit types representations defined in the program to be used...)

orenbenkiki:
@mbudiu-vmw, can you elaborate on the C `struct` issue?

Some points that come to mind:

1.  Andy's proposal applies only to structures containing only `int<W>` and `bit<W>` (possibly nested) fields, which I expect to be converted to C bit fields in (possibly nested) `struct` with a `__packed__` attribute (or the equivalent in your favorite compiler). Is it an issue to expect such `struct` types to be packed? 

[...]

PR #706, opened 2018-11-30
int type: Fix for issue #331

mbudiu-vmw:
There is a corresponding PR in the compiler: https://github.com/p4lang/p4c/pull/1618

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1773,6 +1773,7 @@ baseType
     : BOOL
     | ERROR
     | BIT
+    | INT
     | BIT '<' INTEGER '>'
     | INT '<' INTEGER '>'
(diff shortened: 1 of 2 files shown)


jafingerhut:
Is there a different issue than #383 that this is intended to be a fix for?  Or is the issue of adding a type `int` discussed somewhere in that long list of comments in #383 and is not easy to find?

cc10512:
Based on the discussion in the LDWG, I would guess that he meant #331

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1773,6 +1773,7 @@ baseType
     : BOOL
     | ERROR
     | BIT
+    | INT
```
jnfoster: Minor: the `int` type should be added to the list "P4 supports the following built-in base types" above...

Issue #710, opened 2018-12-01
P4_16 maybe wrong cross-reference within spec for parsing variable-length headers


jafingerhut:
These comments refer to version 1.1.0 of the P4_16 language specification.

The sentence below can be found near the end of Section 7.2.2 "Header Types":

"As discussed in Section 8.11, headers that contain variable-length fields may need to be parsed in multiple steps by being broken into multiple headers."

Section 8.11 is "Operations on Lists", and does not mention anything about variable-length fields in headers that I can see.

Possible improvements:
+ Remove the sentence.
+ Change cross-reference from 8.11 to instead reference section 8.8 "Operations on variable-size bit types".  The only down side to this is that while Section 8.8 mentions the 2-argument extract method, it doesn't really discuss the topic of "variable-length fields may need to be parsed in multiple steps by being broken into multiple headers".  I don't know anywhere in the spec that discusses this, or if there ever was such a discussion written in earlier versions of the spec.

mbudiu-vmw:
I checked the original Word document and the bug has been in the spec for at least 2 years, inherited from that document, where the reference was already broken. What this means to say is that you can treat a header either as having variable-sized components, or you can write a more detailed parser, using a stack of (unions) of headers to extract various components (e.g., IPv4 options), without using any variable-sized fields.

mbudiu-vmw:
A link to your document https://github.com/jafingerhut/p4-guide/tree/master/variable-length-header would be very appropriate. Or incorporating that document in the spec...

mbudiu-vmw:
Let's change the cross-reference to section 8.8 and make it clear that this is the intended example.
@jafingerhut : will you do this?

jafingerhut:
Yes, I can create a PR for changing the xref to section 8.8.

jafingerhut:
Closing this issue as PR https://github.com/p4lang/p4-spec/pull/711/files has addressed it.

PR #711, opened 2018-12-04
P4_16: Update cross reference to an example of IPv4 option parsing

jafingerhut:


jafingerhut:
This is intended to address https://github.com/p4lang/p4-spec/issues/710 although in a slightly different way than discussed there, because I was confused about which section contained the relevant code example.  It wasn't section 8.8 but later in the doc.  This PR does contain a cross ref to the relevant section, which is Section 12.8.2 "Variable width extraction"

PR #713, opened 2018-12-27
Add @match annotation to list of predefined annotations

jafingerhut:


p4-16/spec/P4-16-spec.mdk
```diff
@@ -5545,7 +5545,7 @@ match_kind {
 ~ End P4Example
 
 These identifiers correspond to the P4~14~ match kinds with the same
-names. The semantics of these annotations is actually not needed to
+names. The semantics of these match kinds is actually not needed to
```
jafingerhut: This is not directly related to adding `@match` as another predefined annotation, but I noticed this potential misuse of the word "annotations" here while looking for all occurrences of the word "annotations" and similar words, and thought I would try improving on it.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6941,6 +6941,13 @@ to two controllable entities, which is illegal.
...
 can be used to enforce the atomic execution of a code block.
 
+### Value set annotations { #sec-value-set-annotations }
+
+The `@match` annotation, described in Section [#sec-select], is used
+to specify a `match_kind` value other than the default `match_kind` of
+`exact` for a field of a `value_set`.
+
+
 ## Package Documentation Annotations { #sec-package-anno }
```
jafingerhut: Question for others: Should this section have `###` instead of `##`, so that it appears as a sub-section inside of the `## Predefined annotations` section?  Right now it is a sibling in the section hierarchy, not a child, which seems a little off to me.

cc10512: I agree it should be a subsection.

jafingerhut: I added another commit that moved section 'Package Documentation Annotations' one level lower in the hierarchy.

antoninbas:
Can we merge this?

mbudiu-vmw:
The merge button does not work the phone for me...

PR #717, opened 2019-01-07
Describe structure initializer expressions

mbudiu-vmw:
The grammar has not been tested.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3816,7 +3816,7 @@ a `struct`---the list fields are assigned to the header fields in the
 order they appear.  In this case the header automatically becomes
 valid:
 
-Similar to a `struct, a header object can be initialized with a list
+Similar to a `struct`, a header object can be initialized with a list
 expression [#sec-list-exprs] --- the list fields are assigned to the
 header fields in the order they appear, or with a struct initializer
 expression [#sec-ops-on-structs].  When initialized the header
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2859,7 +2859,6 @@ expression
     | expression '[' expression ':' expression ']'
     | '{' expressionList '}'
     | '{' kvList '}'
-    | typeName '{' kvList '}'
     | '(' expression ')'
     | '!' expression
     | '~' expression
@@ -3743,8 +3742,8 @@ field access, written using dot (".") notation---e.g., `s.field`. If
 copying `struct`s using assignment when the source and target of the
 assignment have the same type. Finally, `struct`s can be initialized
 with a list expression, as discussed in Section [#sec-list-exprs], or
-with a structure-valued expression, as described in
-[#sec-structure-expressions].  Both these cases must initialize all
+with a structure initializer, as described in
+[#sec-structure-initializers].  Both these cases must initialize all
 fields of the structure.
 
 Two structs can be compared for equality (==) or inequality (!=) only
@@ -3752,11 +3751,11 @@ if they have the same type and all of their fields can be recursively
 compared for equality.  Two structures are equal if and only if all
 their corresponding fields are equal.
 
-## Structure initializers { #sec-structure-expressions}
+## Structure initializers { #sec-structure-initializers }
 
 Structures can be initialized using structure initializers, which
 specify explicitly the intialized fields.  A structure initializer
-expression evaluates to an struct; it can be used on the right-hand
+expression evaluates to a struct; it can be used on the right-hand
 side of an initialization of a variable with a struct type.
 
 ~ Begin P4Grammar
@@ -3775,7 +3774,7 @@ kvPair
 ~ End P4Grammar
 
 The following example shows a structure initialized using a
-structure-valued expression:
+structure initializer:
 
 ~ Begin P4Example
 struct S {
@@ -3822,7 +3821,7 @@ valid:
 
 Similar to a `struct`, a header object can be initialized with a list
 expression [#sec-list-exprs] --- the list fields are assigned to the
-header fields in the order they appear, or with a struct initializer
+header fields in the order they appear --- or with a structure initializer
 expression [#sec-ops-on-structs].  When initialized the header
 automatically becomes valid:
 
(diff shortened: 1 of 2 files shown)


jafingerhut:
This idea seems sound to me, and a worthwhile convenience for P4 programmers to have in the language.

mbudiu-vmw:
The only thing that may be worth adding is allowing struct initializers to be left-values in some circumstances. That's a more tricky case, we can do it in a separate PR if we do it at all. The same thing holds for list expressions, BTW.

hanw:
FYI, I found StructInitializerExpression useful to approximate P4-14 field list during the P4-14 to 16 translation for primitives that accepts a field list.

For example,
```
field_list { hdr.f0, hdr.f1 } fl;
clone3(id, fl);
```

is translated to
```
header fl_t {
   bit<8> f0;
   bit<8> f1;
}
fl_t fl;
fl = { hdr.f0, hdr.f1 };
mirror.emit(id, fl);
```

The nice thing about StructInitializerExpression is it carries information about the original struct/header type, which is not available in ListExpression (the type for ListExpression is always a tuple<>).

hanw:
We used to convert FieldList directly to ListExpression during P4-14 to 16 translation, which loses information.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3703,10 +3737,17 @@ a `struct`---the list fields are assigned to the header fields in the
 order they appear.  In this case the header automatically becomes
 valid:
 
+Similar to a `struct, a header object can be initialized with a list
```
jnfoster: The "`" before `struct` is not closed.

mbudiu-vmw:
```
struct S {
   bit<32> x;
   bit<32> y;
}

struct T {
    S s;
}

T t = { s = { x = 0, y = 1 } };
```

mbudiu-vmw:
```
S s = _;
```

mbudiu-vmw:
```
S = { x = 0, ... };
```

mbudiu-vmw:
```
(S){ x = 0, y = 1 };
```

mbudiu-vmw:
```
{ x = 1 }.x;
```

jafingerhut:
Without a type name, these struct intializer expressions can have one of several possible types, and in some programs it might not be possible to narrow it down to only one.

One possibility would be to _require_ a type name as part of the initializer expression.

liujed:
I think we should distinguish between struct initializers and literal struct expressions. Requiring types for struct initializers seems redundant always, but I think we ought to always require types for literal struct expressions. I think my preference would be for something like `S_t { x = 0, y = 1 }` over cast-style syntax (`(S_t) { x = 0, y = 1}`).

mbudiu-vmw:
To revise the grammar to allow only initializers.

hanw:
Compiler currently cannot compile the following code.
```
header mirror_metadata_h {
    bit<48> neigh_mac_hi;
    bit<1> l2_egress_lkp_flag;
}
control SwitchIngressDeparser(packet_out pkt, inout headers_t hdr, in user_metadata_t meta) {
    Mirror() mirror;
    apply {
        mirror.emit<mirror_metadata_h>( meta.session_id, { neigh_mac_hi=meta.srcAddr, l2_egress_lkp_flag=meta.lkp_flag2 });
    }
}
```
```
syntax error, unexpected =, expecting } or ","
     mirror.emit<mirror_metadata_h>( meta.session_id, (mirror_metadata_h){ neigh_mac_hi=
```

will spec allow this use case?

mbudiu-vmw:
I think it would. But we agreed we'll probably start with just initializers and add struct-valued expressions later.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2856,6 +2856,8 @@ expression
     | expression '[' expression ']'
     | expression '[' expression ':' expression ']'
     | '{' expressionList '}'
+    | '{' kvList '}'
+    | typeName '{' kvList '}'
```
mbudiu-vmw: Remove this from PR.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3724,13 +3726,49 @@ field access, written using dot (".") notation---e.g., `s.field`. If
...
 Two structs can be compared for equality (==) or inequality (!=) only
 if they have the same type and all of their fields can be recursively
 compared for equality.  Two structures are equal if and only if all
 their corresponding fields are equal.
 
+## Structure initializers { #sec-structure-expressions}
+
+Structures can be initialized using structure initializers, which
+specify explicitly the intialized fields.  A structure initializer
+expression evaluates to an struct; it can be used on the right-hand
```
liujed: evaluates to *a* struct

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3724,13 +3726,49 @@ field access, written using dot (".") notation---e.g., `s.field`. If
 `s` is an l-value, then `s.field` is also an l-value. P4 also allows
 copying `struct`s using assignment when the source and target of the
 assignment have the same type. Finally, `struct`s can be initialized
-with a list expression, as discussed in Section [#sec-list-exprs].
+with a list expression, as discussed in Section [#sec-list-exprs], or
+with a structure-valued expression, as described in
```
liujed: Replace "structure-valued expression" with "structure initializer".

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3724,13 +3726,49 @@ field access, written using dot (".") notation---e.g., `s.field`. If
...
+    | kvList "," kvPair
+    ;
+
+kvPair
+    : name "=" expression
+    ;
+~ End P4Grammar
+
+The following example shows a structure initialized using a
+structure-valued expression:
```
liujed: Replace "structure-valued expression" with "structure initializer".

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3766,10 +3804,17 @@ a `struct`---the list fields are assigned to the header fields in the
 order they appear.  In this case the header automatically becomes
 valid:
 
+Similar to a `struct`, a header object can be initialized with a list
+expression [#sec-list-exprs] --- the list fields are assigned to the
+header fields in the order they appear, or with a struct initializer
```
liujed: Should the comma on this line be replaced with an em dash to close off the dash on the previous line?

"Structure initializer" or "structure initializer expression" instead of "struct initializer"?

mbudiu-vmw:
I think we agreed to merge this as well for 1.2.0

Issue #342, opened 2017-07-02
P4_16: Allow headers to contain 'bit-vector structs'


jafingerhut:
Feel free to mark this issue "Post P4_16" if it helps.  I just want to get the issue out there right now.

Define a 'bit-vector struct' to be one that contains only these kinds of fields:
+ fields of type `bit<W>` or `int<W>`
+ fields of a struct type that is also a 'bit-vector struct'

Such a 'bit-vector struct' has a straightforward way to define how extract() and emit() methods would work on them, by following the order that fields are defined in the struct.

The proposal is to extend 'header' types to allow them to have 'bit-vector struct' types as members.

Motivation: This can be a programmer convenience to group multiple consecutive related header fields together, passing them as arguments as a unit, making them control plane action parameters as a unit rather than as individual fields, etc.  It does not provide any additional _capability_ over what can do with the P4_16 v1.0.0 language -- just a shorter way to express it in some cases.

There is no need to introduce any new keyword or syntax for what I call 'bit-vector struct' here, and there may be a better name for that concept.  The compiler would simply check that a struct that one tried to make a member of a header was a bit-vector struct, or not, and give an error if it was not.

jnfoster:
Super nice!

Any volunteers to prototype it as an extension in the p4c front-end?

familysingh:
Andy, 
I use familysingh with github,  but my work email is hemant@mnkcg.com.  This issue came to my attention yesterday.  The problem is a show stopper because there will always be a need to save instructions in switching ASIC or some switch code may not compile.  Using a bit-vector struct in P4 header saves instructions.  **The frontend code change for this issue is straightforward**.   However, even if we fix this issue, P4 code won't compile in certain common situations unless Issue# 383 is also resolved.  Switching code can use a search key which comprises only of bit fields.  One has to assign a header's struct fields to the search key.  For example the search key uses 32 bits for dst ip.  But the struct has lower28 and upper4 bits.  We are back to square one.  

Hemant
http://mnkcg.com

hesingh:
Here is a note for specific changes to the P4-16, version 1.1.0 specification.

In section 7.2.7, the table incorporates the following change.
[...]

PR #719, opened 2019-01-08
Change P4-16 spec for nested bit-vector struct

hesingh:


jafingerhut:
Those tables are intended to be summaries of what is allowed, or not allowed, in other places throughout the specification. It would be best to find out where else in the spec that it is stated that headers are only allowed to have `bit<W>`, `int<W>`, and `varbit<W>` field types, and generalize those, too.

jafingerhut:
Hmm.  I was thinking of different tales, in an appendix.  This table you are modifying is much earlier, closer to the definition of headers, so perhaps my comment should be ignored.

It does seem worth perhaps having one example of such a header containing a struct inside of it in Section 7.2.2 "Header Types", explaining that when a struct is inside of a header, the order of the fields for the purposes of `extract` and `emit` calls is the order of the fields as defined in the source code.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2420,12 +2420,15 @@ infinite-precision integer, without a width specified.
...
 | `header_union` | error     | error   |  allowed |
-| `struct`       | error     | error   |  allowed |
+| `struct`       | allowed[^struct_header]   | error   |  allowed |
 | `tuple`        | error     | error   |  allowed |
 |----------------|-----------|---------|----------|
 
 [^enum_header]: an `enum` type used as a field in a `header` must specify a
     underlying type and representation for `enum` elements.
+[^struct_header]: a `struct` type used as a field in a `header` must contain
+    only a bit-field, an integer, a Serializable Enums, or a Boolean.
```
jafingerhut: Do you mean values of type `bit<W>` when you say "bit-field", and values of type `int<W> when you say "integer"?  If so, I would recommend using `bit<W>` and `int<W>` instead of those words, because at least the term "integer" could be confused to also include unlimited precision integers.

hesingh: Will make the change to bit<W> and int<W>

hesingh:
> Hmm. I was thinking of different tales, in an appendix. This table you are modifying is much earlier, closer to the definition of headers, so perhaps my comment should be ignored.
> 
> It does seem worth perhaps having one example of such a header containing a struct inside of it in Section 7.2.2 "Header Types", explaining that when a struct is inside of a header, the order of the fields for the purposes of `extract` and `emit` calls is the order of the fields as defined in the source code.

Sure, I will add the example.

hesingh:
Checked in a new version with some comments addressed.

jafingerhut:
Sorry, I haven't been following your p4c changes in detail so do not know this already, but do your changes enable a `header` to contain a `struct` that itself contains other `struct` fields, nested arbitrarily, as long as all of the "leaf" types are `bit<W>`, `int<W>`, `bool`, or serializable `enum`?

If nested structs are enabled, then it would be good to say so in the spec.

hesingh:
No problem.  Yes, the p4c changes do support arbitrarily-deep nested bit-vector struct inside a P4 Header.  I can elaborate such a fact (with an example) in the p4 spec when I get a chance once the code fires are out.

hesingh:
Thanks for the text suggestion.  I added explicit text to say nested structs, nested arbitrarily, are supported.  See section 7.2.2.  I also added you as a Collaborator to this PR.

For p4c code changes, the PR only supports bit<W> in nested struct.  int<W>, bool, and SerEnum are punted to a subsequent PR.  More test cases and some thought is needed to support these other types.

hesingh:
What do we do with this text in the P4-16 spec now that a header allows a boolean?  The boolean may cause the header size to not be an integer number of bytes.   This fact may be obvious and we can choose to not change any padding text in the spec.  Or do we change the padding text such as the one below?

"There is no padding or alignment of the header fields. Architectures may impose additional constraints on header types—e.g., restricting headers to sizes that are an integer number of bytes. "

jafingerhut:
If booleans are allowed in headers, I would have guessed that they would be explicitly defined to auto-convert to `bit<1>` for emit operations, and from `bit<1>` for extract operations.  Is that the thinking in allowing them in headers?  If not, what is the idea there?

hesingh:
Yes, a boolean is not a problem for emit or extract functionality.  The idea to allow a boolean in a header came about three days back in discussion of Issue 342. 

https://github.com/p4lang/p4-spec/issues/342

You see, the bmv2 backend does not support nested structs.  So the nested structs PR developed developed a midend pass (FlattenHeaders) to flatten struct in a header.  A boolean is legal inside a struct.  So, soon as a struct with a boolean is included in a P4 header, when the struct is flattened by the midend pass, the boolean ends up inside the flattened header.  Currently, P4-16 does not allow boolean in a header.  Mihai voted to change the spec to allow boolean in a header.

I changed this spec PR to allow a boolean inside a header.   I was only wondering if the spec should add additional text related to padding?  For example, a header has size equal to integer number of bytes.  Then, one adds a boolean to the header.  The size changes to be not a multiple of integer number of bytes.   For such a case, it seems obvious, the programmer adds padding bits so that the size remains a multiple of integer number of bytes.   I think, the spec can be silent on padding when the spec adds "allow boolean in a header."

jafingerhut:
The P4_16 spec does _not_ mandate that all P4_16 implementations limit headers to be multiples of 8 bits long.

It _allows particular implementations_ to limit headers to be multiples of 8 bits long.

Unless I am misunderstanding you, if we allow booleans in headers, we should mandate their encoding and lengths, e.g. to be 1 bit long, for all implementations.

Then if the P4_16 language spec continues to allow some implementations to restrict header lengths to multiples of 8 bits, but allows implementations with no such restriction, is there any issue that remains there?

hesingh:
No issue exists after I read your reply - thanks.
The bmv2 backend does use one bit to represent a boolean.  I can add your proposed text to the spec for single bit.

jnfoster:
It would be nice to explain the philosophical reasons why headers can't be nested within headers.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2208,9 +2208,12 @@ structField
 ~ End P4Grammar
 
 where each `typeRef` is restricted to a bit-string type (fixed or
-variable) or an integer type. This declaration introduces a new
-identifier in the current scope; the type can be referred to using
-this identifier. A header is similar to a `struct` in C,
+variable), an integer type, a boolean type, or a struct that itself contains other
```
mbudiu-vmw: make it clear that 'integer type' is actually a fixed-width integer

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2415,17 +2442,20 @@ infinite-precision integer, without a width specified.
...
 | header stack   | error     | error   |  allowed |
 | `header_union` | error     | error   |  allowed |
-| `struct`       | error     | error   |  allowed |
+| `struct`       | allowed[^struct_header]   | error   |  allowed |
 | `tuple`        | error     | error   |  allowed |
 |----------------|-----------|---------|----------|
 
 [^enum_header]: an `enum` type used as a field in a `header` must specify a
     underlying type and representation for `enum` elements.
+[^struct_header]: a `struct` type used as a field in a `header` must contain
```
mbudiu-vmw: or a nested struct that has the same properties.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2208,9 +2208,12 @@ structField
...
 where each `typeRef` is restricted to a bit-string type (fixed or
-variable) or an integer type. This declaration introduces a new
-identifier in the current scope; the type can be referred to using
-this identifier. A header is similar to a `struct` in C,
+variable), an integer type, a boolean type, or a struct that itself contains other 
+struct fields, nested arbitrarily, as long as all of the "leaf" types are
+are `bit<W>`, `int<W>`, a Serializable Enum, or a `bool`.  If a `bool` is used
+inside a P4 header, all implementations encode the `bool` as a one bit long field. 
+This declaration introduces a new identifier in the current scope; the 
+type can be referred to using this identifier. A header is similar to a `struct` in C,
```
mbudiu-vmw: Nate was asking for emphasizing that nested headers are not supported - if this isn't in the text already.

hesingh: The text does not exist in the spec.  I will add it.

p4-16/spec/P4-16-spec.mdk
hesingh: Will make the change.

p4-16/spec/P4-16-spec.mdk
hesingh: Will do.

hesingh:
> It would be nice to explain the philosophical reasons why headers can't be nested within headers.

Will do.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6137,8 +6167,8 @@ header, header stack, `struct`, or header union to the output packet.
 - When applied to a `struct` or header union, `emit` recursively
   invokes itself to each field.
 
-It is illegal to invoke `emit` on an expression of whose type is a
-base type, `enum`, or `error`.
+It is illegal to invoke `emit` on an expression whose type is
+`enum`, `error`, or a base type other than an integer (`int<W>`) or Boolean type.
```
jafingerhut: Is the intent here to change the spec so that `emit` is now explicitly _allowed_ if the expression has type `bit<W>`, `int<W>`, a serializable enum, boolean?  That is, emit directly on a value of those types, even if they are not inside of a header, or a struct inside of a header?

If so, I think it should be stated more clearly, and I can help with wording suggestions.  But first I wanted to check whether that was the intent.

hesingh: Good point.

The current spec does not allow any base type, enum, or error as an argument to the emit() method.   However, since an ```int<W>```  and ```bool``` can be serialized, one could allow them as arguments to the emit method.  But what would the use case be? 

For the other point you raised, indeed, it makes sense to clarify that when a struct is an argument to emit, the struct can only include certain types which are serializable fields.

hesingh:
> It would be nice to explain the philosophical reasons why headers can't be nested within headers.

Done.  I checked in new text.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Before your changes, the spec explicitly said that emit was allowed on headers.  Your changes could be viewed as generalizing what can be _inside_ of a header, while still allowing emit (and extract) to work on them.

If so, and if there are no immediately plans to generalize emit to work on new types, is there any reason to change this sentence at all?

hesingh: Before the nested struct code changes, a header only allowed bit-fields and fixed-width signed integers.  With the changes, a nested struct is allowed, but the struct can include a kitchen sink of types some of which are enum and error which a header cannot support.   Thus, I thought a clarification would help.  However, the clarification may not be in the illegal sentence of section 15.1 - we could make the clarification elsewhere.  Or we could skip any clarifications altogether because it may be obvious that an error and enum are not serializable.

jafingerhut: Maybe I am reading it incorrectly, but when it says that it is illegal to invoke emit on an expression whose type is X, Y, or Z, I read that as the type of the entire expression cannot be X, Y, or Z.

It does not specify anything about whether the expression might contain inside of it, as part of its definition, a field with type X, Y, or Z.

jafingerhut: An alternate suggested change in the text, which is intended to mean exactly what it does now:

"It is illegal to invoke `emit` on an expression whose type is not one of those listed above."

The list above already explicitly says you can do `emit` on a header.  The changes you are making earlier in the spec explicitly say that a header can contain `bool`, `int<W>`, etc.

hesingh:
I realized today that the P4Runtime spec also uses the same table from the P4-16 spec related to 'Type Nesting Rules'.  See Table 6 in the P4Runtime spec, rc4.  Since the P4-16 spec has changed the table, the P4Runtime spec also needs to change the table.  Or the P4Runtime spec should just reference the table from the P4-16 spec.

jafingerhut:
I will let @antoninbas comment on the idea of the P4Runtime spec using that table by reference, rather than by copying.  Even if it were by reference, I think it would have to be by reference _to a particular version of the P4_16 language spec_.

If you tried instead to have a reference in the P4Runtime spec to "whatever is, at this moment, the latest version of the P4_16 language spec", it seems to me that the effect would be to prevent any changes to the P4_16 language spec, forever, that would be incompatible with the 1.0 version of the P4Runtime spec.

hesingh:
If the PI spec is changed to use a reference to the P4-16 spec, then this is the proposed text change. 

PI spec old:
"Table 6 shows the different P416 types and how they are allowed to be used, as per the 
P416 specification."

PI spec new:
"The table in section 7.2.7 of the P416 specification shows the different P416 types and how they are allowed to be used."

jafingerhut:
I do not see how to leave a comment on Github's web page for this PR that looks like a direct reply to @hesingh's comment that I am responding to.

I do not know how I did it, but in the list of types explicitly allowed for an `emit` call, I missed the `struct` mention.  Given that, having the explicit list of types that are illegal to emit, including `enum`, `error` and "base types" (which are listd in their own section of the spec), seems reasonable for now.

If someone later (perhaps soon) wants to propose extending the kinds of structs that can legally be emitted, that would lead to further editing of that text, but save that editing for when/if emittable structs are generalized.

hesingh:
> I do not see how to leave a comment on Github's web page for this PR that looks like a direct reply to @hesingh's comment that I am responding to.
> 
> I do not know how I did it, but in the list of types explicitly allowed for an `emit` call, I missed the `struct` mention. Given that, having the explicit list of types that are illegal to emit, including `enum`, `error` and "base types" (which are listed in their own section of the spec), seems reasonable for now.
> 
> If someone later (perhaps soon) wants to propose extending the kinds of structs that can legally be emitted, that would lead to further editing of that text, but save that editing for when/if emittable structs are generalized.

Ok, thanks for the feedback.  As of yesterday, the p4c implementation matches the changes to the spec being made in this PR.  Good work by our P4 community.

hesingh:
Folks,

For the upcoming P4Lang meeting on Monday, 02/04/2019, please review the spec changes.  Let's try and close this PR soon and then anyone who has merge authorization, can please merge the changes with the repo.  thanks.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2208,16 +2208,26 @@ structField
...
 its "validity" bit is automatically set to `false`. The "validity"
 bit can be manipulated by using the header methods `isValid()`, `setValid()`,
 and `setInvalid()`, as described in Section [#sec-ops-on-hdrs].
 
+Note, nesting of headers is not supported.  The reason is if one used an 
+extract() on a header that contained nested headers, then all of the headers 
+involved would become valid.  Additionally, what about invoking `hdr.setInvalid()` on a 
+"sub-header", without doing so on the top level header? What if one tried to then emit() 
+the top level header in that state? What should be emitted?  Due to such reasons, nested
+headers are not supported.
```
jafingerhut: Sorry for so many comments here, but wanted to suggest some alternate phrasing for this paragraph.  I do not think there is any technically wrong with it -- it is simply worded in a way that sounds a bit more like giving a talk, or a conversation, versus somewhat more dry formal technical presentation that one typically sees in language specifications :-)  A suggested alternative is given below:

"Nesting of headers is not supported.  One reason is that headers containing structs covers many of the use cases where headers containing other headers might be useful.  Another is that it leads to complications in defining the behavior of arbitrary sequences of `setValid`, `setInvalid`, and `emit` operations.  Consider an example where header `h1` contains header `h2` as a member, both currently valid.  A program executes `h2.setInvalid()` followed by `packet.emit(h1)`.  Should all fields of `h1` be emitted, but skipping `h2`?  Similarly, should `h1.setInvalid()` invalidate all headers contained within `h1`, regardless of how deeply they are nested?"

hesingh: Thanks for the review - great feedback!  I will take your proposed text and update the doc.

hesingh:
Regarding, how a nested struct inside a header is emitted, if a backend includes the midend FlattenHeaders pass, emit sees a header with only Type_Bits.  Only, if a backend uses the header without any flattening, then we specify text that the struct has no padding or alignment. Emit would recursively emit the individual bit-fields of any struct.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2208,16 +2208,21 @@ structField
 ~ End P4Grammar
 
 where each `typeRef` is restricted to a bit-string type (fixed or
-variable) or an integer type. This declaration introduces a new
-identifier in the current scope; the type can be referred to using
-this identifier. A header is similar to a `struct` in C,
+variable), a fixed-width signed integer type, a boolean type, or a struct that itself contains other 
+struct fields, nested arbitrarily, as long as all of the "leaf" types are
+are `bit<W>`, `int<W>`, a Serializable Enum, or a `bool`.  If a `bool` is used
```
jafingerhut: Minor recommendation: Replace "Serializable Enum" with "serializable `enum`".  No reason to capitalize it that I can see.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2415,17 +2444,21 @@ infinite-precision integer, without a width specified.
...
 | header stack   | error     | error   |  allowed |
 | `header_union` | error     | error   |  allowed |
-| `struct`       | error     | error   |  allowed |
+| `struct`       | allowed[^struct_header]   | error   |  allowed |
 | `tuple`        | error     | error   |  allowed |
 |----------------|-----------|---------|----------|
 
 [^enum_header]: an `enum` type used as a field in a `header` must specify a
     underlying type and representation for `enum` elements.
+[^struct_header]: a `struct` or nested 'struct' type that has the same properties,
```
jafingerhut: Recommend replacing single quotes around the second occurrence of struct with backquotes, so it is like the first occurrence on the same line.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2415,17 +2444,21 @@ infinite-precision integer, without a width specified.
...
-| `struct`       | error     | error   |  allowed |
+| `struct`       | allowed[^struct_header]   | error   |  allowed |
 | `tuple`        | error     | error   |  allowed |
 |----------------|-----------|---------|----------|
 
 [^enum_header]: an `enum` type used as a field in a `header` must specify a
     underlying type and representation for `enum` elements.
+[^struct_header]: a `struct` or nested 'struct' type that has the same properties, 
+    used as a field in a `header` must contain only 
+    `bit<W>`, `int<W>`, a Serializable Enum, or a `bool`.
```
jafingerhut: As above, recommend replacing "Serializable Enum" with "serializable `enum`"

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2208,16 +2208,21 @@ structField
...
 
 where each `typeRef` is restricted to a bit-string type (fixed or
-variable) or an integer type. This declaration introduces a new
-identifier in the current scope; the type can be referred to using
-this identifier. A header is similar to a `struct` in C,
+variable), a fixed-width signed integer type, a boolean type, or a struct that itself contains other 
+struct fields, nested arbitrarily, as long as all of the "leaf" types are
+are `bit<W>`, `int<W>`, a Serializable Enum, or a `bool`.  If a `bool` is used
+inside a P4 header, all implementations encode the `bool` as a one bit long field. 
+This declaration introduces a new identifier in the current scope; the
```
hanw: Please be more precise on what is `This declaration`, what is `current scope`, and what is `this identifier`.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2208,16 +2208,21 @@ structField
...
+type can be referred to using this identifier. A header is similar to a `struct` in C,
 containing all the specified fields. However, in addition, a header also
 contains a hidden Boolean "validity" field. When the "validity" bit is `true`
 we say that the "header is valid". When a header is created
 its "validity" bit is automatically set to `false`. The "validity"
 bit can be manipulated by using the header methods `isValid()`, `setValid()`,
 and `setInvalid()`, as described in Section [#sec-ops-on-hdrs].
 
+Note, nesting of headers is not supported.  One reason is that headers containing structs covers many of the use cases where headers containing other headers might be useful. Another is that it leads to complications in defining the behavior of arbitrary sequences of `setValid`, `setInvalid`, and `emit` operations. Consider an example where header `h1` contains header `h2` as a member, both currently valid. A program executes `h2.setInvalid()` followed by `packet.emit(h1)`. Should all fields of `h1` be emitted, but skipping `h2`? Similarly, should `h1.setInvalid()` invalidate all headers contained within `h1`, regardless of how deeply they are nested?
```
hanw: If you are editing with emacs, it make sense to wrap the line around column 80 or 100 for easier diff in the future.

mbudiu-vmw:
@jafingerhut : I can see the text for emit, but not for extract.

p4-16/spec/P4-16-spec.mdk
hanw: `One reason is that headers containing structs covers many of the use cases where headers containing other headers might be useful.` I don't think this can be a reason for not supporting nested header.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2226,6 +2231,30 @@ header Empty_h { }
 
 Note that an empty header still contains a validity bit.
 
+When a struct is inside of a header, the order of the fields for the purposes 
+of extract and emit calls is the order of the fields as defined in the source code.
```
jafingerhut: @mbudiu-vmw Both extract and emit are mentioned here.  Is this enough?

@hesingh As a typographic nicety, it might be nice to have them in backquotes so they show up as a different font like other identifiers and keywords in the spec.

p4-16/spec/P4-16-spec.mdk
hesingh: Please see the original p4-16 spec at the section at this URL

https://p4.org/p4-spec/docs/P4-16-v1.1.0-spec.html#sec-header-types

The declaration is more clear there.  In this PR, there is more text added and thus the "declaration" , current scope and this identifier is obscured.  I will have to think about this one,

hesingh:
Thanks Andy, Han, and Mihai for your review today.  I have checked in a new version.

jafingerhut:
@mbudiu-vmw @hanw As far as I can tell, the latest version of this PR addresses all of your comments.  Do you think any comments remain unaddressed?

mbudiu-vmw:
I think we agreed to merge this.

hesingh:
Thanks, @mbudiu-vmw

PR #724, opened 2019-02-04
P4₁₆: Relax the structure of annotation bodies

liujed:


liujed:
@jnfoster: I think we can merge this.

PR #727, opened 2019-02-15
Add psa-example-hello-world.p4 example program

jafingerhut:


jnfoster:
I never thought I'd say this, but it almost makes you miss the brevity and elegance of "Hello World!" in Java :-P

cc10512:
That should be our goal for the next release of the language!

p4-16/psa/examples/psa-example-hello-world.p4
```diff
@@ -0,0 +1,176 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <core.p4>
+#include "../psa.p4"
```
hanw: Why do we need `..` in the path?

p4-16/psa/examples/psa-example-hello-world.p4
```diff
@@ -0,0 +1,176 @@
...
+                in    psa_ingress_input_metadata_t  istd,
+                inout psa_ingress_output_metadata_t ostd)
+{
+    apply {
+        if (hdr.ipv4.isValid()) {
+            // Direct IPv4 packets out of ports 0 through 3, based
+            // upon the 2 least significant bits of the IPv4
+            // destination address.
+            send_to_port(ostd,
+                (PortId_t) ((PortIdUint_t) hdr.ipv4.dstAddr[1:0]));
```
hanw: The latest master fails to parse this line? 
```
hanwang-lt1:build hanwang$ ./p4c-bm2-psa --std p4-16 psa-example-hello-world.p4 -Tpass_manager:1
psa-example-hello-world.p4(93):syntax error, unexpected IDENTIFIER "hdr"
                (PortId_t) ((PortIdUint_t) hdr
                                           ^^^
error: 1 errors encountered, aborting compilation
```

p4-16/psa/examples/psa-example-hello-world.p4
jafingerhut: All P4_16 files in the p4-16/psa/examples have this, to force them to include the psa.p4 file in this repository.  At least when we started, there was no psa.p4 file in p4c at all.

If we want to make future changes to psa.p4, having this ../psa.p4 for these example files ensures that 'make check' will use the locally modified psa.p4, not the sure-to-be-older one in p4c.

p4-16/psa/examples/psa-example-hello-world.p4
jafingerhut: Do you see anything wrong with the line of code?  If not, it seems like it must be a bug in p4c-bm2-psa to me.

jafingerhut: The latest p4test parses it without errors.

hanw: I don't see anything wrong in the source. must be a bug in the frontend.

hanw: ah, it is because I do not have the updated psa.p4.

jafingerhut:
Thanks for quick review.  I am going ahead and merging this.  We can fix any issues that folks find with it in future commits, if this program has bugs.

PR #728, opened 2019-02-16
Update psa.p4 include file to have example widths for BMv2

jafingerhut:
This will make the file match what we plan to have in the p4c
implementation, to make it easier to keep these files synchronized in
the future.


jafingerhut:
It seems best for reviewing this PR to wait until we can agree on this PR, and then make the version of psa.p4 committed with this PR match that one: https://github.com/p4lang/p4c/pull/1730

PR #731, opened 2019-02-21
Add a table listing which types upon which you can do typedef and type

jafingerhut:
The contents of the table are based upon the current p4c
implementation, tested via the p4test command.


mbudiu-vmw:
I think we should just merge this.

jafingerhut:
Recent changes in p4c have made it no longer an error to have a definition of `typedef int foo;`, so we could perhaps consider changing that entry of the table from error to allowed.

There was a question raised at the LDWG of whether `type tuple ... foo;` should be allowed or not.  I do not know if there is any urgent use case for supporting that, personally.  I do know that the PSA's use of `type` definitions only needs support for `type bit<W> foo;` and `type A B;`, where A is `typedef bit<W> A;`.  Anything over and above that is harmless to PSA, but currently unnecessary.

Issue #730, opened 2019-02-17
Should all 7 'special' PSA types have `@p4runtime_translation` annotation?


jafingerhut:
Sorry I did not notice this earlier, but the spec and the p4type.info file both say that if a `type` defined in a P4_16 source file has no `@p4runtime_translation` annotation, then the p4info file should have `original_type` information that comes from the P4_16 source code.

Currently the psa.p4 file only has `@p4runtime_translation` annotations on two types: `PortId_t` and `ClassOfService_t`: https://github.com/p4lang/p4c/blob/master/p4include/psa.p4#L77-L85

My belief is that all PSA implementations are expected to modify their target-specific psa.p4 #include file, at least slightly, e.g. by using their target's custom bit widths for these 7 PSA types: https://github.com/p4lang/p4c/blob/master/p4include/psa.p4#L60-L66

According to P4Runtime v1.0, that implies that the other 5 types, e.g. `MulticastGroup_t`, will have their platform-specific bit widths in the P4Info file, and thus the P4Info file generated by different target-specific compilers would differ from each other in the contents of their original_type messages for the same P4 PSA source program.

It seems like there are at least the following ways to improve on this situation:

(a) Change psa.p4 so that all 7 PSA types expected to have custom bit widths per target, have a `@p4runtime_translation` annotation on them.  According to the existing P4Runtime v1.0 spec, they will thus _not_ have their target-specific bit widths present in the P4Info files, only the `translated_type` P4Info message.

(b) Change the P4Runtime spec to say that for each architecture, there can be types that do not have a `@p4runtime_translation` annotation on them, and yet they still have the `translated_type` message anyway, and no `original_type` message.

There are probably other choices, too, but I am not thinking of others right now.

Choice (a) seems like a reasonable way forward to me.  Thoughts?

jafingerhut:
Created issue #732, reviewed, and now it has been merged into PSA spec.

PR #732, opened 2019-02-21
PSA - Add @p4runtime_translation for all seven PSA types

jafingerhut:


jafingerhut:
Intended to address issue https://github.com/p4lang/p4-spec/issues/730

p4-16/psa/PSA.mdk
```diff
@@ -512,22 +511,15 @@ translation in the agent software, between target-specific values, and
...
-access values in these externs, the numerical translation may not be
-performed on the index.  It is reasonable to use a `DirectCounter` or
-`DirectMeter` with a table key that includes a `PortId_t` field,
-because the agent software will perform port translation for table
-keys.
-
-If this is not a reasonable thing to do, the example program
-`psa-example-counters.p4` should be modified so that it does not do
-this, and scrub all other examples for any similar behavior.
-~
```
jafingerhut: Removing this TBD could be done in a separate PR if you wish.  Antonin addressed this issue in P4Runtime v1.0 in Jan 2019.

p4-16/psa/PSA.mdk
```diff
@@ -477,8 +477,7 @@ The list above is not intended to be exhaustive.
 All of the comments above apply to all types where numerical
 translation occurs between the controller and the data plane.
 Below is a complete list of such types planned for numerical
-translation by default with PSA, and have an annotation
-`p4runtime_translation` in the `psa.p4` include file:
+translation by default with PSA:
```
antoninbas: so it seems that the term "numerical translation by default" implies that the server must support any bijective function between the set of SDN ports and the set of device ports?
for types with no default numerical translation (all PSA types save for `PortId_t` and `ClassOfService_t`), the server is expected to at least support type casting between the SDN type and the device type (truncation or padding as appropriate)?

jafingerhut: I am open to clearer wording here if you have any ideas.  Originally when this text was written, we were only planning on `PortId_t` and `ClassOfService_t` having numerical translation on, and it seemed highly expected that for `PortId_t` that pretty much all PSA implementations would be doing so.

What do you think of changing that last sentence to "Below is a complete list of such planned types where many P4Runtime agents will perform numerical translation in PSA:"

antoninbas: I don't really have a good suggestion TBH, I wast just checking that we are on the same page. By adding a `p4runtime_translation` annotation to every type, we are basically implying that translation could be expected for all the types. When in reality, we are just using `p4runtime_translation` to suggest that the P4Info bitwidth should be the same regardless of the device. For all types except `PorId_t` and `ClassOfService_t`, we expect that the "translation" will likely be a simple cast.

I think the rest of your changes make this clear, which is why I approved the PR:

> Note that all of these types have a `@p4runtime_translation`
annotation in the `psa.p4` include file.  This is to ensure that when
the compiler generates a P4Runtime P4Info file from source programs,
it will not include the target-specific bit width in the P4Info file.
For the same P4 source program, the P4Info file contents are intended
to be identical for all targets.
This also enables P4Runtime systems to easily enable numerical
translation for any of these types later, if desired.

So to summarize it seems that the presence `@p4runtime_translation` implies that the P4Runtime server may have to perform casting between the SDN type and the device type. Having the annotation also implies that more complex "numerical translation" (arbitrary bijection) can be enabled at runtime. This is done by associating the translation URI with a specific bijection and communicating this to the P4Runtime server. We expect that a P4Runtime server running on a PSA device should be able to support arbitrary "numerical translation" for `PorId_t` and `ClassOfService_t` (other types are optional and only need to support basic casting).

If we are on the same page, I would just clarify the above block of text I quote with: "If the type bitwidth specified as the second parameter to the `p4runtime_annotation` is different from the device-specific bitwidth (of the underlying type), we expect the P4Runtime server to perform the appropriate casting. Additionally more advance numerical translation can be enabled at runtime for any type annotated with `p4runtime_translation`, although arbitrary numerical translation is only mandated for `PorId_t` and `ClassOfService_t`. To request arbitrary numerical translation for a give type, the P4Runtime system will expect the URI (first parameter to the  `p4runtime_annotation`) and the desired mapping."

jafingerhut: We are on the same page.  I added the text you suggested, in place of the last sentence of mine that you quoted.  (with p4runtime_annotation -> p4runtime_translation substitutions)

p4-16/psa/PSA.mdk
```diff
@@ -512,22 +511,23 @@ translation in the agent software, between target-specific values, and
...
+in the `psa.p4` include file.  This is to ensure that when the
+compiler generates a P4Runtime P4Info file from source programs, it
+will not include the target-specific bit width in the P4Info file.
+For the same P4 source program, the P4Info file contents are intended
+to be identical for all targets.
+
+If the type bitwidth specified as the second parameter to the
+`p4runtime_translation` is different from the device-specific bitwidth
+(of the underlying type), we expect the P4Runtime server to perform
+the appropriate casting. Additionally, more advance numerical
```
antoninbas: s/advance/advanced

p4-16/psa/PSA.mdk
cc10512: Nit (editorial): Below is a complete list of numeric types translated by default in PSA:

p4-16/psa/PSA.mdk
```diff
@@ -512,22 +511,23 @@ translation in the agent software, between target-specific values, and
...
-keys.
-
-If this is not a reasonable thing to do, the example program
-`psa-example-counters.p4` should be modified so that it does not do
-this, and scrub all other examples for any similar behavior.
-~
+Note that all of these types have a `p4runtime_translation` annotation
+in the `psa.p4` include file.  This is to ensure that when the
+compiler generates a P4Runtime P4Info file from source programs, it
+will not include the target-specific bit width in the P4Info file.
```
cc10512: it will include the P4Info file the type specified by the `p4runtime_translation` rather than the target-specific bit width.

p4-16/psa/PSA.mdk
jafingerhut: Changed.

p4-16/psa/PSA.mdk
jafingerhut: Changed.

jafingerhut:
Given the two approvals, going ahead and merging this in.  Thanks for review.

Issue #733, opened 2019-02-21
Clarify whether emit is supported on varbit values


jafingerhut:
Section 8.8 "Operations on variable-size bit types" of P4_16 language spec v1.1 says:

"Variable-length bit-strings support a limited set of operations:"

followed by a bullet item:

"The emit method of a packet_out extern object, which inserts a variable-sized bit-string with a
known dynamic width into the packet being constructed"

There seem to be at least two ways this could be interpreted (or misinterpreted):

(a) If `v1` is a variable of type `varbit<W>`, it is legal to do the operation `packet.emit(v1);`
(b) The previous scenario is _not_ supported.  It is supported to have a `varbit` type field inside of a header, and call `packet.emit(h1);` on a header, header stack, or header_union that contains a `varbit` field.

I think (b) is what is intended, but if so, it seems easy to misinterpret it as (a), given that this section is for operations _on the `varbit` type_, and all other operations in the other bullets can be performed directly on values of type `varbit`.

If (b) is the intended meaning, I am happy to suggest some changes that make this clearer, e.g. taking the extract and emit bullet items in Section 8.8 and discussing them separately than the operations that can be done directly on a value of type `varbit`.

jafingerhut:
Note: the latest version of p4c as of 2019-Feb-17 gives the following error if you attempt to make a call like `packet.emit(v1);` where `v1` is of type `varbit<W>`.  This error message is from `p4test`, so not specific to any architecture, I would think:
```
$ p4test checksum-ipv4-with-options.p4
checksum-ipv4-with-options.p4(215): [--Werror=type-error] error: packet.emit: argument must be a header, stack or union, or a struct or tuple of such types
        packet.emit(hdr.ipv4.options);
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

ChrisDodd:
As you might guess from the error message, you'll get the same error if `v1` is a `bit<N>` -- its not just varbits that are disallowed.

hanw:
Yes. emit should not be allowed on bit<N>, int<N> or varbit<N> directly. Spec is clear in section 8.5 and 8.6 that 'emit' is not a the supported operation. What is confusing is 8.8, where spec mentions 'emit' is a supported operation on varbit. @jafingerhut has pushed a PR to fix the confusion in spec.

jafingerhut:
Spec clarification has been merged in, so closing this issue.

jafingerhut:
[...]

PR #734, opened 2019-02-21
P4_16: Clarify the operations supported on values of type varbit

jafingerhut:
Also correct the cross-reference that was to section

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3318,21 +3318,30 @@ variable-size bit-string has an unknown dynamic width.
 
 Variable-length bit-strings support a limited set of operations:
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
One way to resolve issue https://github.com/p4lang/p4-spec/issues/733 if the option (b) described in that issue was the original intent, as I am currently guessing.

jafingerhut:
I will leave it up to Nate and Mihai what steps they wish to take to decide whether this proposed change is merged in, but thanks for reading it over, Han, and reporting the potential confusion.

jnfoster:
Cheers. We should briefly chat about this at an LWDG meeting.

mbudiu-vmw:
I will merge this, since I don't think it is controversial.

Issue #726, opened 2019-02-12
Is error allowed in header field?


hanw:
P4-16 v1.1.0 spec has a sentence in section 8.14.

>For a field with a type of enum or error, the unspecified value that is read might not be equal to any of the values defined for that type. 

Is error allowed in a header?

The following code currently has a type error
```
#include <core.p4>
#include <v1model.p4>

enum bit<8> e_t {
  e0 = 8w0
};

header h {
    e_t algo;
    error err;
}
```
```
../testdata/p4_16_samples/enum_field.p4(10): [--Werror=type-error] error: Field err of header h cannot have type error
    error err;
          ^^^
```

We should probably change the sentence to 
> For a field with a type of serializable enum, ...

or fix the compiler.

jafingerhut:
I agree that sentence is confusing, given that values of those types are not allowed inside of headers.

I think those sentences are intended to be true for uninitialized variables of those types, wherever they appear, e.g. inside a struct or as the out parameter of a control or function that was not initialized inside of that control/function, so those sentences seem like they ought to be kept somewhere in the spec, even if not in the section on headers.

hesingh:
The compiler is correct to not allow an error inside a header.  One does not know yet how to serialize an error.

jafingerhut:
Another data point: section 7.2.7 "Type nesting rules" says that `error` and non-serializable `enum` types are not allowed in a header.

jafingerhut:
The change moving that text into its own section has been merged into the spec now, so closing this issue.  Hopefully with these changes, with no mention of fields of type `error` or `enum` (serializable or otherwise) inside of the section on headers, it is clearer.

PR #735, opened 2019-02-21
P4_16: Move text on reading uninit vals and writing to invalid headers

jafingerhut:
I moved this text into its own section, and added cross-references to
it from where the text was moved from, Section 8.14 "Operations on
headers", and another in Section 8.13 "Operations on structs", even
though it really does apply to reading uninitialized of _any_ types.

Motivation for this movement: The moved text specifially describes the
behavior when an uninitialized value of type `enum` or `error` is
read, even though a header is not allowed to contain fields with those
types.  With the text in Section 8.14, one could be easily be confused
into believing that it implied that a header could contain fields of
those types.

The only changes I made to the moved text are as follows:

+ Changed the first sentence to "The result of reading a field" to
  "The result of reading a value", since reading uninitialized values
  applies not only to fields of headers and structs, but also to local
  variables with types like `bit<W>` or `bool`.

+ Added a few more explicit examples of situations where an
  uninitialized value can be read, changing this text: "reading any
  other value that has not been initialized, e.g. a field from a
  `struct`." to this: "reading any other value that has not been
  initialized, e.g. a field from a `struct`, any uninitialized
  variable inside of an `action` or `control`, or an `out` parameter
  of a `control` or `action` you have called, which was not assigned a
  value during the execution of that `control` or `action` (this list
  of examples is not intended to be exhaustive)."

+ Added this paragraph at the end of the new section:
  "For a top level `parser` or `control` in an architecture, it is up
  to that architecture to specify whether `control` parameters with
  direction `in` or `inout` are initialized when the control is
  called, and under what conditions they are initialized, and if so,
  what their values will be."

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3640,6 +3640,10 @@ if they have the same type and all of their fields can be recursively
 compared for equality.  Two structures are equal if and only if all
 their corresponding fields are equal.
 
+See Section [#sec-uninitialized-values-and-writing-invalid-headers]
+for a description of the behavior if `struct` fields are read without
+being initialized.
+
 ## Operations on headers { #sec-ops-on-hdrs }
 
 Headers provide the same operations as `struct`s. Assignment between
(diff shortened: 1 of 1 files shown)


jafingerhut:
Intended to address issue https://github.com/p4lang/p4-spec/issues/726

jnfoster:
Do you think we should review this (quickly) at a LDWG? I'm tempted to classify it as just a superficial re-organization of the existing text, but there are a few non-trivial changes.

jafingerhut:
It seems perfectly reasonable for anything but minor typo fixes to go through the LDWG, to me, so I don't mind at all if it is brought up there.

mbudiu-vmw:
Since we have 2 reviews I will merge this as well.

PR #737, opened 2019-02-23
PSA Add match kind docs

jafingerhut:


jafingerhut:
The text is based on this PR merged into the BMv2 documentation: https://github.com/p4lang/behavioral-model/pull/699

It adds notes on the relative priority of 'const entries' table entries, which should resolve this issue: https://github.com/p4lang/p4c/issues/1663

It resolves that issue without any changes required to p4c, by saying that if you have a longest-prefix match table, the relative priority of multiple entries specified via 'const entries' is based upon the prefix lengths, ignoring the order that the entries appear in the source code.

PR #738, opened 2019-02-23
Clarify the significance of order multiple table entries in 'const en…

jafingerhut:
…tries'

With this PR: https://github.com/p4lang/p4-spec/pull/737 the PSA
architecture now defines the significant of the order of multiple
entries in tables with a 'const entries' table property for all kinds
of tables, in a way that is consistent with the changes proposed here.

I will soon create a similar PR for the v1model architecture for BMv2
simple_switch that makes the same kinds of statements for it.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -5730,8 +5730,9 @@ Sec. [#sec-table-props]). The table key type must match the type of
 the element of the set. `actionRef` must be an action which
 appears in the table actions list, with all its arguments bound.
 
-Entries in a table are matched in the program order, stopping at the
-first matching entry.
+Entries in a table with at least one `ternary` search key field are matched in
+the program order, stopping at the first matching entry. Architectures should
+define the significance of entry order (if any) for other kinds of tables.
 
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
What if you have exact matching with the same key showing up multiple times?

jafingerhut:
Duplicate keys are not allowed in P4Runtime.  I guess I was assuming they should not be allowed in 'const entries', either.  I can add that explicitly in the text if you think it would be useful.

mbudiu-vmw:
I think this is a bug fix and it should be merged in the current spec.

jafingerhut:
AI Andy: Change new text to this:

"If the runtime API requires a priority for the entries of a table, e.g. tables with at least one `ternary` search key field, then the entries are matched in program order, stopping at the first matching entry.  Architectures should define the significance of entry order (if any) for other kinds of tables."

jafingerhut:
I believe I was given the verbal go-ahead to merge this in after addressing the comment during the 2019-Apr-01 LDWG meeting.  If I am in error there, my apologies.  Let me know and I can create another PR to fix it.

jnfoster:
👍

PR #739, opened 2019-02-25
Add variant of PSA example program that should work with P4Runtime v1.0

jafingerhut:
Also add comments to existing example program
psa-example-parser-error-handling.p4 explaining why it will not work
with P4Runtime v1.0.

### DIFF
--- a/p4-16/psa/examples/psa-example-parser-error-handling.p4
+++ b/p4-16/psa/examples/psa-example-parser-error-handling.p4
@@ -220,6 +220,11 @@ control packet_path_to_bits(out bit<3> packet_path_bits,
     }
     table packet_path_convert {
         key = {
+            // Note that P4Runtime v1.0 does not support P4 programs
+            // with table search key fields that have any type except
+            // bit<W>.  See psa-example-parser-error-handling2.p4 for
+            // a variant of this program that avoids this issue using
+            // an if-then-elseif daisy chain instead of a table.
             packet_path : exact;
         }
         actions = {
@@ -348,6 +353,11 @@ control handle_parser_errors(
     }
     table parser_error_count_and_convert {
         key = {
+            // Note that P4Runtime v1.0 does not support P4 programs
+            // with table search key fields that have any type except
+            // bit<W>.  See psa-example-parser-error-handling2.p4 for
+            // a variant of this program that avoids this issue using
+            // an if-then-elseif daisy chain instead of a table.
             parser_error : exact;
         }
         actions = {
--- a/p4-16/psa/examples/psa-example-parser-error-handling.p4
+++ b/p4-16/psa/examples/psa-example-parser-error-handling.p4
@@ -222,9 +222,16 @@ control packet_path_to_bits(out bit<3> packet_path_bits,
         key = {
             // Note that P4Runtime v1.0 does not support P4 programs
             // with table search key fields that have any type except
-            // bit<W>.  See psa-example-parser-error-handling2.p4 for
-            // a variant of this program that avoids this issue using
-            // an if-then-elseif daisy chain instead of a table.
+            // one of these:
+            //
+            // + bit<W>
+            // + a serialized enum with a base type of bit<W>
+            // + a typedef or type that has a base type of one of the
+            //   above.
+            //
+            // See psa-example-parser-error-handling2.p4 for a variant
+            // of this program that avoids this issue using an
+            // if-then-elseif daisy chain instead of a table.
             packet_path : exact;
         }
         actions = {
@@ -353,11 +360,8 @@ control handle_parser_errors(
     }
     table parser_error_count_and_convert {
         key = {
-            // Note that P4Runtime v1.0 does not support P4 programs
-            // with table search key fields that have any type except
-            // bit<W>.  See psa-example-parser-error-handling2.p4 for
-            // a variant of this program that avoids this issue using
-            // an if-then-elseif daisy chain instead of a table.
+            // See the comments for table packet_path_convert, which
+            // also apply for this table.
             parser_error : exact;
         }
         actions = {
--- a/p4-16/psa/examples/psa-example-parser-error-handling2.p4
+++ b/p4-16/psa/examples/psa-example-parser-error-handling2.p4
@@ -1,0 +1,619 @@
+/*
+Copyright 2019 Cisco Systems, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <core.p4>
+#include "../psa.p4"
+
+// This program is very similar in behavior to
+// psa-example-parser-checksum.p4, and was originally copied from that
+// program and then edited.
+
+// It is intended to demonstrate how to send a copy to the control
+// plane of any packet that experiences a parser error, regardless of
+// whether that parser error occurred in the ingress parser or egress
+// parser.
+
+// At least for packets that experience a parser error in the ingress
+// parser, the packet sent to the control plane should be exactly the
+// same as the one that arrived at the ingress parser when the parser
+// error occurred, prepended by the to_cpu_error_header_t with its
+// fields filled in so that the control plane can know as much about
+// the packet that experienced the parser error as possible.  It does
+// not matter whether the packet might have been resubmitted or
+// recirculated in order to cause it to arrive at the ingress parser,
+// or arrived there because the packet came from a normal port or the
+// CPU port.
+
+// For packets that experience a parser error in the egress parser,
+// whether the packet that comes after the to_cpu_error_header_t
+// header is identical to the one that arrived at the egress parser
+// when the error occurred depends upon the P4 implementation.  See
+// the comments in the egress control block for more details.
+
+// Below are the steps taken for packets that experience a parser
+// error in the ingress parser:
+
+// (1) The packet is parsed by ingress parser, experiencing a parser
+// error.
+
+// (2) The packet is sent to ingress control block with
+// istd.parser_error equal to the type of error that occurred.
+
+// (3) ingress control block calls control handle_parser_errors to
+// fill in meta.to_cpu_error_hdr.  This entire header is a field in
+// the user-defined metadata.  It also drops the ingress packet, and
+// sets ostd.clone to true so that an ingress-to-egress clone will be
+// created.  It assigns user-defined metadata field clone_reason a
+// value of PARSER_ERROR, so that the egress deparser knows why the
+// packet should be cloned.  This program does not contain code that
+// might make an ingress-to-egress clone for several different
+// reasons, but it is written in a way that has comments showing where
+// you would extend it to do so.
+
+// (4) the ingress deparser is run, which for cloned packets with
+// clone_reason equal to PARSER_ERROR, copies meta.to_cpu_error_hdr
+// into a field of clone_i2e_meta that will be carried with the
+// original packet to the egress pipeline.
+
+// (5) The egress parser receives the packet with istd.packet_path
+// equal to CLONE_I2E, and clone_i2e_meta initialized.  The egress
+// parser is written to copy the contents of clone_i2e_meta into a
+// user-defined field meta.to_cpu_error_hdr.  It also skips the normal
+// packet header parsing, so the packet will be considered all
+// payload, no headers.  This prevents any parser errors from
+// occurring in the egress parser.  It also assigns meta.clone_reason
+// to PARSER_ERROR.
+
+// (6) In the egress control block, packets with meta.clone_reason
+// equal to PARSER_ERROR have the user-defined hdr.to_cpu_error_hdr
+// copied from the meta.to_cpu_error_hdr, and then exit the egress
+// control block to proceed to the egress deparser.
+
+// (7) The egress deparser will receive a packet with packet_path
+// NORMAL, because the egress control block did not attempt to create
+// a clone, nor drop the packet.  It will emit the contents of
+// hdr.to_cpu_error_hdr at the beginning of the output packet, then
+// emit as many other headers as are in the program, but all of them
+// will be invalid because of the special egress parser case we
+// followed for this packet.  Then it will append the payload, which
+// is the entire unparsed packet.
+
+// Below are the steps taken for packets that experience a parser
+// error in the egress parser:
+
+// (1) packet is parsed by the egress parser, experiencing a parser
+// error.
+
+// (2) The packet is sent to egress control block with
+// istd.parser_error equal to the type of error that occurred.
+
+// (3) The egress control block will call the same control
+// handle_parser_errors mentioned in step (3) of the ingress parser
+// error case above, and do the same things.  It also drops the egress
+// packet, and sets ostd.clone to true so that an egress-to-egress
+// clone will be created.
+
+// (4) The egress deparser is run, which for cloned packets with
+// clone_reason equal to PARSER_ERROR, copies meta.to_cpu_error_hdr
+// into a field of clone_e2e_meta that will be carried with the packet
+// back to the beginning of the egress pipeline.  Unlike step (4) for
+// egress, the packet contents are not the packet contents as the
+// arrived to the egress parser, but as the rest of the egress
+// deparser code emits it.  If that is identical to the packet that
+// arrived at the egress parser, good.  Otherwise, the control plane
+// will unfortunately not receive a copy of the packet that caused the
+// egress parser error, but something different.  C'est la vie.
+
+// The rest of the steps are the same as ingress steps (5) and onwards
+// described above.
+
+
+
+typedef bit<48>  EthernetAddress;
+
+header ethernet_t {
+    EthernetAddress dstAddr;
+    EthernetAddress srcAddr;
+    bit<16>         etherType;
+}
+
+header ipv4_t {
+    bit<4>  version;
+    bit<4>  ihl;
+    bit<8>  diffserv;
+    bit<16> totalLen;
+    bit<16> identification;
+    bit<3>  flags;
+    bit<13> fragOffset;
+    bit<8>  ttl;
+    bit<8>  protocol;
+    bit<16> hdrChecksum;
+    bit<32> srcAddr;
+    bit<32> dstAddr;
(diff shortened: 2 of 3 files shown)


jafingerhut:
@hesingh Just pinging you for this updated example PSA program, for your information.

No, I don't actually _like_ the P4Runtime v1.0 restriction that doesn't let us use tables to look up enum and error type search key fields, but that is where things are today.  But we have to walk before we can run.

p4-16/psa/examples/psa-example-parser-error-handling.p4
```diff
@@ -220,6 +220,11 @@ control packet_path_to_bits(out bit<3> packet_path_bits,
     }
     table packet_path_convert {
         key = {
+            // Note that P4Runtime v1.0 does not support P4 programs
+            // with table search key fields that have any type except
+            // bit<W>.  See psa-example-parser-error-handling2.p4 for
```
antoninbas: To be accurate, we also support serializable enums and custom types where the underlying type is `bit<W>` even though the P4Runtime spec just says (in the known-limitations section):

> FieldMatch, action Param, and controller packet metadata fields only support unsigned bitstrings, i.e. values of type `bit<W>` (not the more general P4Data).

jafingerhut: Understood.  Open to a PR on the P4Runtime spec that adds those other supported cases?  I know it doesn't help support _this_ example program, but may as well list everything that we expect to work in the P4Runtime spec, if it is more than what the text implies now.

jafingerhut: Updated the comment to more accurately reflect P4Runtime v1.0 limitations.

antoninbas: Yes a PR to update the P4Runtime spec would be great

hesingh:
> @hesingh Just pinging you for this updated example PSA program, for your information.
> 
> No, I don't actually _like_ the P4Runtime v1.0 restriction that doesn't let us use tables to look up enum and error type search key fields, but that is where things are today. But we have to walk before we can run.

@jafingerhut Works for me - thanks for the ping.

p4-16/psa/examples/psa-example-parser-error-handling.p4
hemant-avatar3: > To be accurate, we also support serializable enums and custom types where the underlying type is `bit<W>` even though the P4Runtime spec just says (in the known-limitations section):
> 
> > FieldMatch, action Param, and controller packet metadata fields only support unsigned bitstrings, i.e. values of type `bit<W>` (not the more general P4Data).

The support for SerEnum and `type` to typeSpecConverter.cpp is being added via the PSA 1.1 PR: https://github.com/p4lang/p4c/pull/1736

jafingerhut:
@cc10512 Any concerns if I merge this in?  It makes no changes to the PSA spec at all -- it only adds an example program that is similar to an existing one, but avoids the use of values of type `error` and `enum` as table search keys, since those are not supported by P4Runtime v1.0.

p4-16/psa/examples/psa-example-parser-error-handling2.p4
```diff
@@ -0,0 +1,619 @@
...
+            packet_path_bits = 4;
+        } else if (packet_path == PSA_PacketPath_t.CLONE_E2E) {
+            packet_path_bits = 5;
+        } else if (packet_path == PSA_PacketPath_t.RESUBMIT) {
+            packet_path_bits = 6;
+        } else if (packet_path == PSA_PacketPath_t.RECIRCULATE) {
+            packet_path_bits = 7;
+        } else {
+            packet_path_bits = 0;
+        }
```
cc10512: This is a clear example why we should have serializable enums. This code should have been simply
```
packet_path_bits = packet_path
```
Or am I missing something?

p4-16/psa/examples/psa-example-parser-error-handling2.p4
```diff
@@ -0,0 +1,619 @@
...
+            error_idx = 6;
+        } else if (error_enum == error.BadIPv4HeaderChecksum) {
+            error_idx = 7;
+        } else if (error_enum == error.UnhandledIPv4Options) {
+            error_idx = 8;
+        } else {
+            error_idx = 0;
+        }
+    }
+}
```
cc10512: And similarly here.

jafingerhut:
There are multiple possible ways to improve upon the situation, e.g. in P4_16 language spec enhancements, P4Runtime enhancements, and/or PSA spec changes.  I created an issue here to start with for P4Runtime, which mentions (only briefly) some of the alternatives: https://github.com/p4lang/p4runtime/issues/191

p4-16/psa/examples/psa-example-parser-error-handling2.p4
jafingerhut: We could improve on this by changing the PSA-specific `PacketPath_t` type from `enum` to a serializable `enum`, but that won't help for the `error` case.

jafingerhut: It won't help for the error case with the current specs, that is, because there is no "serializable error" type.

cc10512: Indeed, that's why I think we need to consider whether we want to introduce implicit casts.

antoninbas: I don't think the ability to assign numerical values to errors and `PacketPath_t` members is the solution. This is really a short-coming of P4Runtime v1.0 which does not support errors & un-serializable enums as match keys.
If you support implicit casts for errors or change `PacketPath_t` to a serializable enum, you introduce more complexity for the target and potentially a lot of headaches for the programmer. As long as the numerical values are not embedded in a packet and are not used in arithmetic expressions you are fine (this is very similar to P4Runtime translation for user-defined types, for which we have explicit casts to the underlying type, but no implicit casts). But as soon as you emit the numerical value to a packet, either you are fine with a target-dependent value or the compiler itself needs to detect that situation and introduce a translation table.
I personally recommend keeping errors and un-serializable enums as they are, this is an issue that can be solved entirely in the realm of P4Runtime.

hesingh:
Regarding error, a table key has to use an integer to interoperate with the data plane.  However, P4Runtime protobuf definition uses a human-readable string.  We have to store the string somewhere and link the string with the integer.

jafingerhut:
@hesingh Note that on that implementation detail, it is perfectly reasonable for P4Runtime messages to pass strings between the P4Runtime client (controller) and P4Runtime server to represent enum values.  The knowledge of how these are represented in the data plane need _never_ be exposed to the clients/controllers.  They don't need to know that detail, and it could be different from target to target.

All of those details can be encoded somewhere inside the "binary blob" that is the result of a compiler compiling P4 source code to a specific target.

hesingh:
Sounds reasonable.  Note, Google's protobuf supports error reporting between client and server using an integer error and string using English text. 

Adding error as a table key would be useful today.

jafingerhut:
If you want to implement error as a table key, I am not arguing against it.  However, there is no place to put that in P4Runtime v1.0 Protobuf messages, so you would have to extend beyond what it is capable of, which a future version of P4Runtime seems likely to do.

People are welcome to make prototypes that go beyond the specs, but whether they are willing today to add experimental things into the p4lang Github repositories depends upon the details of the experimental thing.  I am not the gatekeeper there.

hesingh:
Ah, forgot about P4Runtime table key restriction.  We cannot bring such a change in P4Runtime 1.x because the change won't be backward compatible.  We'd need P4Runtime 2.0 for a new table key type.

jafingerhut:
Your statement that it _cannot_ be backwards compatible is, I am pretty sure, not true.

I believe that every Google Protobuf 3 message field, which P4Runtime is based on, are _optional_, at least as far as the Protobuf implementation and gRPC are concerned.  Which fields are required to be present or optional for P4Runtime's purposes is up to P4Runtime to define.

One could therefore add a new field to the MatchField messages, for example, that let you specify more arbitrary types, e.g. via a field of type P4DataTypeSpec (sorry, going from memory on that message name there because I'm lazy -- sorry if I gave the wrong one), and then P4Runtime could require that either that field is set, or the current fields describing the type are set, but never both.

Backwards compatibility: unlocked!

p4-16/psa/examples/psa-example-parser-error-handling2.p4
hesingh: @antoninbas I agree an explicit Cast. Working within the realm of P4Runtime, what do you think of this change for `error` and table key?  It's either bitwidth or type_error as set. Wouldn't such a change be a 1.x kind of change to P4Runtime?

```
diff --git a/proto/p4/config/v1/p4info.proto b/proto/p4/config/v1/p4info.proto
index bc95375..58823d6 100644
--- a/proto/p4/config/v1/p4info.proto
+++ b/proto/p4/config/v1/p4info.proto
@@ -170,6 +170,8 @@ message MatchField {
   Documentation doc = 6;
   // unset if not user-defined type
   P4NamedType type_name = 8;
+  // MatchField is an error
+  P4ErrorType type_error = 9;  
 }
 
 message Table {

```

jafingerhut: It seems to me that _if_ you are going to generalize it in that way, why not instead add a line like:
```
P4DataTypeSpec other_type = 9;
```
then say that at most one of `bitwidth` and `other_type` is set.  Then you can specify _any_ legal P4_16 type in the message.  The P4Runtime 1.x spec adding this could still limit what is supported to less than all types, e.g. maybe you do not want to jump so quickly to adding support for a header, header_union, or header stack as supported (if ever).

hesingh: `P4DataTypeSpec` uses `oneof`. Thus, we are adding a `oneof` and run into the issue below.

https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility-issues

jafingerhut: I read the text at that link, and it isn't clear to me that it is recommending that you should not add a oneof.  Why do you believe it is an issue?

hesingh: The P4DataTypeSpec is included below.  The thought is "it's either bitwidth is set (non-zero) or the P4DataTypeSpec member is".  The user set P4DataTypeSpec->type_spec to bitstring.   So, do we use that fact that P4DataTypeSpec is set or because the user is using bitstring, the bitwidth is also set?  

Thus, it is best to remove bitwidth from the MatchField message - if we do, we are not backward compatible. 

```
// Describes a P4_16 type.
message P4DataTypeSpec {
  oneof type_spec {
    P4BitstringLikeTypeSpec bitstring = 1;
    P4BoolType bool = 2;
    P4TupleTypeSpec tuple = 3;
    P4NamedType struct = 4;
    P4NamedType header = 5;
    P4NamedType header_union = 6;
    P4HeaderStackTypeSpec header_stack = 7;
    P4HeaderUnionStackTypeSpec header_union_stack = 8;
    P4NamedType enum = 9;
    P4ErrorType error = 10;
    P4NamedType serializable_enum = 11;
    P4NamedType new_type = 12;
  }
}
```

hesingh: With p4runtime v1.0, the server is checking if bitwidth is set and assumes the MatchField type is `bit<W>/init<W>`.  One adds P4DataTypeSpec to p4runtime v1.1.  Now, the server checks if P4DataTypeSpec is set and determines the type.  If P4DataTypeSpec is not set, it's a failure or is the message a v1.0 message? With v1.1, bitwidth is used only to serialize the type.   I think, it's best to wait for Antonin to give his input.

jafingerhut: I am not here proposing the rules for a future P4Runtime spec.  I am saying that if backwards compatibility is desired, it is straightforward to achieve.  Yes, achieving backwards compatibility in the way I have suggested means that one must create a rule to specify how to interpret the future extended P4Info messages.

For example, future P4Info message creators could simply stop using `bitwidth` as a field entirely, leaving it out, using the P4DataTypeSpec message instead when the type of the table search key field is `bit<W>`.  Or alternately, the P4Runtime spec could have the rule that if the type is `bit<W>`, an implementation is free to use either the older `bitwidth` field or the newer way of doing it, and either should be accepted to have the same meaning.

To say that one _must_ break backwards compatibility by removing a field is completely against the goals of Protobuf messages.  There is no absolute need or requirement to break backwards compatibility.  The only reason to do so is some imagined convenience.

p4-16/psa/examples/psa-example-parser-error-handling.p4
hesingh: Having heard "P4Runtime 1.0 already supports `serEnum` and custom types.", I'd like to investigate `error`.  How do we use `error` as a table key?

P4-16 `error` type does not have serialization defined yet.  A table key requires a key size in `bit<W>` format.  The data plane reads this key for its use.  Therefore, one possible format of the key is a serializable error enum.  For brevity, let's call it `serErrorEnum` similar to use of `serEnum` internally in p4c code.  `serErrorEnum` is a new type for p4c.  An example is provided below.

```
error bit<16> ParserErrors {
    ParseError = 0x0, 
    PacketTooShort = 0x1
}
```
p4c folks can comment on proposed p4c changes and P4Runtime folks can comment on p4runtime implications.

jafingerhut: Where have you heard this? "P4Runtime 1.0 already supports serEnum and custom types."

There is some truth to it, but maybe not in the way you are thinking.

P4Runtime 1.0 supports serEnum pretty much everywhere, including a table search key field, value_set key field, action parameter, controller header metadata field.  It does this by encoding it in P4Runtime messages the same way it encodes `bit<W>` values -- as binary numbers.

P4Runtime 1.0 also supports arbitrary P4 types, but _only in specific use cases_.  For example, it supports arbitrary P4 typed values as digest message fields, or as register array elements.  P4Runtime 1.0 _does not support_ arbitrary P4 types as any of the above 4 things (table search key fields, etc.).  That is the reason there is a specific note in Section 20 of the spec that says so.

The method for serializing `enum` types that are not serializable enums (but only for the cases I mentioned above where they are supported, like digest messages and register array elements) is to encode them in P4Runtime messages as strings, which are the names of the `enum` members from the P4 source code.  Even though they are likely encoded as small integers in the data plane, they are _not_ encoded as integers in P4Runtime messages.  It is the responsibility of the P4runtime server in such cases to translate between the target-specific small integer values in the data plane, and the strings used in P4Runtime messages.  That translation can be different from one target to another.  The P4Info file does _not_ contain the name-to-number mapping for `enum` types that are not also serializable enums, because it is target-specific, and the P4Info file is intended to be the same across all targets.

jafingerhut: Here is a v1model architecture program with a digest, where the digest message contains a field of a non-serializable enum type, and also a field of type error: https://github.com/jafingerhut/p4-guide/blob/master/p4runtime/v1model-grab-bag.p4

And here is the text format of the P4Info file created via the following command, using latest p4c as of 2019-Feb-26:
```
$ p4c-bm2-ss v1model-grab-bag.p4 --p4runtime-files v1model-grab-bag.p4.p4.info.txt
```
https://github.com/jafingerhut/p4-guide/blob/master/p4runtime/v1model-grab-bag.p4.p4info.txt

Here is the part of that file defining a struct type `test_digest_t` that has the enum and error type fields: https://github.com/jafingerhut/p4-guide/blob/master/p4runtime/v1model-grab-bag.p4.p4info.txt#L199-L228

Immediately below those lines you see the P4Info messages that define the names of the members of type error, and the names of the members of type `MyPacketTypes`.  There are no integer encodings, because they are target-specific and thus do not belong in the P4Info file, by design.

hesingh: In this Issue's note from Antonin above as follows:
"To be accurate, we also support serializable enums and custom types where the underlying type is bit<W> even though the P4Runtime spec just says (in the known-limitations section):"

Thanks for the clarification about `type` not used as a table key.  However, I am not interested in `type` used as a table key.   My main focus if to use new `serErrorEnum` as table key. 

Regarding translation from data plane integers to string, for sending string between p4runtime client and server, is a topic you explained yesterday - thanks.  If a string is target-specific, it's included in the binary blob. 

> Where have you heard this? "P4Runtime 1.0 already supports serEnum and custom types."
> 
> There is some truth to it, but maybe not in the way you are thinking.
> 
> P4Runtime 1.0 supports serEnum pretty much everywhere, including a table search key field, value_set key field, action parameter, controller header metadata field. It does this by encoding it in P4Runtime messages the same way it encodes `bit<W>` values -- as binary numbers.
> 
> P4Runtime 1.0 also supports arbitrary P4 types, but _only in specific use cases_. For example, it supports arbitrary P4 typed values as digest message fields, or as register array elements. P4Runtime 1.0 _does not support_ arbitrary P4 types as any of the above 4 things (table search key fields, etc.). That is the reason there is a specific note in Section 20 of the spec that says so.
> 
> The method for serializing `enum` types that are not serializable enums (but only for the cases I mentioned above where they are supported, like digest messages and register array elements) is to encode them in P4Runtime messages as strings, which are the names of the `enum` members from the P4 source code. Even though they are likely encoded as small integers in the data plane, they are _not_ encoded as integers in P4Runtime messages. It is the responsibility of the P4runtime server in such cases to translate between the target-specific small integer values in the data plane, and the strings used in P4Runtime messages. That translation can be different from one target to another. The P4Info file does _not_ contain the name-to-number mapping for `enum` types that are not also serializable enums, because it is target-specific, and the P4Info file is intended to be the same across all targets.

hesingh: I have already said several times, I know the p4runtime 1.0 spec specifies using human-readable text for `error` and `enum`.  If a table key uses a new `serErrorEnum`, the key uses integers, but p4c API generation changes the integers to string representation.

jafingerhut: OK, then I am unfamiliar with what `serErrorEnum` is, probably because of my ignorance of the compiler internals.  If it is some p4c internal representation of a P4_16 value of type `error`, then it cannot be used as a table search key for P4Runtime v1.0.  Full stop.  It cannot.  Don't try.

Extend P4Runtime v1.0 somehow so it can be used as a table search key is possible.  Trying to cram it into P4Runtime v1.0 is a job for Don Quixote.

jafingerhut: And if you want to include an `enum` or `error` field in a digest message, it already works (or, at least it already successfully generates a correct-looking P4Info file -- other parts of the implementation may be incomplete), as I demonstrated with the example  P4_16 v1model architecture program I linked to in a comment earlier today.

PR #740, opened 2019-02-26
Mention that an enum with an underlying type is also called serializable

jafingerhut:
Before this change, the word "serializable" does not occur anywhere in
the language specification.  An enum value with an underlying type is
often called a serializable enum in conversation between P4
implementers, including in the p4c implementation, so having a way to
find it in the language spec by that name seems like a good idea.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2134,9 +2134,9 @@ created type.  The underlying representation of the `Suits` enum is not
 specified, so their "size" in bits is not specified (it is
 target-specific).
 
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -2135,6 +2135,8 @@ specified, so their "size" in bits is not specified (it is
 target-specific).
 
 It is also possible to specify an `enum` with an underlying representation.
+These are sometimes called serializable `enum`s, because fields of
```
mbudiu-vmw: "headers are allowed to have fields with such enum types"

"fields of" sounds confusing.

PR #743, opened 2019-03-07
Proposed new error value for implementations that restrict parser ali…

jafingerhut:
…gnment

Similar to the existing statement that some implemtations are allowed
to restrict headers to be multiples of 8 bits long, such
implementations are also likely to impose similar restrictions on the
2-argument extract and advance calls, so that the parsing 'cursor' is
always at a multiple of 8 bits from the beginning of the packet.
These changes propose a way to make that explicit in the
specification.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2231,7 +2231,7 @@ size." Headers containing `varbit` fields have "variable
 size." The size (in bits) of a fixed-size header is a constant, and it
 is simply the sum of the sizes of all component fields (without
 counting the validity bit). There is no padding or alignment of the
-header fields. Architectures may impose additional constraints on
+header fields. Implementations may impose additional constraints on
 header types---e.g., restricting headers to sizes that are an integer
 number of bytes.
 
@@ -5021,6 +5021,8 @@ is captured by the following pseudo-code:
 ~ Begin P4Pseudo
 void packet_in.extract<T>(out T headerLvalue,
                           in bit<32> variableFieldSize) {
+   // implementations _may_ include the following line, but need not
+   verify(variableFieldSize[2:0] == 0, error.ParserInvalidAlignment);
    bitsToExtract = sizeOfFixedPart(headerLvalue) + variableFieldSize;
    lastBitNeeded = this.nextBitIndex + bitsToExtract;
    ParserModel.verify(this.lengthInBits >= lastBitNeeded, error.PacketTooShort);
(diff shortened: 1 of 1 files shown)


jafingerhut:
Proposed change motivated by implementation work in p4c and BMv2 on advance: https://github.com/p4lang/behavioral-model/issues/727

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5021,6 +5021,8 @@ is captured by the following pseudo-code:
 ~ Begin P4Pseudo
 void packet_in.extract<T>(out T headerLvalue,
                           in bit<32> variableFieldSize) {
+   // implementations _may_ include the following line, but need not
+   verify(variableFieldSize[2:0] == 0, error.ParserInvalidAlignment);
```
mbudiu-vmw: I think I would prefer this check to be commented-out.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7090,7 +7094,9 @@ error {
     NoMatch,           /// 'select' expression has no matches.
     StackOutOfBounds,  /// Reference to invalid element of a header stack.
     HeaderTooShort,    /// Extracting too many bits into a varbit field.
-    ParserTimeout      /// Parser execution time limit exceeded.
+    ParserTimeout,     /// Parser execution time limit exceeded.
+    ParserInvalidAlignment  /// extract or advance would leave parser with
```
mbudiu-vmw: This error name is very specific. How about InvalidArgument, which could probably be reused for other purposes?

jafingerhut:
Agreed with all you said in the previous two comments.  No rush on my part to get it merged.  Just wanted to create the PR while the issue was fresh in mind.

p4-16/spec/P4-16-spec.mdk
jafingerhut: done

p4-16/spec/P4-16-spec.mdk
jafingerhut: done

jafingerhut:
AI Andy: Add this change to the change log at the end of the spec.  Replace "Implementations" with "Targets".

jafingerhut:
I believe I was verbally given the go-ahead to merge this in after addressing the comments made during the 2019-Apr-01 LDWG meeting.  If I am mistaken, or further changes or needed, let me know and I can fix things.

jnfoster:
👍

PR #747, opened 2019-03-25
Correct a cross reference for operations allowed on tuple types

jafingerhut:


jafingerhut:
I am not sure about this, but the section on operations on list expressions, which have type tuple, seem to be what was probably intended as the cross reference here, rather than to the section on operations on set expressions, which as far as I can see have nothing to do with type tuple.

jnfoster:
I haven't built the Madoko, but this looks right to me.

PR #750, opened 2019-04-02
P4_16: Add header size operations

jafingerhut:


jafingerhut:
Sorry, I tried rebasing the original PR here: https://github.com/p4lang/p4-spec/pull/712  but got lost in some kind of conflict.  Creating a new PR.

After the LDWG meeting on 2019-Apr-01, there were some additional comments made on the issue, asking for clarification of behavior of minSizeInBits/Bytes on header stacks and header_union.  Frankly, the real basic use case of this new functionality is for getting the size of plain old simple headers.  That is enough.  So I've left out defining it on header stacks and header_union.  It is about as streamlined as it can get now.  If someone really doesn't want minSizeInBits(), we can remove that, too.  I think minSizeInBytes is enough for the basic minimal use case that motivated creating this PR.

jnfoster:
Sounds good to me.

mbudiu-vmw:
How about a one line changelog?

jafingerhut:
I added a line to the change log for this addition, as well as several others made since v1.1.0 that did not have lines there.  I believe it may be complete, depending upon how small of changes one wishes to mention there.

jafingerhut:
Removing additions to change log that are not in this PR.  I will create a separate PR that adds lines to the change log that I think are missing.

mbudiu-vmw:
Anything else we need to merge this PR?

jafingerhut:
Merged in after resolving Git conflicts, per verbal go-ahead given in 2019-Apr-16 LDWG meeting, and approvals by reviewers.

PR #751, opened 2019-04-04
Add some changes to change log made since spec version 1.1

jafingerhut:


jafingerhut:
Some of these may be considered too minor to be included in the change log, but they do seem to have been merged in since version 1.1 of the spec was released, and not yet in the change log.

jafingerhut:
Fixed the markup, and added section references for the items being added to the change log.  I removed the item "Constrain keys to be unique in annotation values specified as key-value pairs." because as far as I can tell that change was later superseded by the change "Relaxed the structure of annotation bodies."

PR #755, opened 2019-04-23
One explicit cast was omitted

mbudiu-vmw:
The diff is actually quite small, but there was some extraneous spaces as well.


mbudiu-vmw:
I will merge it, because this was really a bug.

PR #773, opened 2019-07-13
Added structure-valued expressions

mbudiu-vmw:
This should be merged after #717

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3677,6 +3677,44 @@ tuple<bit<32>, bool> x = { 10, false };
 
 The empty list expression has type `tuple<>` - a tuple with no components.
 
+## Structure-valued expressions { #sec-structure-expressions}
+
+One can write expressions that evaluate to a structure or header.  The
+syntax of these expressions is given by:
+
+~ Begin P4Grammar
+expression ...
+    | '{' kvList '}'
+    | typeName '{' kvList '}'
+    ;
+
+kvList
+    ;
+~ End P4Grammar
+
+If present, the `typeName` must be the name of a `struct` or `header`
+type.  The `typeName` can be omitted if it can be inferred from
+context, e.g., when initializing a variable with a `struct` type.  The
+following example shows a structure-valued expression used in an
+equality comparison expression:
+
+~ Begin P4Example
+struct S {
+    bit<32> a;
+    bit<32> b;
+}
+
+S s;
+...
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
Fixes #772

vgurevich:
Can we also agree that if a field is not mentioned in kvlist, it will be automatically assigned a 0 value and also agree that kvlist can be empty, meaning that:
```
mystruct_t mystruct = {};
```
will do the right thing?

mbudiu-vmw:
@vgurevich : There is a discussion on this topic https://github.com/p4lang/p4-spec/pull/717
I have added this as a topic to discuss in the LDWG.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3605,6 +3605,44 @@ tuple<bit<32>, bool> x = { 10, false };
...
 
+## Structure-valued expressions { #sec-structure-expressions}
+
+One can write expressions that evaluate to a structure or header.  The
+syntax of these expressions is given by:
+
+~ Begin P4Grammar
+expression ...
+    | '{' kvList '}'
+    | typeName '{' kvList '}'
```
liujed: This creates the awkward situation where all of the following are valid syntax:
* `{ a=0, b=1 }`
* `S { a=0, b=1 }`
* `(S) { a=0, b=1 }`
* `(S1) S2 { a=0, b=1 }`

One way around this awkwardness would be to only allow the `'{' kvList '}'` production and remove the `typeName '{' kvList '}'` production, but then typing of struct-valued expressions can become ambiguous when the program has defined multiple types that match.

Another alternative would be demote `'{' kvList '}'` from being a first-class expression, and only allow it as a struct initializer: `S s = { a=0, b=1 };`. Then, struct-valued expressions would always use the `S { a=0, b=1 }` form. My impression is that this would be compatible with what we released in 1.2.0, but I haven't confirmed.

mbudiu-vmw: the typechecker will reject today casts with struct types.

mbudiu-vmw: maybe that's a good rule: if a cast is to a Struct or header type then it is only allowed for struct-valued expressions.

mbudiu-vmw:
We should try the syntax with a cast instead of a type.

jafingerhut:
Would it make sense that if/when this is merged in, it should _replace_ the text about structure initializers?  It seems that this proposal does everything that does, and more.

PR #775, opened 2019-08-06
Allow the string type

mbudiu-vmw:


jafingerhut:
Is it implied by the "no assignment operation exists for strings" that they cannot be passed as parameters with a direction?

jnfoster:
The section on copy-in / copy-out semantics talks about "copying" and "l-values," so we should either use the same terminology here or (better) say explicitly what we mean.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1857,10 +1860,21 @@ Boolean values are not integers or bit-strings.
...
-Section [#sec-string-literals]. String literals can only be used in
-annotations (described in Section [#sec-annotations]). For example,
+The type `string` represents strings.  There are no operations on
+string values; one cannot declare variables with a `string` type.
+Parameters with type `string` can be only directionless (see Section
+[#sec-calling-convention]).  P4 does not support string manipulation
+in the dataplane; the `string` type is only allowed for denoting
+compile-time constant string values.  These may be useful, for
+example, a specific target architecture may support an extern function
+for logging with the following signature:
```
cc10512: Reference the string encoding from the 'String literals' section or tighten it more to UTF8?

mbudiu-vmw: I don't think it's necessary: we don't say anything else about string literals in this section.

mbudiu-vmw:
I think we agreed to merge this for 1.2.0

Issue #761, opened 2019-05-11
Add miss to the list of apply() results


vgurevich:
In the course of P4_16 programming I find that whenever I use `table.apply().hit` I always use it as follows:
```
    if (!ipv4_host.apply().hit) {
        if (!ipv4_route.apply().hit) {
            last_resort.apply();
        }
    }
```

It looks to me that adding an opposite of `hit`, i.e. `miss` will allow the code to be more readable, i.e.:
```
    if (ipv4_host.apply().miss) {
        if (ipv4_route.apply().miss) {
            last_resort.apply();
        }
    }
```

vgurevich:
https://github.com/p4lang/p4-spec/pull/778

mbudiu-vmw:
Can we close this issue?

vgurevich:
Absolutely. I think it is all done.

PR #778, opened 2019-08-17
Issue #761: Add support for the apply() miss field

vgurevich:
This is a small change to the spec that allows table.apply().miss in addition to table.apply().hit

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6035,7 +6035,8 @@ struct apply_result(T) {
 ~ End P4Pseudo
 
 The evaluation of the `apply` method sets the `hit` field to `true`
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -6035,16 +6035,20 @@ struct apply_result(T) {
 ~ End P4Pseudo
 
 The evaluation of the `apply` method sets the `hit` field to `true`
-if a match is found in the lookup-table. This bit can be used to drive
-the execution of the control-flow in the control block that invoked
-the table:
+and the field `miss` to `false` if a match is found in the lookup-table.
+These bits can be used to drive the execution of the control-flow in the
```
liujed: Also specify behaviour when match is not found in the look-up table?

Issue #789, opened 2019-11-04
Structured annotations to represent properties (kv-pairs, lists of values, etc)


chrispsommers:
The P4 Runtime WG has identified the need for annotations which declare kv-pairs, expressions, etc. which will get transformed into P4Info. We made some starts at this (https://github.com/p4lang/p4runtime/issues/234, https://github.com/p4lang/p4runtime/pull/238) but the recent change to unstructured annotations in p4-16 1.2 stymied that. We'd like to begin discussing P4-16 enhancements to support this in a uniform way. Below are recent minutes of the P4 API WG which summarize the status to date. Structured annotation are appearing in many venues both open and proprietary and we should define a uniform mechanism before too many ad-hoc conventions get entrenched.

Update on discussion with LDWG regarding annotations support:

* Reluctance from the LDWG to change anything before the P4_16 1.2 release, they wish we had brought it up to them sooner
* Justification for moving to totally unstructured annotation bodies: “Because annotations are used for extensions, they tend to go beyond what can be recognized by the lexer/parser. So minimizing the syntactic requirements maximizes flexibility. We could put them into strings, but then any annotations dealing with strings would need to be escaped.” - Nate Foster
* The P4Runtime WG recognizes this but we also think that our desire for structured annotations support is justified, based on past feedback from members of the WG.
Nate’s proposal to provide support for structured annotations: use a different syntax for different types of annotations.
```
@annot(...) -> parsed uninterpreted, as now
@annot{k1=v1, k2=v2, ...} -> parsed as list of key-value pairs
@annot[v1, v2, ...] -> parsed as list of values
```
The proposed syntax is reminiscent of dictionary / list syntax in other common programming languages such as Python
The same annotation can combine lists and key-value pairs but one needs to repeat the annotation:
```@foo{k1=v1, k2=v2} @foo[v1, v2] my_table { … }```

Vladimir: there is consensus in the LDWG that we should try to rely less on annotations and more on language-native constructs. We also discussed being able to define the runtime APIs for different P4 objects (such as extern types) directly in the P4 program (architecture). We could have some built-in way to attach properties to all P4 objects. There are also some inconsistencies: for some objects the annotations come before the declaration and for other they come after.

Chris: this is true but annotations are definitely needed, being able to specify custom extensions is a must and it has been demonstrated multiple times in this WG. For common scenarios, we should try to uniformize at the language level.
Vladimir: annotations are clunky when you need to refer to other P4 objects such as fields

[...]

PR #796, opened 2019-12-04
Add structured annotations to P4 grammar. Issue #789

chrispsommers:


stefanheule:
I have a few questions:
- It appears that string literals are allowed, but integer expressions as long as they evaluate to a constant at compile time.  Why the discrepancy?
- Are integers infinite-precision? Or is there a bit width?
- Would it make sense to allow boolean constants?
- It seems that multiple annotations with the same name accumulate their contents, which allows the mixing of expressions and key-value pairs.  However, you can't mix them in the same annotation.  That seems odd, why is that?
- What happens if a key-value list has the same key several times?

Probably some of these questions have multiple reasonable answers, but they should probably all be answered before integrating this into the spec.

jafingerhut:
@stefanheule In response to your question: "It appears that string literals are allowed, but integer expressions as long as they evaluate to a constant at compile time. Why the discrepancy?"

Is your question "Why aren't expressions whose results is a string allowed, as well as expressions whose result is an integer?"  If so, I believe one answer would be: "P4 has no expressions whose result is a string, except string literals".  i.e. no string concatenation operations, substrings, etc.  So one way to think of it is that as far as P4_16 is defined now, it _does_ allow arbitrary expressions whose result is a string, but those currently always look like string literals.

stefanheule:
Makes sense, I didn't think about that.  I would prefer a phrasing that treats all constants the same, even if right now there is no functional difference.  But obviously this is a very minor point.

jafingerhut:
It does seem reasonable to say that there could be arbitrary expressions that at compile time can be evaluated, and result in one of the acceptable types, e.g. if boolean is added to the list, then the result of the expression could be any value of type string, int, or boolean.

chrispsommers:
@stefanheule @jafingerhut Thanks for the comments, I agree any allowable types should require the same requirement for compile-time evaluation.

antoninbas:
@stefanheule regarding integer precision, I'll copy here what I posted to the p4-api mailing list

> Regarding one topic we discussed at the meeting: support for infinite-precision integers as values in the body of structured annotations. I think one thing we could do is specify in the P4Runtime spec (and not the P4 spec) that P4Info only supports signed 64-bit integers. When generating P4Info, if we encounter a value that doesn't it within an int64, we can print an error message inviting the user to switch to a string.

chrispsommers:
@stefanheule Thanks for the insightful comments and questions.
>It appears that string literals are allowed, but integer expressions as long as they evaluate to a constant at compile time. Why the discrepancy?

Agreed, all types should be evaluate to compile-time constants and I should update the wording (but bear in mind P4 treats strings very simply at this point in time).
> Are integers infinite-precision? Or is there a bit width?

As @antoninbas said, the language should allow  infinite-precision but P4Runtime or any other consumer  may impose restrictions e.g.  64-bit.
> Would it make sense to allow boolean constants?

That seems to be the popular consensus, I'll revise accordingly

>It seems that multiple annotations with the same name accumulate their contents, which allows the mixing of expressions and key-value pairs. However, you can't mix them in the same annotation. That seems odd, why is that?

I was trying to keep it simple for the language and grammar. I'll leave this as-is and see how it goes in the LDWG with the other changes/clarifications.  I'm not sure many use-cases need free mixing of the kv's and expressions,  but let's see what others think. One of the suggestions in the API WG was to keep it simple and sufficient for known use-cases.

>What happens if a key-value list has the same key several times?

Should be an error, I'll clarify that.

mbudiu-vmw:
We proposed removing from the spec the request for annotations to be merged.

liujed:
During today's discussion, there was a question of whether it's useful to disambiguate empty lists, empty kv-lists, and empty unstructured annotations for annotation consumers.

stefanheule:
Another thing that was suggested was to allow P4 lists as a value in addition to int, bool and string.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7097,16 +7099,157 @@ annotationBody
...
+`expressionList` may be empty or contain a comma-separated list of members. A
+`kvList` consists of one or more `kvPair`s, each consisting of a key and a
+value; the value is an `expressionList`. 
+
+To avoid ambiguity, an empty `stucturedAnnotationBody` is by definition
+considered to contain an empty `expressionList`. Furthermore, the value of a `kvPair` may
+be the empty `expressionList`, in which case the pair is considered to
+contain a key without a value.
+
+The `expression`s within an `expressionList` must evaluate at compile-time to
```
mbudiu-vmw: This is actually not true for annotations such as `@length`; in that annotation the expression actually refers to some structure field members and leads to code being generated to evaluate the expression dynamically.

I notice that the `@length` annotation is actually not documented in the P4-16 spec; it is used to represent p4-14 varbit field length information. I wonder whether we should document it. 

The point is that we do not mandate in the spec when and how the expression that is an argument to the annotation is evaluated. It does not have to be a compile-time evaluated expression.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7097,16 +7099,157 @@ annotationBody
...
+~ End P4Example
+
+
+#### Illegal Duplicate Annotation
+
+The following example is invalid because the annotation `name` occurs more
+than once on the same token, e.g. `table t`:
+
+~ Begin P4Example
+@DupAnno[k1=4]
```
mbudiu-vmw: Do we want to forbid duplicates?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7097,16 +7099,157 @@ annotationBody
...
+
+kvPair
+    : name '=' expressionList
+    ;
+
+~ End P4Grammar
+
+### Structured Annotation Examples
+
+#### Empty Expression List
```
mbudiu-vmw: I don't think that a subsection is warranted for each example.

p4-16/spec/P4-16-spec.mdk
chrispsommers: That was my intention.

p4-16/spec/P4-16-spec.mdk
chrispsommers: Sure, I can remove the headings.

p4-16/spec/P4-16-spec.mdk
chrispsommers: Are you suggesting that these restrictions should appear in the P4Runtime spec instead?

stefanheule:
Right now the spec says

> All `expression`s (and nested expressions) within a `stucturedAnnotationBody`
> must evaluate at compile-time to string literals, infinite-precision integers or
> booleans.

It is not clear what kind of nested expressions are allowed, and it is also not clear if nested expressions can themselves be nested.

Is `{1,2,{1,2,{1,2}}}` okay? Is `{a=1,b=2}` a valid "compile-time constant" expression?

Another question I had was if there is any kind of type checking? Can list elements be of different types?

chrispsommers:
@stefanheule Sorry for the delayed reply. See responses to your questions below. Do you think the spec still requires clarification?

chrispsommers:
@stefanheule Please see answers to your questions below. Do you still feel the spec requires clarification?
>It is not clear what kind of nested expressions are allowed, and it is also not clear if nested expressions can themselves be nested.

This is actually specified quite precisely in the grammar portion of the spec. An expression can be of the form {expression, expression..} i.e. a nested expression.

>Is {1,2,{1,2,{1,2}}} okay?

Yes! (I added an example to https://github.com/p4lang/p4runtime/pull/265/ which prompted this discussion.)

> Is {a=1,b=2} a valid "compile-time constant" expression?

Not according to the grammar for an expression. This looks more like an assignment statement.

> Another question I had was if there is any kind of type checking? Can list elements be of different types?

According to the grammar, each element in an expression list can be a different type.

Bear in mind P4Runtime will impose its own constraints on Structured Annotations and it is a policy to not specify those constraints in the P4 language spec itself. So you could write something in "legal p4-16" which would throw an error in the P4Info generator (generating P4Info is a `p4c` command-line option  so the program can still be "compiled").  For example, an expression list containing nested kvPairs would be OK in P4-16 but not for P4Info.

stefanheule:
> This is actually specified quite precisely in the grammar portion of the spec. An expression can be of the form {expression, expression..} i.e. a nested expression.

The spec does not say what is and is not a compile-time constant, this needs to explicitly defined in my opinion. Compile-time constant probably makes sense for base types like int, but it less clear for nested types. Can we just explicitly list and say what is allowed?

To give a few examples of why this is necessary:
- I asked if `{a=1,b=2}` is a valid compile-time constant (it's a kvPair expression), and you said no, but it's not clear to me why not. It is an expression (in the expression grammar: `'{' kvList '}'`), and it only has constants.
- What about `a[3]` if `a` at compile time is known to be a fixed array?
- What about `{1,2}[1]`? I think this is legal P4, at least according to the grammar.

> Bear in mind P4Runtime will impose its own constraints on Structured Annotations and it is a policy to not specify those constraints in the P4 language spec itself. So you could write something in "legal p4-16" which would throw an error in the P4Info generator (generating P4Info is a `p4c` command-line option so the program can still be "compiled"). For example, an expression list containing nested kvPairs would be OK in P4-16 but not for P4Info.

I don't think this is a great idea. If we want to restrict what is and isn't valid, lets be clear about that and put it in the spec. There can be exceptions, of course, like requiring ints to fit in a certain bit-width, but we should not have a completely different set of expressions be valid in P4 and P4Info. What's the advantage of allowing nested kvPairs in the spec but not P4Info? Isn't P4Info the main use-case for this?

Finally, a last question: Are we going to be able to implement this spec easily in the compiler? Does the compiler have code to constant-fold nested expressions?

jnfoster:
@stefanheule 

Here is the relevant [section](https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-ct-constants) of the specification. Happy to see it improved if you see things that are missing or confusing.

stefanheule:
Ah, thank you Nate, I missed that. That answers most of my concerns then, but maybe it needs to be improved a bit. Is `{a=1,b=2}` a compile-time constant according to that definition? It seems no, but it also seems like `const a = {a=1,b=2}` makes `a` a compile-time constant? It would be good to have clarity on that.

And maybe Chris can comment on why there is a discrepancy in the allowed expressions in P4 vs P4Info if there still is any (other than max bit width).

chrispsommers:
@stefanheule

chrispsommers:
@stefanheule Regarding your comment: 
>I don't think this is a great idea. If we want to restrict what is and isn't valid, lets be clear about that and put it in the spec. There can be exceptions, of course, like requiring ints to fit in a certain bit-width, but we should not have a completely different set of expressions be valid in P4 and P4Info. What's the advantage of allowing nested kvPairs in the spec but not P4Info? Isn't P4Info the main use-case for this?

Your objection to having the language spec allow more complexity of expressions than P4Runtime is understandable. In fact, at one time I'd added content to p4-16 spec which was more specifically-tailored to P4Info on an earlier version of structured annotations: a special case of  kv-pairs inside `()`.  I had the same inclination to state it all in the language spec. But the language experts on the WG wanted the spec to remain more "pure" from P4-Runtime and it was removed (later this became moot because we removed all structure from regular annotations). So I'm following the "party line" here as I interpret it. Feel free to make your case otherwise, it's an open forum.

> And maybe Chris can comment on why there is a discrepancy in the allowed expressions in P4 vs P4Info if there still is any (other than max bit width).

It's a convenience that structured annotations are capitalizing on existing grammar for expressions, and thereby also benefits from the implementation which parses, checks and reduces expressions. This doesn't mean we have to support every permutation for P4Info.

There was a sentiment expressed in some earlier P4-API WG meetings to start out with simple support of structured annotations for known use-cases. There was concern that making it too complex would only delay implementation and multiply the work enormously (imagine all the test-cases we'd have to create). Since nobody made a case for arbitrarily-nested kv-lists of kv-lists of expression-lists (a contrived example), those cases aren't supported. We just stuck with something simple and immediately useful.

stefanheule:
To me that sentiment extends to the P4 spec as well, though. Otherwise, if you take the P4Info doesn't have to support everything that P4 supports to the extreme, then there is no need for a new kind of annotations at all, and you can just use the new unconstrained annotations, but have P4Info only support what further parses into whatever format we want to support.

If we want to disallow nested kv-lists (which seems like a reasonable idea), then we should say so in the P4 spec. If we want to allow it, we should allow it in P4Info, too.

jafingerhut:
@stefanheule I am not here with any necessarily conclusive arguments, but just a couple of observations.

Think of this from the perspective of someone maintaining the P4 language specification and the parts of the p4c implementation (grammar, front/mid-end code, etc., all of it except the part that generates the P4Info file -- the part that generates the P4Info file is pretty small compared to the rest of it).  Would you want to:

(a) define and implement the most general thing that P4Runtime API people want, including possibly things they do not want now, in this 'container' (i.e. the annotations being discussed)?  Or

(b) would you want to lock it down to exactly what P4Runtime API people say they want today, even if they might want to change it in several kinds of detailed ways N times in the next 3 years?

I think the folks in the P4 language work group are thinking of this where they prefer an approach more like (a), and hope that any changes to the constraints of what is legal/supported/etc. by the P4Info folks can be isolated to that small part of p4c that generates the P4Info file, and/or to consumers of that P4Info file, and require no changes to anything else.  It is a kind of separation of concerns.

I understand such a separation of concerns can also feel like putting up walls between things that need not have walls between them.  I just wanted to write down my interpretation of the preferences you are encountering (Note: this interpretation is my own.  I don't claim to speak for anyone else here.)

stefanheule:
Thank Andy, I think this is a useful way to look at it, and I also do not claim to have a any definite answers.

Two thoughts:
1. It seems to me that deciding what structures annotations should look like is not just a P4RT/P4Info issue, I would think this is a P4 issue in general, and so we here in the P4 LDWG community should make a decision what we want these to look like, not just go with whatever the P4RT people want. But maybe I'm the only one who feels that way?
2. If we go with an extreme version of (a), then there is nothing to do here, because the unstructured annotations already can do anything the P4RT people want (because they can do basically *anything*). Since we are talking about this here, we have decided that we want some version of (b), I guess now the question is just how much (b) we want.

I think it would be good to specify clearly what structured annotations are allowed to be and have P4Info and P4 largely match, but if everyone else feels differently that is fine too.

jafingerhut:
Perhaps part of the reason for the separation of concerns is that from the point of view of some in the P4 community, the P4Runtime API is one among several control plane APIs that exists.  Others were devised before it existed, and continue to be supported even after the P4Runtime API has been created.  (e.g. the Thrift API implemented by the simple_switch code, and probably at least one proprietary one implemented by each P4 device vendor)

jafingerhut:
Granted that the P4Runtime API is the only control plane API developed under a P4.org working group, so one could reasonably ask whether it should have "special privileges" that the others do not.  It might be the case that these new structured annotations might find their way to be used to augment other proprietary control plane APIs, too, and certainly there the P4 language work group's desire would be to allow what can be allowed, and not be restricted by the lowest common denominator of what different ones want there.

stefanheule:
I agree that P4RT should not be treated as the only control plane API, which is why I feel more strongly that what structured annotations look like, which are legal, how they can and cannot be nested is something that should be decided at the P4 level.  This will allow different control planes to all implement the same support for what is in P4, instead of P4 being overly permissive and then every control plane API deciding to support a slightly different subset.

jafingerhut:
@stefanheule wrote: "If we want to disallow nested kv-lists (which seems like a reasonable idea), then we should say so in the P4 spec. If we want to allow it, we should allow it in P4Info, too."

Hmm.  Looking at the grammar, the syntax `{a=1,b=2}` was made legal in the spec's grammar with this commit: https://github.com/p4lang/p4-spec/commit/8bfa001d4ed5d6d52af71f1c90ac26c46e2d1c1e for the purposes of creating a syntax that represents the entire value of one whole struct, as an expression.

So I believe if inside of a structured annotation that `{a=1,b=2}` was a value, it maybe is expected that it is not another annotation, but the value of a struct.  If all of the values in that expression are compile-time constants, then it is not terribly explicit now, but it seems to be in the spirit of the other compile-time constant expressions, and perhaps could be explicitly included there for clarity.

Stefan, does that answer your question about whether kv-lists can be nested?  I see the answer as "it might look like it in the syntax, but that thing inside the structured annotation is a struct, not another kv-pair".

stefanheule:
I agree that we should clarify whether `{a=1,b=2}` is a compile-time constant or not, right now the spec is ambiguous.

You say "it might look like it in the syntax, but that thing inside the structured annotation is a struct, not another kv-pair", but I'm not sure I understand what that means.  Can kv-lists be nested or not? In the context of P4 in general vs within structured annotations? Does the spec need clarification?

jafingerhut:
@stefanheule I will try to be more careful in my wording here, as I may be confusing myself, and taking you with me along for the ride :-)

When you say "can kv-lists be nested or not?", at least part of the answer, even without any change to the language spec that adds structured annotations, is that I believe the program attached to this issue I just created https://github.com/p4lang/p4c/issues/2213 is a legal P4_16 program, including this line:
```
        meta.s2 = {s1={f1=2, f2=3}, f3=5, f4=8};
```
I may be wrong about that, given that p4c does not support it right now, but we will see in the resolution of the issue.

If we assume for the moment that program is legal, it is a nested struct initializer, i.e. one struct literal value, nested inside of another.  There are no structured annotations in that code snippet.

So my guess is this:

+ struct initializer expressions are already a defined part of the language
+ I think they should be allowed to nest inside of each other.  We will see based upon issue I linked above.
+ If structured annotations are added to P4, I think they should be allowed to contain struct initializer expressions, as long as they contain only compile-time constant field values.  That is, it would make sense to me personally if a struct initializer expression where all field values are compile time constants, should itself be a compile time constant, too.
+ If you are asking "should a structured annotation be allowed to nest inside of another structured annotation", then I do not know what nesting of annotations would even mean.  Do you think annotations should nest inside of annotations, and if yes, what should it mean?  Do you have a use case for it?

jafingerhut:
Oh, and. bullet point I forgot, but is probably obvious from the context:

+ struct initializer expressions syntactically look very much like the proposed syntax of structured annotations.  Yes, this might be potentially confusing to some.

jafingerhut:
Created a separate small issue here asking about explicitly mentioning whether struct initializer expressions should be documented as compile-time known values: https://github.com/p4lang/p4-spec/issues/816

stefanheule:
> struct initializer expressions syntactically look very much like the proposed syntax of structured annotations. Yes, this might be potentially confusing to some.
Is that true? A struct initializer expression looks like this, right? `{a=1, b=2}`. But the proposed syntax is `@Name[a=1, b=2]`.  If we allow nesting of struct initializer expressions, then you could also write `@Name[a=1, b={a=1, b=2}]`.

Another question if we allow nesting: On the top-level, the spec requires keys not to be reused, i.e. `@Name[a=1, a=2]` is invalid. What about nested expressions, can keys be repeated there? I'm not sure if P4 says anything about that.

jafingerhut:
@stefanheule wrote: "Is that true? A struct initializer expression looks like this, right? {a=1, b=2}."

Yes, it does.

"But the proposed syntax is @Name[a=1, b=2]. If we allow nesting of struct initializer expressions, then you could also write @Name[a=1, b={a=1, b=2}]."

Sorry, my mistake on the proposed syntax of structured annotations, which as you say is proposed to use square brackets, not curly braces.  They don't look quite as similar as I was misremembering.

"Another question if we allow nesting: On the top-level, the spec requires keys not to be reused, i.e. @Name[a=1, a=2] is invalid. What about nested expressions, can keys be repeated there? I'm not sure if P4 says anything about that."

When you say "if we allow nesting", I am going to assume you mean "if we allow nesting of structure initializer expressions inside of structured annotations".  If am wrong in that assumption, please correct me.

I am assuming that if we allow structure initializer expressions inside of structured annotations, then it is because someone wants to pass the value of a structure initializer expression as the value of some key in the structured annotation.  The keys in the structure initializer expression would never 'mix' with the keys of the structured annotation in any way.  They are in different 'scopes'.

jafingerhut:
stefanheule If your question is "In a structure initialization expression like {a=1, b=2}, is it legal/illegal/undefined/etc. to repeat keys, e.g. {a=1, b=2, a=3}?", then I agree that question is not explicitly addressed in the language spec.  If it were addressed, that seems to best belong in the section on structure initializers, not in the definition of structured annotations.   I have created a separate spec issue here to raise that question: https://github.com/p4lang/p4-spec/issues/817

stefanheule:
Thanks Andy, I really appreciate you being so thorough.  Yes, I was talking about the question your raised in #817, and I agree that static initializer expressions and annotations have different scopes, I wasn't trying to say they are mixed.

Taking a step back, the goal of structured annotations was to have annotations that are easier to understand and consume by humans and tools, restricting them more than the current annotations. They can only be a list or set of kv-pairs, and every expression has to be a compile-time constant. Now one question is whether we want to allow all compile-time constants or not.  To me, allowing nested constants (i.e. lists, and struct initializers) seem like a quite complicated addition to structured annotations (violating the design goal to make this easy for humans and tools to consume), and it is not clear to me how useful they would be.
1. So one sensible proposal would be to say that structured annotations are a list or set of kv-pairs, where all expressions are *base-type* (int, bool, string) compile-time constants.
2. Another way is to say expressions can be any compile-time constant (including lists and struct initializers), and then expect P4Info (and any other consumer) to support that full generality.

At the moment the the proposal (here and in P4RT) seems to be to do 2., but not have P4Info support all of 2. That seems bad. If the point of structured annotations is to make it easy to consume, we should not have the first consumer (P4Info) already give up and only support a subset.

jafingerhut:
Thanks for writing up that good summary :-)

I think a good way of arguing for this with the language design group is to refer to that summary, and propose that the language spec start by only allowing things in structured annotations that we know "a customer" (i.e. the P4Runtime API working group) wants, and that in general it is often good to start by allowing only a known-useful list of cases (types of values in this case), and leave it open to future changes to the language spec to extend it as and if needed.

stefanheule:
We had a discussion about this in the P4RT meeting today, and everyone there would be happy with 1. Before Chris makes all these changes, I wanted to see if anybody here would object to that.

Specifically, that means, structured annotations contain either a list of values, or a list of kv-pairs, where the values in both cases are any compile-time constants of a base type (bool, int, string).  There is no nesting.  This is to keep with the goal of giving more structure to this new kind of annotations, and make it easier for them to be consumed by humans/tools downstream.

Allowed would be, for instance:
- `@Name[1,2,3]`
- `@Name[1, "foo"]`
- `@Name[a=2, b=3+2, c="foo"]`

Not allowed would be:
- `@Name[identifier]`
- `@Name[a=1, 2]`
- `@Name[3+identifier]`
- `@Name[a={b=1}]`
- `@Name[{b=1}]`
- `@Name[a={1,2,3}]`

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7097,16 +7100,155 @@ annotationBody
...
+`expressionList` may be empty or contain a comma-separated list of member
+`expression`s. A `kvList` consists of one or more `kvPair`s, each consisting of
+a key and a value `expression`. Note the syntax for `expression` is rich, see
+Appendix [#sec-grammar] for details.
+
+To avoid ambiguity, an empty `stucturedAnnotationBody` is by definition
+considered to contain an empty `expressionList`.
+
+All `expression`s within a `stucturedAnnotationBody` must evaluate at
+compile-time to one of the following base types: string literals,
```
stefanheule: I would phrase this as "... must evaluate to a compile-time constant of the following types: string, infinite-precision integer, or boolean".

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7097,16 +7100,155 @@ annotationBody
...
+infinite-precision integers or booleans. Nested `expression`s (e.g. an
+`expression` containing an `expressionList`, a `kvList`, etc.) are not
+allowed. Note that P4Runtime information (`P4Info`) may stipulate additional
+restrictions. For example, an integer expression might be limited to 64-bit
+values.
+
+It is illegal to duplicate a key within an annotation.  It is illegal to use the
```
stefanheule: Maybe "It is illegal to duplicate a key within the key-value list of a structured annotation."?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7097,16 +7100,155 @@ annotationBody
...
+same annotation `name` on a given `annotationToken`.
```
stefanheule: "... same annotation `name` on a given `annotationToken` for structured annotations". Since I assume this is only a restriction we have for structured annotations.

p4-16/spec/P4-16-spec.mdk
jafingerhut: In another recent discussion, I was reminded that the term that the language spec maintainers would like to use for this is "compile-time known value", so if a change is made here, perhaps a desired change would be to replace this:

"All `expression`s within a `stucturedAnnotationBody` must evaluate at
compile-time to one of the following base types: string literals, infinite-precision integers or booleans."

with this:

"All `expressions within a `stucturedAnnotationBody` must be compile-time known values, either literals or expressions requiring compile-time evaluation, with a result type that is one of: string literals, infinite-precision integer, or boolean."

chrispsommers: Done, thanks.

chrispsommers:
@liujed Regarding https://github.com/p4lang/p4-spec/pull/796#issuecomment-566221065 and the following text in  the PR:
>To avoid ambiguity, an empty `structuredAnnotationBody` is by definition
considered to contain an empty `expressionList`.

liujed-intel:
> @liujed Regarding [#796 (comment)](https://github.com/p4lang/p4-spec/pull/796#issuecomment-566221065) and the following text in the PR:
> 
> > To avoid ambiguity, an empty `structuredAnnotationBody` is by definition
> > considered to contain an empty `expressionList`.
> 
> The P4-API WG is proposing we remove this statement from the spec. The current PR [p4lang/p4c#2227](https://github.com/p4lang/p4c/pull/2227) doesn't really do this (empy anno body results in empty kvLists and expressionLists in the IR) and P4Info would be burdened with generating an empty expressionList to remain consistent with spec clause above.
> 
> Any objections/concerns from a language/grammar point of view?

This seems fine to me. I can't immediately think of any cases where it's worth distinguishing between the two kinds of empty lists.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7097,16 +7100,157 @@ annotationBody
...
+evaluation, with a result type that is one of: string literals,
+infinite-precision integer, or boolean. Nested `expression`s (e.g. an
+`expression` containing an `expressionList`, a `kvList`, etc.) are not
+allowed. Note that P4Runtime information (`P4Info`) may stipulate additional
+restrictions. For example, an integer expression might be limited to 64-bit
+values.
+
+It is illegal to duplicate a `key` within the `kvList` of a structured
+annotation.  It is illegal to use the same annotation `name` on a given
+`annotationToken` for structured annotations.
```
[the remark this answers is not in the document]
chrispsommers: It means you can't reuse a given structured annotation name, i.e.  no duplicates. The spec shows an example of an illegal duplicate.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7097,16 +7100,157 @@ annotationBody
...
+Appendix [#sec-grammar] for details.
+
+To avoid ambiguity, an empty `structuredAnnotationBody` is by definition
+considered to contain an empty `expressionList`.
+
+All `expression`s within a `structuredAnnotationBody` must be compile-time
+known values, either literals or expressions requiring compile-time
+evaluation, with a result type that is one of: string literals,
+infinite-precision integer, or boolean. Nested `expression`s (e.g. an
+`expression` containing an `expressionList`, a `kvList`, etc.) are not
```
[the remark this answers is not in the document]
chrispsommers: Sure, seems like a good suggestion.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
liujed-intel: Ah, then I think "`annotationToken`" is the wrong word to use here. Maybe something like "program element"?

liujed-intel:
What happens if we have something like:
```
@my_anno("foo")
@my_anno["bar"]
table t { ... }
```

chrispsommers:
@liujed That's an example of a duplicate error.

mbudiu-vmw:
The spec and the implementation do not make this clear.

chrispsommers:
@mbudiu-vmw I'm not sure what isn't clear here is the spec text:

p4-16/spec/P4-16-spec.mdk
liujed-intel: As written, this sentence makes no sense: an annotation token is a part of the body of an unstructured annotation.

It should also be made more explicit that it is a program error to have a structured annotation and an unstructured annotation with the same name on a table, control, etc. I interpreted the current wording as just saying you can't have two structured annotations with the same name.

chrispsommers:
regarding https://github.com/p4lang/p4-spec/pull/796#issuecomment-597917976 I made a mistake reading it.
```
@my_anno("foo")
@my_anno["bar"]
table t { ... }
```

liujed-intel:
I remember that discussion. But I don't think it touched on whether it made sense to mix structured and unstructured instances of an annotation. Naïvely, I'd expect an annotation to be structured or unstructured; having both seems to be a programmer error. Absent any strong use case for mixing the two kinds of annotations, I'd be inclined to do the conservative thing and forbid it.

chrispsommers:
@liujed-intel thanks for the comments. @mbudiu-vmw Can you comment on https://github.com/p4lang/p4-spec/pull/796#issuecomment-598874750? I'm OK with disallowing using the same name for unstructured and structured annotations. How does the implementation deal  with this, or does it?

mbudiu-vmw:
I implemented both versions at some point. Currently we only reject structured duplicates.

chrispsommers:
I pushed commits to hopefully resolve the above conversations, thanks for all the detailed reviews and feedback. @liujed please review, @mbudiu-vmw this will require a change to  `p4c` to disallow reusing an annotation name between structured and unstructured. Sorry for the reversal, it seems for the best.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7078,17 +7084,18 @@ annotations
 annotation
     : '@' name
     | '@' name '(' annotationBody ')'
+    | '@' name '[' structuredAnnotationBody ']'
```
mbudiu-vmw: Perhaps you can use the grammar from the compiler implementation ; it essentially inlines the structuredAnnotationBody in this rule.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7058,11 +7058,17 @@ diagnostics.
...
+mechanism for extending the P4 language to some limited degree without changing
+the grammar. To some degree they subsume C `#pragma`s. Annotations are attached
+to types, fields, variables, etc. using the `@` syntax (as shown explicitly in
+the P4 grammar). Unstructured annotations, or just "annotations," have an
+optional body; structured annotations have a mandatory body, containing at least
+a pair of square brackets `[]`.
+
+Structured annotations and  unstructured annotations must not use the same
+`name`. Thus, a given `name` can only be applied to one type of annotation or
+the other.
```
liujed-intel: This sounds like it might forbid the following, but it's not clear whether that is the intent. We should clarify and ensure that the implementation matches.
```
@my_anno(1) table T { ... }
@my_anno[2] table U { ... }
```

p4-16/spec/P4-16-spec.mdk
liujed-intel: If this suggestion is taken, be sure to patch up those places where the text refers to `structuredAnnotationBody`.

stefanheule: What is the advantage of inlining here? I think I like that we have a name for structured annotation bodies.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7097,16 +7099,157 @@ annotationBody
...
+    | expressionList ',' expression
+    ;
+...
+kvList
+    : kvPair
+    | kvList ',' kvPair
+    ;
+
+kvPair
+    : name '=' expressionList
```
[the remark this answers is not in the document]
stefanheule: I agree, this should definitely be expression. We don't actually allow a list in this position.

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: The spec grammar is almost identical to the bison grammar, because then we know it works. If we deviate it may actually not be implementable. Bison is very peculiar.

chrispsommers:
@liujed-intel thanks for pointing this out, it is an ambiguity. The spec should read:
Structured annotations and  unstructured annotations on any one element  must not use the same
`name`. Thus, a given `name` can only be applied to one type of annotation or
the other for any one element. An annotation used one one element does not affect the annotation on another because they have different scope.

This is legal:
```
@my_anno(1) table T { ... }
@my_anno[2] table U { ... }
```
This is illegal:
```
@my_anno(1)
@my_anno[2] table U { ... }
```
Does this make sense?

chrispsommers:
@liujed-intel re: 
>We seem to have lost the text that specified that a program element can't have multiple structured annotations with the same name.

I propose to add the following paragraph:

Multiple unstructured annotations using the same `name` can appear on a given
element; they are cumulative. Each one will be bound to that element. In
contrast, only one unstructured annotation using a given `name` may appear on an
element; multiple uses of the same `name` will produce an error.

chrispsommers:
@mbudiu-vmw @stefanheule re: https://github.com/p4lang/p4-spec/pull/796#discussion_r363454903:

The latest PR contains the following and has for awhile, unless I'm mistaken:
```
kvPair
    : name '=' expression
    ;
```

chrispsommers:
@mbudiu-vmw re: https://github.com/p4lang/p4-spec/pull/796#discussion_r398215825 I'm not confident interpreting this suggestion; this grammar has been in my PR for a while. Perhaps we can merge my PR and someone else can propose restating the grammar.

liujed-intel:
Re: https://github.com/p4lang/p4-spec/pull/796#issuecomment-604717022

Yes, makes total sense, and this was also my thinking. But now that the question is being raised, I wonder whether it makes sense to do the stricter thing instead. At least, that would seem to be more consistent with the restriction that you can't mix structured and unstructured for the same name on a single program element. If you consider all annotations with the same name as expressions of the same concept, then I'd expect consistency with structured vs. unstructured in their usage. Thoughts?

chrispsommers:
@liujed-intel re: https://github.com/p4lang/p4-spec/pull/796#issuecomment-604723569 I know what you mean, but I'd err on the side of letting the author write confusing code. It's no differet than letting them apply the same unstructured annotation but imply a different meaning from place to place. Scoping ensures correct compilation but we can't always prevent people from writing bad code.

PR #797, opened 2019-12-10
Introduce default values

mbudiu-vmw:
Note that currently default values are not used anywhere in the spec. The intention is to have an additional pull request that will specify when or how default values can be used. But I think that this part will be less controversial so I factored it out separately.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4529,6 +4529,51 @@ direction `in` or `inout` are initialized when the control is called,
 and under what conditions they are initialized, and if so, what their
 values will be.
 
+## Initializing with default values
+
+A left-value can be initialized automatically with a default value of
+the suitable type using the syntax `...`.  A value of type `struct`,
+`header`, or `tuple` can also be initialized using a mix of explicit
+values and default values by using the notation `...` in a list
+expression initializer; in this case all fields not explicitly
+initialized are initialized with default values.  When intializing a `struct`,
+`header`, and `tuple` with a value containing partially default values
+using the `...` notation the three dots must appear last in the initializer.
+
+~ Begin P4Example
+struct S {
+     bit<32> b32;
+     bool b;
+}
+
+enum int<8> N0 {
+   one = 1,
+   zero = 0,
+   two = 2
+}
+
+enum N1 {
+     A, B, C, F
+}
+
+struct T {
+    S s;
+    N0 n0;
+    N1 n1;
+}
+
+N0 n0 = ...;  // initialize n0 with the default value 0
+N1 n1 = ...;  // initialize n1 with the default value N1.A
+S s0 = ...;  // initialize s0 with the default value { 0, false }
+S s1 = { 1, ... };  // initialize s1 with the value { 1, false }
+S s2 = { b = true, ... }; // initialize s2 with the value { 0, true }
+T t0 = ...;  // initialize t0 with the value { { 0, false }, 0, N1.A }
+T t1 = { s = ..., ... }; // initialize t1 with the value { { 0, false }, 0, N1.A }
+T t2 = { s = ... }; // error: no initializer specified for fields n0 and n1
+tuple<N0, N1> p = { ... }; // initialize p with default value { 0, N1.A }
+T t3 = { ..., n0 = 2}; // error: ... must be last
+~ End P4Example
+
 # Function declarations { #sec-functions }
 
 Functions can only be declared at the top-level and all parameters must have a direction.
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1341,7 +1341,7 @@ In this document we adopt the following conventions:
 
 - Built-in types are written with lowercase characters---e.g., `int<20>`,
 - User-defined types are capitalized---e.g., `IPv4Address`,
-- Type variables are always uppercase---e.g., `parser P<H, IH>(...)`,
+- Type variables are always uppercase---e.g., `parser P<H, IH>()`,
 - Variables are uncapitalized--- e.g., `ipv4header`,
 - Constants are written with uppercase characters---e.g., `CPU_PORT`, and
 - Errors and enumerations are written in camel-case--- e.g. `PacketTooShort`.
@@ -1391,7 +1391,7 @@ Some P4 constructs act as namespaces that create local scopes for names includin
   declaration:
 
 ~ Begin P4Example
-extern E<H>(...) { ... } // scope of H ends here.
+extern E<H>(/* parameters omitted */) { /* body omitted */ } // scope of H ends here.
 ~ End P4Example
 
 The order of declarations is important; with the exception of parser
@@ -1670,10 +1670,10 @@ switch architecture where the second stage is optional. This could be
 declared as a package with an optional parameter:
 
 ~Begin P4Example
-package pipeline(...);
+package pipeline(/* parameters omitted */);
 package switch(pipeline first, @optional pipeline second);
-...
-pipeline(...) ingress;
+
+pipeline(/* arguments omitted */) ingress;
 switch(ingress) main;   // a switch with a single-stage pipeline
 ~End P4Example
 
@@ -1882,7 +1882,7 @@ the following annotation indicates that a specific name should be used
 for a table when generating the control-plane API:
 
 ~ Begin P4Example
-@name("acl") table t1 { ...}
+@name("acl") table t1 { /* body omitted */ }
 ~ End P4Example
 
 ### Integers (signed and unsigned) { #sec-integers }
@@ -2177,7 +2177,6 @@ enum bit<16> EtherType {
   MPLS      = 0x8847,
   IPV4      = 0x0800,
   IPV6      = 0x86dd
-  // ...
 }
 ~ End P4Example
 
@@ -2432,8 +2431,8 @@ current scope. An empty struct is legal. For example, the structure `Parsed_head
 below contains the headers recognized by a simple parser:
 
 ~ Begin P4Example
-header Tcp_h { ... }
-header Udp_h { ... }
+header Tcp_h { /* fields omitted */ }
+header Udp_h { /* fields omitted */ }
 struct Parsed_headers {
     Ethernet_h ethernet;
     Ip_h       ip;
@@ -2450,7 +2449,7 @@ type of tuples with n component types `T1`,...,`Tn` is written
 as
 
 ~ Begin P4Example
-tuple<T1,...,Tn>
+tuple<T1,/* more fields omitted */,Tn>
 ~ End P4Example
 
 ~ Begin P4Grammar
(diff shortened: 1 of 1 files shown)


jafingerhut:
I guess the situations in which these default values will actually be used/assigned are yet to be specified?  As written, someone might get the wrong impression that these default values are assigned to struct member fields, for example, whenever the user program does not explicitly do so, like Java does.

mbudiu-vmw:
Yes, the comment on the PR says that.

mbudiu-vmw:
I have added a second commit to this PR which proposes a syntax for default intializers. The two commits should be reviewed separately, but the second one depends on the first one.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4465,6 +4465,51 @@ direction `in` or `inout` are initialized when the control is called,
...
 values will be.
 
+## Initializing with default values
+
+A left-value can be initialized automatically with a default value of
+the suitable type using the syntax `...`.  A value of type `struct`,
+`header`, or `tuple` can also be initialized using a mix of explicit
+values and default values by using the notation `...` in a list
+expression initializer; in this case all fields not explicitly
+initialized are initialized with default values.  When intializing a `struct`,
```
jafingerhut: typo on "initializing"

jafingerhut:
The second commit looks good to me.  The set of examples looks like it includes corner/error cases that I had wondered about, including one with a nested struct/header inside of another, so nice!

jafingerhut:
Is it stated elsewhere, or should it be stated somewhere in the text of the second commit, whether it is allowed, or an error, to mix positional and named fields in a list expression?

vgurevich:
I love the first part, i.e the clear definition of the default values,  but i still do not see the solution to my main problem in the second part,  where it looks like I can either initialize the whole structure with the default value or I have to initialize each field explicitly,  whereas there is a strong desire to initialize only some fields explicitly and leave the rest initialized with the default values.  Am I missing something or this is not doable?

jafingerhut:
@vgurevich. There are examples in the proposed changes that show how to initialize several fields of a header or struct via "positional" specification of the fields values, in order, followed by `...` to indicate "the rest all have default values after the last one I specified".

You can also specify fields to initialize by "name1 = value1, name2=value2, ...", and if you put `...` at the end, then the rest of the fields you did not specify are given default values.

Is there some capability you see missing there that you wanted?

mbudiu-vmw:
This issue about mixing named and positional fields in a struct initializer is not new, it was introduced when we added initializers with named fields. So I didn't write anything about it in this PR. We actually have two types of expressions: list expressions and struct initializers, there is no hybrid which allows a mix. This issue is similar to function calls with named parameters. I think we should disallow mixes, since explaining what is legal is complicated, and it's a virtue to keep things as simple as possible.

vgurevich:
I think that my main quip is about the fact that in order to have "future-proof" initializers in the code, people will always have to put "..." in it. 

In other words, the following code is "future-proof" in C
```
typedef struct foo {
    int bar;
    int baz;
} foo_t;

foo_t ggg = { 1, 5 }; /* Adding more fields to the end won't change the result */
foo_t hhh = { .bar=1, .baz=5 ;};      /* Adding fields anywhere in foo_t won't change the result */
```
In P4 the equivalent "future-proof"  code is: 
```
foo_t = { .bar=1, .baz=5,  ... };
```
meaning that people will get into a habit of putting the ellipsis in all structure initializers. Mixing named/unnamed fields adds a little pain, but I think that's survivable, although I do like the Python rule more (named an unnamed items can be mixed as long as the unnamed ones are only in the beginning and do not follow the named ones).

jafingerhut:
There is a tradeoff of concerns here.

You seem to prefer that unmentioned fields get default values, with the perceived advantage of not having to update initializers when new fields are added to a struct/header.

Others might see that as a disadvantage, and would _prefer_ to get a compiler error message if they did not write the `...`, so the compiler reminds them that they forgot to add an value for a field in some assignment somewhere.

Which of those is considered better behavior, or having one be the default with a compiler option for the other, is part of language design, and it isn't clear to me what most users of the language might prefer.  My own personal preference would be for the second choice, but it doesn't sounds like a life and death difference to me.

vgurevich:
@jafingerhut  -- this is certainly not a life-and-death situation. I am simply pointing out that a common usage pattern required for modular design requires mandatory typing of some extra characters. The choice can go either way.

Also, do we allow "_" in the initializers, meaning that we can ask some fields to *not* be initialized while supplying the whole structure?

mbudiu-vmw:
We do not allow `_` in initializers. You can do this by explicitly assigning to the fields you want. If you can show us some programs that do this we can consider this feature.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2795,6 +2795,37 @@ important what the type variable binds to (during type unification the
...
+* For `bool` the default value is `false`.
+* For `error` the default value is `error.NoError` (defined in core.p4)
+* For `string` the default value is the empty string `""`
+* For `varbit<N>` the default value is a string of zero bits (there is currently
+   no P4 literal to represent such a value).
+* For `enum` values with an underlying type the default value is 0,
+  even if 0 is actually not one of the named values in the enum.
+* For `enum` values without an underlying type the default value is the
+  first value that appears in the `enum` type declaration.
+* For `header` types the default value is `invalid`.
```
marko-stanoj: As far as I can see it is impossible to set header validity bit to zero with existing methods of initialization (list expression or struct initializer expression).  Do you find it convenient that you can do it with shorten notation of existing initializations(`= ...`), and with regular ones you cannot?

jafingerhut: It is my understanding (corrections welcome if I am wrong here) that nothing in this proposal enables another way to assign an invalid header value to a header.  Whether this proposal goes forward as it is now, or not, these are the only ways I know of for a header to become invalid:

+ You make a call to the method `myhdr.setInvalid()`
+ Assign to a header, the value of another variable with the same header type, and that variable is currently an invalid header.
+ direction `out` parameters with a header type are defined to be initialized to invalid, without any explicit statement in the program that does so. (I am not sure whether one can define an `out` parameter with a header stack type -- not something I have tried before or seen brought up).
+ header stack push_front and pop_front operations can change which elements of the header stack are valid or invalid.

jafingerhut: @marko-stanoj Are you asking "Would others like to introduce another way to make a header invalid, other than the ones that exist now, which uses an initialization expression?"

marko-stanoj: Later in this document it's said:

> A left-value can be initialized automatically with a default value of
> the suitable type using the syntax `...`

so if left-value is header type (or struct type that contains header type as a field) it should be initialized to it's default value. Due to this spec proposal, the default value is invalid, or am I missing something?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4434,6 +4465,51 @@ direction `in` or `inout` are initialized when the control is called,
...
+N1 n1 = ...;  // initialize n1 with the default value N1.A
+S s0 = ...;  // initialize s0 with the default value { 0, false }
+S s1 = { 1, ... };  // initialize s1 with the value { 1, false }
+S s2 = { b = true, ... }; // initialize s2 with the value { 0, true }
+T t0 = ...;  // initialize t0 with the value { { 0, false }, 0, N1.A }
+T t1 = { s = ..., ... }; // initialize t1 with the value { { 0, false }, 0, N1.A }
+T t2 = { s = ... }; // error: no initializer specified for fields n0 and n1
+tuple<N0, N1> p = { ... }; // initialize p with default value { 0, N1.A }
+T t3 = { ..., n0 = 2}; // error: ... must be last
+~ End P4Example
```
jafingerhut: Would adding these examples be consistent with the rest of the proposal?  It seems it would to me, but I may be missing something:
```
header H {
    bit<16> f1;
    bit<8> f2;
}

H h1 = ...;   // initialize h1 with a header that is invalid
H h2 = { f2=5, ... };   // initialize h2 with a header that is valid, field f1 0, field f2 5
H h3 = { ... };  // initialize h3 with a header that is valid, field f1 0, field f2 0
```

jafingerhut: And if `H h1 = ...` is intended to be valid syntax, does having a right hand side containing only `...` apply for all types with initial values, e.g. `bit<32> x = ...` initializes `x` to `32w0` ?

If so, may as well have at least one example like that, too.

jafingerhut: And if those examples with only `...` on the right hand side are not intended to be legal, it seems best to add some qualification saying so to this proposed new sentence in this PR: "A left-value can be initialized automatically with a default value of  the suitable type using the syntax `...`."

p4-16/spec/P4-16-spec.mdk
jafingerhut: I hereby withdraw this statement in an earlier comment: "It is my understanding (corrections welcome if I am wrong here) that nothing in this proposal enables another way to assign an invalid header value to a header."

Now upon reading this proposed change, it does appear to me that the syntax `h1 = ...;` is proposed to make a header variable `h1` become invalid, no matter what its state was before, and syntax `h1 = {f1=5, ...}` is to make `h1` valid, with field `f1` becoming 5, and all other fields being assigned their default values, depending upon the type of those fields.

In other comments on this PR, I propose adding explicit examples with comments explaining this, rather than leaving it open to the imagination of the reader.

mbudiu-vmw:
I propose we merge this PR in the spec even without having a full implementation yet.

jafingerhut:
I am perfectly happy merging in the spec change without having a complete implementation.

I do think it would be good to add the examples mentioned in some of my earlier comments, but I can add them with a separate PR after this one is merged, if that is easier.

mbudiu-vmw:
I have amended this PR.
One thing I've done is to replace all appearances of `...` in code examples with a comment, since now `...` is concrete correct syntax...

mbudiu-vmw:
@jafingerhut : if you approve this please merge it as well; we can then claim one more entry from the wiki...

jafingerhut:
Merging this, but creating another (small) PR after it is merged to address some of the comments.

apinski-cavium:
I noticed the '...' syntax was not added to the grammer in the appendix and it was never implemented in the reference compiler.

jafingerhut:
There is a p4lang/p4c PR that at least partially implements this feature, but even in the part that it currently implements, it contradicts the spec in one small way (you can see recent comments on that PR for how): https://github.com/p4lang/p4c/pull/2368

https://github.com/p4lang/p4-spec/issues/1021 attempts to answer one detail issue related to this feature.  I do not know of any agreed-upon resolution for it, although there has been some discussion on it.

One possible way forward is to allow `...` anywhere inside of `{ ... }`, and if it does not appear last, it represents _exactly one_ value in that position of the list expression, with its default value (the value being determined by the type of that element of the list, or whatever struct or header it is being "cast" to).  That is at least partially proposed in this issue https://github.com/p4lang/p4-spec/issues/1032 but there has been no verbal discussion in a LDWG meeting about it, only in comments on the issue itself, which would probably best be able to move forward with some verbal discussion to flush out any misunderstandings.

PR #803, opened 2020-01-08
Eliminate annotations on action bindings

jnfoster:
This PR is motivated by a discrepancy between the language specification and `p4c` (see  p4lang/p4c#2129).

It makes two changes:

1. Modify the grammar to eliminate optional annotations on `actionRef` components `const` table entries.

2. Extend the example of `const` table entries to illustrate a use of an `actionRef` without parentheses.

### DIFF
--- a/p4-16/spec/grammar.mdk
+++ b/p4-16/spec/grammar.mdk
@@ -515,7 +515,12 @@ keyElement
 
 actionList
     : /* empty */
-    | actionList actionRef ';'
+    | actionList optAnnotations actionRef ';'
+    ;
+
+actionRef
+    : prefixedNonTypeName
+    | prefixedNonTypeName '(' argumentList ')'
     ;
 
 entriesList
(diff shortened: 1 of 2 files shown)


p4-16/spec/grammar.mdk
```diff
@@ -515,7 +515,12 @@ keyElement
 
 actionList
     : /* empty */
-    | actionList actionRef ';'
+    | actionList optAnnotations actionRef ';'
+    ;
+
+actionRef
+    : prefixedNonTypeName
```
mbudiu-vmw: would be nice to have the implementation use the same grammar, or at least test them in the implementation to make sure there's no subtle bug.

I think we can take this change as a bug-fix, it does not need to be a new feature in 1.3.0.

jnfoster: Yep, I'll also update p4c to follow this, which seems to be the simplest way to do things (vs. having two `actionRef`-like rules, one with annotations and one without).

jnfoster:
The implementation in p4lang/p4c#2129 uses the same grammar as in this PR.

jnfoster:
Okay. Let's remember to inform the LDWG that we did this in February. I added an entry to the Wiki already that we can discharge.

PR #804, opened 2020-01-09
Empty statements in parser; added missing statement

mbudiu-vmw:


p4-16/spec/P4-16-spec.mdk
```diff
@@ -4917,9 +4917,11 @@ parserStatements
 
 parserStatement
     : assignmentOrMethodCallStatement
+    | directApplication
```
jnfoster: I assume this was an oversight?

PR #808, opened 2020-02-03
Conditional statements in parsers

mbudiu-vmw:


jafingerhut:
There is this phrase near the beginning of section 11 "Statements" that should be removed, too: "For example, conditionals are not supported in parsers"

mbudiu-vmw:
This is not implemented in the compiler, so we could even move it to 1.2.1 if we want.
https://github.com/p4lang/p4c/pull/2386

mbudiu-vmw:
The LDWG has decided to accept this change at the December 2020 meeting, so I am merging this into the spec.

PR #809, opened 2020-02-03
Clean up tokens in expression grammar

ChrisDodd:
- all operator tokens as literals rather than some as names
- precedence is specified by yacc-%left/%right rules, so use consistent
  %prec modifiers in the rules.


mihaibudiu (approved):

mbudiu-vmw:
I think we can merge this immediately, since it really a bug fix, but let's get another review.
I suspect we don't need to do anything else for right shifts.

mbudiu-vmw:
@ChrisDodd : there's a phrase in the spec that says that a semantic check is needed for right shifts, I don't think that's necessary anymore either; maybe you can delete it.

ChrisDodd:
I've removed that note as well.

Issue #805, opened 2020-01-16
Assignment to out-of-range element of header stack array


jafingerhut:
This is a low priority issue, as far as I know, as it seems that most P4 implementations restrict programs that access header stacks to do so at index values that are compile time constant values.  All such references can be checked at compile time for whether they are in or out of range.

Given the recently proposed implementation of run-time variable index expressions, both as R-values and L-values, here:

+ https://github.com/p4lang/p4c/pull/2139
+ https://github.com/p4lang/p4c/issues/2007

it made me wonder what the P4_16 specification said about cases where the run-time index was out of range, if it said anything at all.

If there is an expression `my_hdr_stack[j]` where `j` is out of range, the spec already says this in Section 8.16 "Operations on header stacks":

+ Accessing a header stack hs with an index less than 0 or greater than hs.size results in an undefined value.

When the expression `my_hdr_stack[j]` is used as an R-value, I think "undefined value" is reasonably interpreted similarly to attempting to read an uninitialized field in a header or struct, as described in Section 8.21 "Reading uninitialized values and writing fields of invalid headers".

That leaves open what it would mean to use `my_hdr_stack[j]` as an L-value when `j` is out of range, e.g. an assignment with `my_hdr_stack[j]` or `my_hdr_stack[j].field_name` on the left hand side, or used as an `out` or `inout` parameter, or one of the statements `my_hdr_stack[j].setValid()` or `my_hdr_stack[j].setInvalid()` (that might not be an exhaustive list).

One possibility would be to explicitly document that any modifications to such an out-of-range header stack reference as an L-value results in no modification to any defined value in the program, similar to how the current spec says that assignments to fields within an invalid header should be handled (refer to 3rd-to-last and 2nd-to-last paragraphs of Section 8.21).

I am sure there are other possibilities, but none immediately jump to mind that try to limit the "blast radius" of such an invalid header stack element reference.

Background detail: The behavior in the open source behavioral-model implementation right now is that using such an out of range array index either as an L-value or an R-value results in the bmv2 process throwing an exception and crashing [1].  That seems like a reasonable default behavior for behavioral-model, that can aid in finding bugs in a program using this feature.

[1] See comments in this PR: https://github.com/p4lang/behavioral-model/pull/840/files

jnfoster:
Your proposal and Bmv2's behavior both seem sensible to me.

jafingerhut:
Hmm, there may be some subtleties here I did not realize when creating this issue.

[...]

PR #812, opened 2020-02-16
Specify out of range header stack behavior

jafingerhut:


jafingerhut:
First draft of P4_16 spec changes intended to address this issue: https://github.com/p4lang/p4-spec/issues/805

The following question marked "TBD" is currently included in this PR, which should be answered, and the text updated to reflect the answer, before this is merged in:

TBD: Should an expression like
`my_hdr_stack[run_time_variable_index].isValid()` be defined as
guaranteed always false if the index is out of range?  Or should that
value be allowed to be undefined, too?

jafingerhut:
This PR currently contains the proposal that all P4 compilers should give an error if a program accesses a header stack using an expression like `hs[index]` where `index` is a compile-time constant expression, and that expression evaluates to an out of range index value.  I don't think that was in the spec before, but I submit it for the consideration of the language design work group members.

The latest version of p4c as of 2020-Feb-04 does this already, and I believe has done this for years now.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4390,6 +4410,13 @@ some unspecified value will be used for that field.
   called, which was not assigned a value during the execution of that
   `control` or `action` (this list of examples is not intended to be
   exhaustive).
+- reading a field of a header that is an element of a header stack,
+  where the index is out of range for the header stack.
+
+TBD: Should an expression like
+`my_hdr_stack[run_time_variable_index].isValid()` be defined as
```
mbudiu-vmw: Undefined sounds fine. That allows implementations that return 'false'.

jafingerhut: I have updated the PR to specify that the valid bit of an out-of-range element of a header stack is undefined, with a couple of sentences indicating what that means.

mbudiu-vmw:
I didn't see any objections to merging this change, so I will go ahead. If you think this is wrong we can back it off.

Issue #817, opened 2020-02-26
Good idea for spec to specify the behavior of repeated keys in structure initializers?


jafingerhut:
Section 8.14 "Structure initializers" currently does not mention what the behavior should be if one attempts to use a structure initializer with the same field name repeated, e.g. `{a=1, b=2, a=3}`.

The version of p4c below gives the error message shown there for one test program where I attempted to use such a structure initializer expression, which seems like a good thing for a compiler to do.

Does it seem like a good idea to define in the language spec that such a program should give a compilation error?

```
$ p4c --version
p4c 1.2.0 (SHA: 62fd40a0)

$ p4c --target bmv2 --arch v1model struct-initializer-repeated-field-name.p4
struct-initializer-repeated-field-name.p4(63): [--Werror=duplicate] error: f1: Duplicates declaration f1
        meta.s1 = {f1=3, f2=5, f1=3}
                               ^^
struct-initializer-repeated-field-name.p4(63)
        meta.s1 = {f1=3, f2=5, f1=3}
                   ^^
error: 1 errors encountered, aborting compilation
```
[struct-initializer-repeated-field-name.p4.txt](https://github.com/p4lang/p4-spec/files/4253014/struct-initializer-repeated-field-name.p4.txt)

vgurevich:
I think that this is quite reasonable.

jafingerhut:
Closing this issue, as the corresponding PR has been merged into the spec.

PR #819, opened 2020-02-26
Make it a compiler error for a structure initializer to repeat fields.

jafingerhut:


jafingerhut:
Proposed addition to the spec, in case it is decided that changing the spec for issue https://github.com/p4lang/p4-spec/issues/817 is a good idea.

Issue #822, opened 2020-03-04
Propose adding text on behavior of return/exit and copy-out


jafingerhut:
When a control/function/action executes a return or exit statement, should the copy-out behavior of any 'out' or 'inout' direction parameters be skipped, or performed?  The sections on the 'return' and 'exit' statements do not say.  There might be somewhere else in the spec that does, but I do not recall such a mention anywhere.

This p4c issue has a sample program, and what the p4c compiler currently does when transforming it over various passes, that raised this question: https://github.com/p4lang/p4c/issues/2225

jafingerhut:
Hmm, my guess is that people would probably be quite surprised if executing a 'return' inside of a function call would skip the copy-out behavior of any 'out' or 'inout' direction parameters, given that 'return <value>;' is a normal expected way to return a value from inside a function body.

The answer is less clear what people expect for 'return;' executed within an action or control.

jafingerhut:
Another related question that might be worth clarifying at the same time: Is it supported to use an `exit` statement inside of a function body?  The latest p4c as of today gives an error that it is not supported if you attempt to define a function that contains an `exit` statement.

jafingerhut:
Thinking a bit more on examples today, it seems that for a top level control like ingress or egress, where architectures like both v1model and PSA, and I believe other proprietary ones, have direction `out` or `inout` parameters for various kinds of 'standard metadata' for that architecture, P4 developers expect that those copy-out actions _do_ occur if one executes a `return` or `exit` statement inside of that ingress/egress control.  If those copy-out actions did not occur, then there would be no way to have changes to metadata to control things like the desired egress port, whether to clone/not-cone, whether to drop, etc. be visible to the architecture.

jafingerhut:
Discussed at 2020-Apr-06 P4 language design work group meeting, and this PR was merged in to resolve this issue: https://github.com/p4lang/p4-spec/pull/823#event-3207187161

PR #823, opened 2020-03-05
Specify that copy-out behavior still occurs after return/exit statements

jafingerhut:


jafingerhut:
Note: I do NOT consider this to be a minor clarifying point in the spec.  I hope that this is NOT merged unless multiple p4c compiler developers think carefully on the issue, the questions raised in issue https://github.com/p4lang/p4-spec/issues/822, and how these proposed changes in the spec might affect various implementations.  In particular, whether an `exit` statement invoked from a control/action 3 or 4 levels deep in the control 'call stack' might lead to 'exit' being more expensive than currently believed.

jnfoster:
Here's a question this issue made me consider for the first time: are an `exit` and a `return` in a top-level control semantically equivalent? 

An obvious and sensible interpretation is that they are.

Another possible interpretation would be that `exit` actually terminates the "thread" for the current packet. I don't think this is what was intended however.

If we decide that the cost of implementing the "obvious" semantics is too high, we could also imagine enriching the semantics to consider two different types of returns:
- Normal return (with copy-in / copy-out semantics)
- An abnormal return (with different semantics) due to an `exit` or `return`.

jafingerhut:
When you say 'terminates the "thread" for the current packet' as a possible interpretation of `exit`, do you mean that the packet would be dropped?  If so, I am pretty sure that wasn't the intent for those who introduced `exit`.  I think the main motivation for adding both `return` and `exit` was having a way early in a long control to stop making any more changes to the state associated with the packet, without having to require the P4 developer to do something like more and more deeply nested if/then/else statements, and/or adding a "no_more_side_effects" Boolean variable that qualifies all future table apply statements.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4681,7 +4681,10 @@ The `return` statement immediately terminates the execution of the
 are not allowed within parsers.  `return` statements followed by an
 expression are only allowed within functions that return values; in
 this case the type of the expression must match the return type of the
-function.
+function.  Any copy-out behavior due to direction `out` or `inout`
```
mbudiu-vmw: "Enclosing" is rather a lexical concept, but you want to refer to the dynamically executing action or function here. I would rather say that before transferring control to the caller the out and inout parameter values are copied to the corresponding arguments.

ChrisDodd:
I think its is desirable to preserve the property that call-by-reference and call-by-copy-in-out are equivalent if the argument is known not to alias with anything else accessed (directly or indirectly) in the callee.  So that if the compiler knows this (the argument is a value in a scope inaccessable to the the callee, or we analyze the callee to know this), it is free to elide copies and use references if that would be more efficient.

This would imply/require that copy-out does happen as part of exit/return, copying out the value of the local as of the exit/return statement.

p4-16/spec/P4-16-spec.mdk
jafingerhut: The first sentence of the paragraph also uses words that look like lexically enclosing thing, too: "The `return` statement immediately terminates the execution of the `action`, `function` or `control` containing it."

Do you want to change that sentence, too?  It seems to me that the lexically enclosing action, function, or control is exactly what `return` is meant to affect, isn't it?

mbudiu-vmw: Ok, you can leave the text as it is.

PR #828, opened 2020-03-10
Fix issue 826

jafingerhut:


mbudiu-vmw:
I think we can merge this safely.

Issue #830, opened 2020-03-23
Side-effects in table keys and select labels


mbudiu-vmw:
The spec does not say when the side-effects that may occur in table keys and select labels occur.
One simple solution would be to forbid function and method calls in these places.

jafingerhut:
I don't think you want to forbid `hdr.isValid()` method calls in table key expression, do you?

Perhaps the most precise way to express the restriction is something like "key expressions may only contain pure function and method calls, and even among those, only ones that have no direction `out` or `inout` parameters"?  This is assuming the definition of `pure` proposed in this article: https://github.com/jafingerhut/p4-guide/tree/master/extern-annotations

Given the contents of that article, it would be correct for the built-in method `hdr.isValid()` to be annotated with the `@pure` annotation, and it has no direction `out` nor `inout` parameters.

jafingerhut:
I believe that all P4_16 functions, by which I mean the ones written in P4_16, not an extern function, must be 'pure' in the same sense as in that linked article in my previous comment, and might be reasonable to call inside a table key expression, as long as they have no  direction `out` nor `inout` parameters.

mbudiu-vmw:
Indeed, the implementation allows "built-in" functions, like isValid.
Still, the question is "when are these functions evaluated", and in what order.
For tables it's perhaps easier to answer this question, but for select labels the situation is trickier. There is no obvious place to lift the function evaluation, esp. if you want to evaluate only as many labels as necessary.

jafingerhut:
If they are pure functions with neither `out` nor `inout` direction parameters, isn't the order of evaluation unimportant, and their time of evaluation the only thing that matters is "before the value is needed"?

mbudiu-vmw:
There is the semantics, but there is also the cost of evaluating the function, that some of the people may care about. Perhaps we should not worry too much about this in the spec; I maintain, however, that these restrictions won't surprise too many people.

jafingerhut:
So are you proposing for the spec to say something like "table key expressions may not contain any function or method calls, except the following list of built-in ones: ...", where that list of allowed ones includes `hdr.isValid()`?

mbudiu-vmw:
Something to that effect; I put this on the agenda for the next design meeting.

mbudiu-vmw:
But there's no PR yet...

jafingerhut:
I am asking because I started to write a PR, and had questions about what it should say.  May create a proposed one soon.

jafingerhut:
[...]

PR #831, opened 2020-03-24
Disallow most function and method calls in select expressions or table keys

jafingerhut:


jafingerhut:
This PR is a first draft intended to address issue https://github.com/p4lang/p4-spec/issues/830

jafingerhut:
Closing this PR, since the LDWG decided to merge #834 instead of this one, and they were two opposing approaches to resolving the issue.

PR #834, opened 2020-04-04
Define the order of table key evaluation explicitly

jafingerhut:
Include an example demonstrating how a P4 compiler might choose to
implement a table apply call where the table key expressions have side
effects.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6264,11 +6264,18 @@ apply {
 }
 ~ End P4Example
 
-It is expected that evaluating a key expression will usually have no
-side effects, so the order of their evaluation makes no difference in
-the resulting lookup key value.  In this common case, there is no need
-for a compiler to generate code that assigns the results of key
-expressions to fresh temporary variables.
+Note that the second code example above is given in order to specify
+the behavior of the first one.  An implementation is free to choose
+any technique that achieves this behavior[^MostP4ProgramsDontDoThis].
+
+[^MostP4ProgramsDontDoThis]: Most existing P4~16~ programs today do
+    not use function or method calls in table key expressions, and the
+    order of evaluation of these key expressions makes no difference
+    in the resulting lookup key value. In this overwhelmingly common
+    case, if an implementation chooses to insert extra assignment
+    statements to implement side-effecting key expressions, but does
+    not insert them when there are no side-effecting key expressions,
+    then in typical programs they will almost never be inserted.
 
 ## The Match-Action Pipeline Abstract Machine { #sec-mau-abstract-mach }
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6211,6 +6211,65 @@ apply_result(m) m.apply() {
 }
 ~ End P4Pseudo
 
+The behavior of the `buildKey` call in the pseudocode above is to
+evaluate each key expression in the order they appear in the table key
+definition.  The behavior must be the same as if the result of
+evaluating each key expression is assigned to a fresh temporary
+variable, before starting the evaluation of the following key
+expression.  For example, this P4 table definition and apply call:
+
+~ Begin P4Example
+bit<8> f1 (in bit<8> a, inout bit<8> b) {
+    b = a + 5;
+    return a >> 1;
+}
+bit<8> x;
+bit<8> y;
+table t1 {
+    key = {
+        y & 0x7  : exact @name("masked_y");
+        f1(x, y) : exact @name("f1");
+        y        : exact;
+    }
+    // ... rest of table properties defined here, not relevant to example
+}
+apply {
+    // assign values to x and y here, not relevant to example
+    t1.apply();
+}
+~ End P4Example
+
+is equivalent in behavior to the following table definition and apply
+call:
+
+~ Begin P4Example
+// same definition of f1, x, and y as before, so they are not repeated here
+bit<8> tmp_1;
+bit<8> tmp_2;
+bit<8> tmp_3;
+table t1 {
+    key = {
+        tmp_1 : exact @name("masked_y");
+        tmp_2 : exact @name("f1");
+        tmp_3 : exact @name("y");
+    }
+    // ... rest of table properties defined here, not relevant to example
+}
+apply {
+    // assign values to x and y here, not relevant to example
+    tmp_1 = y & 0x7;
+    tmp_2 = f1(x, y);
+    tmp_3 = y;
+    t1.apply();
+}
+~ End P4Example
+
+It is expected that evaluating a key expression will usually have no
+side effects, so the order of their evaluation makes no difference in
+the resulting lookup key value.  In this common case, there is no need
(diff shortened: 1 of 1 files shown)


jafingerhut:
This is one of two proposed PRs to address issue https://github.com/p4lang/p4-spec/issues/830

The other proposed PR is https://github.com/p4lang/p4-spec/pull/831. That one disallows side effects in table key expressions and parser select expressions.

This one proposes a way to explicitly define the behavior when table key expressions have side effects.  If the approach of this PR is preferred over the other one, I would recommend that this one be enhanced with similar language specifying the order of evaluation of transition select expressions.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6211,6 +6211,65 @@ apply_result(m) m.apply() {
...
+apply {
+    // assign values to x and y here, not relevant to example
+    tmp_1 = y & 0x7;
+    tmp_2 = f1(x, y);
+    tmp_3 = y;
+    t1.apply();
+}
+~ End P4Example
+
+It is expected that evaluating a key expression will usually have no
```
cc10512, line 6267: "expected" leaves too much choice to the compiler implementer. I suggest that you mandate that any function that is called as a table key is side-effect free, otherwise it is an error.

jafingerhut: For this proposed PR (which might be rejected in favor of the restriction you describe, for which there is a different proposed PR already for the same issue), the intent is that the example program transformation is always correct.  The last paragraph is simply a note that the creation of temporary storage is usually unnecessary, given the way P4 developers typically write programs.

cc10512:
I would argue for a middle ground. Rather than banning all expressions and function calls, allow them, but mandate they are side-effect free. If the compiler can't prove that they are side free, we can
ask the programmer to annotate the function/expression. We have annotations to tag such functions, and if functions are annotated, the compiler should be able to prove expressions. The compiler can keep giving errors until the side-effect free execution can be proven.

This would be less restrictive than #831, and yet achievable.

jafingerhut:
@cc10512 #831 already mentions as optional that a P4 compiler could support side-effect free function/method calls.

jnfoster:
@cc10512 in p4lang/p4c#2259 I'd pointed out that in my reading the existing specification at least hints at left-to-right evaluation of table keys.

Evidence:
1. [Section 13.2.3](https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-mau-semantics) defines the semantics of MAU execution. In particular, the pseudocode includes the line:
```
    var lookupKey = m.buildKey(m.key); // using key block
```
which suggests that some work may be done to construct the key in general.

2. [Section 6.7](https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-calling-convention) defines the calling convention, and says "Arguments are evaluated from left to right prior to the invocation of the function itself."

Now, it's not totally clear that the calling convention for the P4 language applies to the pseudocode we use in the language specification itself. So I think it's a good idea to tidy up the spec.

But if we restrict the kinds of expressions that can appear in tables, then we are making a non-backward-compatible change and may break some programs. I think that's something we should try to avoid when possible. In this case, it's trivial for the compiler to add code before the table application statement to setup the keys, so I don't think there's a compelling reason to make this change. 

One push back I've heard about this position is that this transformation has a cost -- extra logic and extra state to store the keys. That's fair, but I would argue that these costs are fairly modest -- linear in the number of key elements that are not already a value. Programmers who are on a resource-restricted target can be made aware and program accordingly. Everyone else can benefit from being able to write

```
table t {
  key = { f() : exact }
  ...
} 
...
t.apply();
```

instead of
T x;
table t { 
  key = { x : exact; }
  ...
}
...
x = f();
t.apply();
```

**Comment by jafingerhut** [comment]
@cc10512 Not trying to dog-pile on you, really, but just another thought.

Think of the code required in a compiler to check whether key expressions have side effects, or not.  That code must be written as part of p4c in order to enforce the restriction that you would like to make in the spec.

That same code can be used to distinguish in the compiler between the case where extra temporaries are needed, or not.  If the check in the restricted version of the P4 language passes for a particular table, no temporaries need to be introduced.  If the check would say it is an error, then the compiler introduces the temporaries, and only then.

**Comment by jnfoster** [comment]
I will pile on! @cc10512 if you feel strongly, you should join us at the LDWG on Monday :-)

**Comment by cc10512** [comment]
@jafingerhut, @jnfoster feel free to keep piling on. That's the only way we can reach a conclusion that has the right compromise. 

@jnfoster unfortunately I am already booked for 11am on Monday, and so I won't be able to join the LDWG meeting. So let me explain why I find #831 unsatisfactory. The 'spec' says:
```
Select expressions may not contain any function or method calls,
except for the following:

The `isValid()` method on headers (Section [#sec-ops-on-hdrs])
[...]
```
This, at least to me, clearly implies that any other implementations that allow calls to other functions are not conformant to the spec. And proposes that the programmer add these function calls ahead of the apply, to compute the expressions. In my view, that just introduces clutter and non-local reasoning. Given these two aspects, I argue that it is better to explicitly allow expressions that include only "pure" functions directly in the table keys. And as @jnfoster outlined, the transformation to a code that stores the temporary values in variables is quite trivial, so there should be no difference from a compilation point of view. The amount of extra state can be managed by targets with clear messages, just like any other temporaries and/or state that needs to be managed in the program. I really do not see how this will add any more burden than what programmers can do today: if they want to pre-compute these values, they can do so with the current spec and in the new proposed spec.

Re backward compatibility: since the current spec is quite imprecise on what can be considered a key, is the concern that the current proposal can be interpreted as: "you can stash the `buildKey` function from Section 13.2.3 as a table key expression even if it has side-effects"? If so, indeed, I can see how the new implementation of the compiler might reject such programs, even though, the programmers already had the ability to pre-compute these functions (and likely they did it outside the table key expression!). And while proving that an expression is side-effect free is trivial, it does seem potentially limiting, when the compiler can simply transform the code to extract these computations ahead of time. However, if that is not the case, I do believe that constraining table key expressions to side-effect free expressions is worthwhile in the long run and it enables high performance targets.

To conclude, FWIW, I am in favor of allowing function calls and implementing the compiler transformation that lifts the computation of table key expressions before the table apply in order to avoid breaking backward compatibility. I would certainly avoid limiting the functions to `isValid()`. If the agreement is not to constrain the expressions to side-effect free ones, and if the compiler can prove that an expression is side-effect free, we may consider optimizations later -- especially because we already have such annotations for functions!

Sorry for the long note. I hope it helps you understand why I thought that neither proposal quite gets to the sweet spot.

jafingerhut:
@cc10512 I've read your previous comment, and it seems to be completely consistent with PR #834.  That PR says that anything can be in a key expression, including method/function calls with side effects.  Am I misinterpreting what you are saying?

jafingerhut:
There is an issue specifically about action arguments I created here: https://github.com/p4lang/p4-spec/issues/840

While these issues are related, I would not be surprised if the answers for these two issues that the LDWG comes up with might differ.

jafingerhut:
I merged in a slightly modified version of the PR that was reviewed.  The only change was to attempt to 'weaken' the last several sentences of the proposed additions, moving them into a footnote.

PR #835, opened 2020-04-09
Add @pure and @noSideEffects annotations on extern functions/methods

jafingerhut:


jafingerhut:
This is the subset of https://github.com/p4lang/p4-spec/pull/802.  It includes only the additions of the `@pure` and `@noSideEffects` annotations.

I did add one phrase to qualify the example of a `hash` function that would be legal to have a `@pure` annotation -- a `hash` function that had control-plane writable seed/initialization-vector value should not be annotated `@pure`.

jafingerhut:
Since creating this PR, I have made a few more changes to the proposed text.  I may have gone overboard with too many words, so please read it and see if my edited version looks reasonable.

jafingerhut:
Is the next version number 1.2.1?  1.3.0?  Something else?  The latest version number mentioned in the revision history right now is 1.2.0, released late 2019.

mbudiu-vmw:
We may need to add more stuff there.
Use 1.2.1 and we can fix this later.

jafingerhut:
Beginnings of a revision history for the next release added.  I did not attempt to add all of them, only the one in this PR.

mbudiu-vmw:
I think we agreed to merge these changes in the spec.

PR #838, opened 2020-04-09
Correct a few spelling mistakes

jafingerhut:


jafingerhut:
I saw a spelling mistake in another PR I was editing, and ran a spell checker over the whole spec.  Only a few problems found that way, all corrected in this PR.  No semantic changes to the spec are intended.

PR #841, opened 2020-04-20
Change typing for masks

mbudiu-vmw:


mbudiu-vmw:
@jafingerhut any observations?

mbudiu-vmw:
Ok, then I will merge this; during the LDWG this solution was suggested, so I assume we agree to push this into 1.2.1.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3798,11 +3798,12 @@ select (hdr.ipv4.version) {
 
 ### Masks { #sec-cubes }
 
-The infix operator `&&&` takes two arguments of type `bit<W>`,
-and creates a value of type `set<bit<W>>`. The
-right value is used as a "mask", where each bit set to `0` in the mask
-indicates a "don't care" bit. More formally, the set denoted by `a
-&&& b` is defined as follows:
+The infix operator `&&&` takes two arguments of type `bit<W>` or
+serializable `enum`, and creates a value of type `set<ltype>`, where
```
vgurevich, line 3802: There is no harm adding this, but given automatic conversion from serializable enum to bit<W> I wonder if this is adding anything of substance

mbudiu-vmw: Yes, because without this the result of `&&&` would be bit - since the serializable enum is automatically converted to bit, and we need it to be actually serializable enum.

Issue #842, opened 2020-04-23
Narrow scope of implicit casts


hackedy:
The implicit casting rules say that if one argument is an `int` and the other is `int<w>`, then the `int` will be casted to `int<w>`. I interpreted the text to mean that in any other situation there wouldn't be implicit casts, but it seems like casts happen in other places.

Here are some examples, just to make this more concrete.
<ol>
<li>p4c thinks that the following program typechecks.
<pre>
  const int<8> y = 10;
  const int<8> x = 2 + 2;
</pre>
</li>
<li>The spec includes examples <a href=https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-structure-initializers>like this</a>:
<pre>
struct S {
    bit<32> a;
    bit<32> b;
}
const S x = { a = 10, b = 20 };
</pre>
Here the bit<32> fields of a struct are being initialized with int values, without casts.</li>
<li>In <a href=https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-entries>another example</a>, both operands to &&& are <code>int</code> but implicit casts are done to make it match the key type <code>tuple&lt;bit&lt;8&gt;,bit&lt;16&gt;&gt;</code>.
<pre>
const entries = {
        (0x01, 0x1111 &&& 0xF   ) : a_with_control_params(1);
        ...
    }
}
</pre>
</li>
<li>p4c thinks that the following program typechecks.
<pre>
int&lt;8&gt; f(int&lt;8&gt; input) {
    return input;
}
control C() {
    apply { int&lt;8&gt; z = f(1); }
}
</pre>
Unlike the other examples, though, this one is <a href=https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-invocations>pretty clearly forbidden by the spec</a>:
<blockquote>The compiler does not insert implicit casts for the arguments to methods or functions—the argument types must match the parameter types exactly.</blockquote>I'll file a bug for it over at p4c but wanted to include it here for completeness.
</li>

ChrisDodd:
This seems like a pretty severe (and undesirable) restriction -- the point of these casts is to allow constants to be used "naturally" rather than requiring excessive work from users.

jafingerhut:
I do not know if this is the intent of the P4 LDWG or not, but would the spec become more internally consistent, and closer to the p4c implementation (perhaps matching its behavior -- I am not sure about that, though), if the sentence you quote:

[...]

PR #843, opened 2020-04-23
Explicitly allow implicit casts for direction 'in' arguments

jafingerhut:
Intended to address issue https://github.com/p4lang/p4-spec/issues/842


mbudiu-vmw:
This is a tiny bug fix.

PR #847, opened 2020-04-29
Add cross-reference for default values section, and fix typo

jafingerhut:


jafingerhut:
@mbudiu-vmw Somehow I missed noticing that you had already added the suggested examples in your last commit before I merged the changes in https://github.com/p4lang/p4-spec/pull/797. Thanks for adding those.

I did find one typo to fix, and this PR also adds a cross-reference within the document that seems worthwhile, since 'default' is used for several different kinds of things in P4_16.

Issue #846, opened 2020-04-29
Possible clarification for initialization of direction out parameters?


jafingerhut:
A quote from version 1.2.0 of the language spec, Section 6.7 "Calling convention: call by copy in/copy out" https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-calling-convention:

'out parameters are uninitialized (parameters of type header or header_union are set to “invalid”) and are treated as l-values (See Section 6.6) within the body of the method or function.'

The v1model architecture parsers have a direction `out` parameter with type `H` that is a struct containing fields that may be any header, header union, or header stack type, and no parser definition I have ever seen explicitly initializes these headers, header unions, and header stack elements to invalid.  It is assumed by the code that they are initially invalid, and the quote above is probably the reason why.

Assuming that is all true, it seems the quote above could be made more explicit, e.g. something like this:

out parameters are uninitialized and are treated as l-values (See Section 6.6) within the body of the method or function.  There are several exceptions where out parameters are explicitly initialized, at least partially:

+ If an out parameter is of type header or header_union, it is set to "invalid"
+ If an out parameter is a header stack containing either header or header_union elements, all elements of the header stack are set to "invalid", and its `nextIndex` counter is initialized to 0, as discussed in 8.16
+ If an out parameter is a compound type such as a struct or tuple, and any of its members is any of the types above, they are initialized as described there.  If any of its members are a compound type, then recurse into those elements.

vgurevich:
Yes, most of the code relies on this rule pretty extensively, so the proposed clarification would be nice. 

For the compound types I would only add that if they have non-header elements, those will remain uninitialized and the compiler is free to produce warnings about those (if they are not assigned by the end of scope).

jafingerhut:
See PR #848 for proposed clarifications to the spec

jafingerhut:
Closing as resolved, since PR #848 was merged.

Issue #845, opened 2020-04-28
Should headers always be initialized as invalid?


jafingerhut:
Consider an action like `a1` below (the same questions apply for a header, header_union, or header stack local variable within a control, parser state, function, or block enclosed in curly braces inside the body of any of these things):
```
header h1_t {
    bit<8> f1;
    bit<8> f2;
}

control c1 () {
    action a1 () {
        h1_t h1;
        // Point A: what can we say must be true about the value of h1 here?
        if (h1.isValid()) {
            h1.f1 = 1;
            h1.f2 = 2;
        } else {
            h1.setValid();
            h1.f1 = 101;
            h1.f2 = 102;
        }
    }

    apply {
        a1();
    }
}
```
At point A in the program's execution, is there anything we can say about the value of h1 according to the language specification?

For example, can we say:

(a) At point A, it is guaranteed that `h1.isValid()` will return false.

Or is the following true?

(b) The return value of `h1.isValid()` undefined, and thus it is implementation-dependent whether it will return true or false, and might even return true when processing some packets, but false when processing other packets.

Or is there some third alternative that makes sense?

Here are parts of the P4_16 language specification version 1.2.0 that I think might be relevant, but my goal in creating this issue is to find out what people think ought to be true, and clarify the spec to match.

Section 7.2.2 "Header types" https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-header-types says: 'When a header is created its “validity” bit is automatically set to false.'  This seems to imply statement (a) above is true.

Section 10.2 "Variables" https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-variables says: 'Variable declarations without an initializer are uninitialized (except for header stacks, which have their nextIndex counter initialized to 0, as discussed in 8.16).'

and a little bit later in that same section 'Reading the value of a variable that has not been initialized yields an undefined result. The compiler should attempt to detect and emit a warning in such situations.'

That seems to imply statement (b) above is true.

There is a somewhat related issue involving direction `out` parameters that I will create separate from this one.  The questions raised in this issue seem to have nothing to do with parameters, only local variables with some header or header-like type.

jafingerhut:
[...]

PR #848, opened 2020-05-02
Issue846 clarify out parameter initialization

jafingerhut:


jafingerhut:
Clarifications intended to address issue #846

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1503,6 +1503,47 @@ Each parameter may be labeled with a direction:
 	- an action parameter that can be set directly by another calling
       action; in this case it behaves like an `in` parameter
 
+Direction `out` parameters are always initialized at the beginning of
+execution of the portion of the program that has the `out` parameters,
+e.g. `control`, `parser`, `action`, function, etc.  This
```
jafingerhut, line 1508: It is probably sufficient to avoid changing the meaning of existing P4_16 programs if this initialization were limited to only "top level controls/parsers" in an architecture, since the main example I know of where this initialization is expected is for the one direction `out` parameter that is typically a struct containing headers that is `out` from the parser, usually declared with a type parameter named `H`.

That restriction might be considered a bit odd in a language spec, perhaps, but I wanted to mention it for consideration.  "top level" in this situation could be defined as "an object given as a parameter to the package named `main`, or via any chain of package invocations starting from `main`".

jafingerhut:
@mbudiu-vmw I see the issue of `out` direction parameters and local variables as separate, primarily because `out` direction parameters are so common among multiple open source and at least one proprietary architecture for the type `H` used to contain headers for the parsers.  If they were declared `inout` instead, I would guess that Vladimir would be doing his best to eliminate the initialization guarantees that we have in the spec now.

For local variables, do you (or anyone) know of any architectures whose behavior would change for typical P4 programs if local header variables were officially "uninitialized" in their validity?  If so, probably best to add a comment explaining the situation to this issue: https://github.com/p4lang/p4-spec/issues/845

p4-16/spec/P4-16-spec.mdk
vgurevich: Not that I am opposed to it, but this proposed restriction pretty much admits that this little exception is a property of an architecture and not of the language, does it not? (And as a consequence this becomes a restriction on all the architectures, mandated by the language spec).

jafingerhut: If you have another idea of how to change the language spec in a way that _doesn't_ change the behavior of all those existing P4_16 programs that rely on `out` parameters being initialized to invalid, now is a good time to mention it :-)

Yes, this proposed way makes it a restriction on all `out` parameters everywhere.  We could restrict to only parsers, or only the top level parsers, etc. trying to limit the extra cost to an implementer, at the possible cost of people reading the spec and wondering what the heck it is restricted to that special case for.  The more restrictive we make it, the more I would propose having an appendix with a rationale for those future readers (and ourselves, after we forget).

jafingerhut: If you truly want to eliminate any and all requirements to initialize `out` parameters from the language spec completely (a slightly backwards-incompatible change at this point, mind you, but I don't mind going to P4_16 v2.x over it if no one else does), then ...

One way to do that is to change the architectures that have these `out` type `H` parameters for their parsers to `inout`.  Yes, I understand that breaks the compilation of all programs written for those architectures, until they are updated to change that one `out` in their parser definition to `inout`, to match the new architecture definition.

If you leave the architectures with direction `out`, there is no language-supported way for an architecture to initialize the value of that parameter from "outside of the parser" by providing an initial value, without basically saying: "I know this architecture says this parameter is `out`, but we fake it in our compiler and treat it like it is `inout` instead."

ChrisDodd:
The spec already says:

   - `out` parameters are uninitialized (parameters of type `header`
     or `header_union` are set to "invalid") and are treated as l-values
     (See Section [#sec-lvalues]) within the body of the method or function.

is that not adequate?  What needs to be changed?

jafingerhut:
@ChrisDodd  What about an `out` parameter that is a struct type that contains 5 members all with type header?  It seems to me one could read the spec and say "well, my out parameter is a struct, not one of those things the language spec says I have to initialize, so I won't".  That seems clearly against what existing P4_16 parser code expects to happen, as far as I can tell.

p4-16/spec/P4-16-spec.mdk
vgurevich: @jafingerhut -- you are pretty much reading my thoughts. I agree that we should write this in an appendix or something. 

It sounds to me that the right thing to do if we would **not** worry about backwards compatibility is to declare that `out` is not initialized, periodm and use `inout` for most of the things that are passed into the top-level controls/parsers by (some typical) architectures. Generally, it matches the traditiona "header/metadata bus" concept well. 

However, we might need to bite the lip and save this for some bigger juncture, unless there is a big desire to do it right now (which I will not be against). The realist in me, though, says that stating that header-like `out` parameters will be pre-initialized is fine for now and especially fine if we limit those to the top-level parsers/controls only as you suggested.

ChrisDodd:
IMO that's covered by what is in there, implicitly -- a header is a header, regardless of whether it is part of a struct or header_stack or not.  In fact I would thin it would make sense to say *all* uninitialized headers are invalid, whether they are out params or locals, or anything else.

jafingerhut:
@ChrisDodd I agree with you that it was very likely the intent of that part of the spec.  I am not hoping with this PR to change the intent of the spec, only to make it more explicit, rather than implicit.

jafingerhut:
This PR was discussed at the 2020-May-04 P4 LDWG meeting, and it was decided to merge this for the forthcoming version 1.2.1 of the specification.

jafingerhut:
I will create a separate issue regarding the ideas in Vladimir's and my conversation on this issue, for future discussion.  During the 2020-May-04 LDWG meeting, it was mentioned that such alternate ideas seem like changes that would not be backwards compatible with the current specification.

Issue #849, opened 2020-05-05
Reducing requirements for initializing headers


jafingerhut:
In the 2020-May-04 LDWG meeting, issue https://github.com/p4lang/p4-spec/issues/846 and its associated PR https://github.com/p4lang/p4-spec/pull/848 were discussed, as well as the related issue https://github.com/p4lang/p4-spec/issues/845

It was decided that for version 1.2.1 of the language spec, for both `out` parameters and local variables, it would be made explicit that all headers, header unions, and header stacks that are in these places of a P4_16 program, or within a struct or tuple in one of these places, would be mandated to be initialized as invalid, as opposed to the main alternate proposal, which was allowing implementations to leave the valid bit of such headers to remain uninitialized and undefined, i.e. it would be unpredictable whether hdr.isValid() would return true or false for such headers.

It was also decided to create another issue for future discussion that in a future version of the language spec, these requirements should be eliminated, i.e., an implementation would be allowed to leave the valid bits of such headers undefined.  This is that issue.

A concern was raised that at least for `out` parameters, but perhaps also for local variables, writing the spec such that the valid bits could be undefined would be a backwards-incompatible change, requiring existing P4_16 programs that rely on such initialization to be changed in order to continue working as they did before.  If we were strict about following the rules of semantic versioning, it seems that such a change in the language spec should cause the language specification to increase the major version number, e.g. from 1.2.1. to 2.0.0.

jnfoster:
With recent improvements to the specification, I think this can effectively be closed.

PR #850, opened 2020-05-05
Require that local variables with header-like types are initialized i…

jafingerhut:
…nvalid

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2287,8 +2287,9 @@ one bit long field. This declaration introduces a new identifier in the
 current scope; the type can be referred to using this identifier. A header is
 similar to a `struct` in C, containing all the specified fields. However, in
 addition, a header also contains a hidden Boolean "validity" field. When the
-"validity" bit is `true` we say that the "header is valid". When a header is
-created its "validity" bit is automatically set to `false`. The "validity"
+"validity" bit is `true` we say that the "header is valid". When a local
+variable with a header type is
+declared, its "validity" bit is automatically set to `false`. The "validity"
 bit can be manipulated by using the header methods `isValid()`, `setValid()`,
 and `setInvalid()`, as described in Section [#sec-ops-on-hdrs].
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
In discussion on the 2020-May-04 LDWG meeting on issue #845 and #846, it was resolved to create a PR to address #845 to strengthen the requirements, such that all headers, header_union, and header stack local variables (even if they are embedded within a struct or tuple) must be initialized as invalid.

This PR attempts to make these changes.  It modifies some text in the only two places in the spec that I know of where such initialization is mentioned, making the stronger initialization requirements explicit.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2287,8 +2287,9 @@ one bit long field. This declaration introduces a new identifier in the
 current scope; the type can be referred to using this identifier. A header is
 similar to a `struct` in C, containing all the specified fields. However, in
 addition, a header also contains a hidden Boolean "validity" field. When the
-"validity" bit is `true` we say that the "header is valid". When a header is
-created its "validity" bit is automatically set to `false`. The "validity"
+"validity" bit is `true` we say that the "header is valid". When a local
+variable with a header type is
+declared, its "validity" bit is automatically set to `false`. The "validity"
```
jafingerhut, line 2292: This mention of the initialization requirement on a value with type header already seemed pretty explicit to me.  The only change here is to change the phrase "when a header is created" to "when a local variable with a header type is declared", since the proposed replacement seems a little clearer to me than "created".

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4700,9 +4701,10 @@ optInitializer
     ;
 ~ End P4Grammar
 
-Variable declarations without an initializer are uninitialized (except for header
-stacks, which have their `nextIndex` counter initialized to `0`,
-as discussed in [#sec-expr-hs]). The language places few restrictions on
+Variable declarations without an initializer are uninitialized (except for
+headers and other header-related types, which are initialized to invalid in the
+same way as described for direction `out` parameters in Section
+[#sec-calling-convention]). The language places few restrictions on
```
jafingerhut, line 4707: Rather than repeating all of the text used to describe how direction `out` parameters must be initialized, I propose referring to that new text from here.

jafingerhut:
I am going ahead and merging this, given that it has been approved on Github, and in the 2020-May-04 P4 LDWG meeting.  I have created a related issue that was raised during that meeting, where it was suggested that perhaps a future P4_16 language spec version should consider making such header initializations as no longer required: https://github.com/p4lang/p4-spec/issues/849

PR #851, opened 2020-05-05
Forbid most side effects in select labels

mbudiu-vmw:


p4-16/spec/P4-16-spec.mdk
```diff
@@ -5228,9 +5228,11 @@ verify(false, error.NoMatch);
...
 expression be compile-time known values. Keysets are evaluated in
 order, from top to bottom as implied by the pseudo-code above; the
-first keyset that includes the value in the `select` argument
-provides the result state. If no label matches, the execution triggers
-a runtime error with the standard error code `error.NoMatch`.
+first keyset that includes the value in the `select` argument provides
+the result state. If no label matches, the execution triggers a
+runtime error with the standard error code `error.NoMatch`. Function
+or method calls are forbidden in keyset expressions, except calls to
+the header methods `isValid()`, `minSizeInBits()`, and `minSizeInBytes()`.
```
jafingerhut, line 5235: The first sentence of the paragraph says "Some targets may require that all keyset expression in a select expression be compile-time known values."

Are you proposing here that even for implementations that allow keyset expressions that are not compile-time known values, the only function or method calls permitted are ones that call the method `isValid()`?

Note that `minSizeInBits()` and `minSizeInBytes()` are already specified to return compile-time known values.

mbudiu-vmw: yes. no target allows such calls.

jafingerhut: Do any targets support keyset expressions whose values are not compile-time known values?  (I believe that the name of a parser value set object instance is a compile-time known value.)

mbudiu-vmw: Is this a breaking change, which changes the set of accepted programs?

mbudiu-vmw: We should think about how this can be implemented in the compiler with side-effects.

mbudiu-vmw:
The LDWG has agreed that this should not be accepted.

PR #852, opened 2020-05-05
Forbid function calls in action arguments

mbudiu-vmw:


jafingerhut:
I personally would not have any objections to such a restriction, although it does seem possible to define the behavior of such function calls in the `actions` table property if one wished to do so.

For example, one could define that it should behave as if those function/method calls were embedded inside the action, executed first before the part of the action body defined in the program's source code.

jafingerhut:
I want to take back at least part of what I said in my last comment.  Here is an example that I think if it were allowed, I am pretty sure it would lead to wailing and gnashing of teeth among P4 implementers, or at least a significant fraction of them:
```
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {

    table sub_table {
        key = {
            h.eth_hdr.eth_type: exact;
        }
        actions = {
            // ... action list for sub_table omitted ...
        }
    }

    action a1 (in bit<32> addr, bit<16> param2) {
        // ... body of a1 omitted ...
    }

    table simple_table {
        key = {
            h.eth_hdr.srcAddr: exact;
        }
        actions = {
            a1(sub_table.apply().hit ? h.ipv4.srcAddr : h.ipv4.dstAddr);
        }
    }
    apply {
        simple_table.apply();
    }
}
```
Having a table that is invoked 'in the middle of applying another table', as is done for sub_table in the example above, after the lookup has been done for simple_table, but before simple_table's action a1 is executed, seems like it would lead us towards madness.  If allowed, one could nest those arbitrarily deeply, and even have conditional table lookups inside of ternary expressions like `a1(sub_table.apply().hit ? (t2.apply().hit ? (t3.apply().hit ? e1 : e2) : e3) : (t4.apply().hit ? e4 : e5))`.

jnfoster:
Not sure if this is madness or not...
````
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    table sub_table {
        key = {
            h.eth_hdr.eth_type: exact;
        }
        actions = {
            // ... action list for sub_table omitted ...
        }
    }
    action a1_first(bit<16> param2) {
      meta.sub_table_apply = true;
      meta.param2 = param2;
    }
    action a1_rest(in bit<32> addr) {
        // use meta.param2 instead of param
    }
    table simple_table {
        key = {
            h.eth_hdr.srcAddr: exact;
        }
        actions = {
            a1_first;
        }
    }

    apply {
      simple_table.apply();
      if(meta.sub_table_apply)
        if(do_sub_table.apply().hit) 
          a1_rest(hdr.ipv4.srcAddr);
        else 
          a1_rest(hdr.ipv4.dstAddr);
    }    
}
```

jafingerhut:
I don't have a proof that it is always possible to transform a P4_16 program with such actions into one that does not use such actions, but it seems likely that it is always possible.

I do seriously question whether it is worth the effort to implement such transformations in a P4 compiler, given that I would guess that no P4 developer would want to write code this way, for their own peace of mind and understanding of the code.

mbudiu-vmw:
That's one point having a front-end: it will do this once and for all, and no one using the front-end will have to worry about it anymore. That's really the point of the SideEffectOrdering pass, but unfortunately currently it's buggy.

jafingerhut:
This message is not about correctness of such a transformation, but about efficiency of generated code, which might not be a big concern, but it should be in the list of concerns somewhere.

Any general transformation for such code will have to somehow remember the action that was selected, and all of its action parameters that came from the control plane API, from the time that the matching is finished, until the time that the action begins.  That could be, for example N bits wide in the worst case, times A different possible actions.  So a simple front-end transformation will create N*A bits of temporary variables with unique names, and use at most a fraction 1/A of that.  Now it is up to the back end of the compiler to either always allocate all N*A bits for every packet, or to reverse engineer from the code produced as output from the front end to determine what the A groups of N bits are that never need to be stored simultaneously.

Either that, or you add unions to the language (header unions are not general enough for this purpose, because they can only have fields of a few bit-string-like types, not the full set of types that the language permits as action parameters), so that the front end can create new unions to make that potential memory sharing explicit, so the back end doesn't have to reverse engineer it.

mbudiu-vmw:
If we do forbid operations with side effects in action arguments then most of this just goes away...

jafingerhut:
Mihai wrote: "If we do forbid operations with side effects in action arguments then most of this just goes away..."

I understand.  I brought up the example because before I thought of it, I mistakenly believed that the only program transformation needed to support arbitrary expressions in action parameters of the `actions` table property was for the compiler to synthesize a new action that evaluated those expressions inside of the new action, before the body of the original action.

The example above shows that such a simple transformation is not always sufficient, because not all expressions are legal to evaluate within an action body.  That is, in the last table of Appendix F of the spec, it says that it is not supported to call a table from within an action body.

jnfoster:
Questions:
* Should we adopt this restriction at all?
* Is it backward compatible?
* How do we want to formulate the restriction? All function calls? All expressions with side effects? All expressions that invoke tables?
* Or should we not adopt it and clarify that targets can forbid invoking tables from actions, including via the expressions provided as arguments to its actions.

jnfoster:
Also, we should check whether side effects are handled correctly.

jafingerhut:
It seems like a non-trivial amount of work to implement and write tests for the fully general capability to write arbitrary expressions in actions parameters within tables, in open source p4c, such that it works.  It also seems of fairly low benefit, given that I don't think I have ever seen a program that did this, nor does it seem like a hardship to avoid writing P4 code that uses this generality.  Think of all of the questions that come up about working around limitations in P4 (e.g. shift operations where the shift amount must be a compile-time known value, `if` statements not supported fully generally within action bodies for the BMv2 back end, division and multiplication operations with restricted operands, etc.) and I doubt more than 0.1% of P4 developers will ever miss this generality if an implementation doesn't have it.

Technically it does seem like a backwards incompatibility to make this restriction, but for the reasons explained above, it would be a backwards incompatibility that no one would notice.

That is all simply background for my position that if we continue to allow this generality in the language spec, perhaps simply by being silent on the subject as the spec has been so far, and also continuing not to implement it in open source p4c, that seems like a low effort viable approach.  We could add this restriction to the short list of things that open source p4c does not support, and then in practice unless someone goes to that non-trivial effort to implement it, it is unlikely that any other P4 implementation will, either.

jnfoster:
At the April '22 LDWG, we agreed to add a restriction to forbid invoking tables from action arguments.

jafingerhut:
Proposed alternative PR that makes fewer new restrictions here: https://github.com/p4lang/p4-spec/pull/1043

mbudiu-vmw:
Closing in favor of the alternative PR.

PR #853, opened 2020-05-05
Allow concatenation with any type of bitstrings

mbudiu-vmw:


mbudiu-vmw:
The implementation already works this way.

mbudiu-vmw:
I will give some more time for people to review, but I think we can just merge this for 1.2.1.

PR #860, opened 2020-05-14
Clarify exit execution during expression evaluation

jafingerhut:


jafingerhut:
I do NOT consider this a minor clarification or bug fix.  It represents a change that could be made to the specification to clarify the behavior of `exit` statements that are invoked during the evaluation of an expression, if the P4 LDWG decides to permit such expressions in the language.

I understand that such epxressions were not explicitly forbidden before now, and thus may be considered as permitted implicitly.  The combination of language features seems a bit subtle to me, and I believe deserves at least a brief discussion in a future LDWG meeting before deciding what to do about this PR.

jafingerhut:
Understood.  Do you think it is a good idea to try to be exhaustive of all of the expression evaluation cases here?  Or should it be enough to mention one or two examples, and refer to the section on the order of expression evaluation (which there is such a cross-ref in this proposed PR already)?

mbudiu-vmw:
This is fine.

fruffy:
Exit statements in function bodies are currently forbidden according to the compiler front-end. Does this change mean that exit statements are allowed there?

mbudiu-vmw:
No, exit statements should not be allowed in functions.
Now I regret that we have them in actions as well, but I think I have a scheme to make them work.

fruffy:
I am very okay with this, exit statements have been a major pain to implement.

jafingerhut:
@jnfoster @mbudiu-vmw This is the last issue I will ping you on before the next LDWG meeting, I promise :-)   I don't see anything necessarily more urgent than any other issue or PR facing the LDWG, just wanted to call attention to it to see if anything should be changed before we decide whether to merge it, or abandon it.

jnfoster:
Update from the 2/7/22 meeting. Folks will read this and we won't discuss it again, assuming there are no objections.

jnfoster (approved):
LGTM

mbudiu-vmw (approved):
I expect in fact that this should work like this in the compiler even today, since the expression with a side-effect involving an apply will be moved into a separate statement.

jnfoster:
Ok, shall we merge this? I think so.

mbudiu-vmw:
2 approvals should be enough

PR #861, opened 2020-05-16
Clarifications for directionless parameters, and parameters with type…

jafingerhut:
… int

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1500,16 +1500,15 @@ Each parameter may be labeled with a direction:
   entity the parameter is for:
   - For anything other than an action, e.g. a control, parser, or
     function, a directionless parameter means that the value supplied
-    as an argument in a call must be a compile-time known value.
-  - For an action, a directionless parameter indicates:
-    - If the action appears in an `actions` table property (Section
-      [#sec-table-action-list]), then the parameter's value must be
-      provided by the control plane software when an entry is added to
-      the table that uses that action.
-    - If the action is called in some other way, e.g. directly from
-      within a control, then the caller must provide a value, and the
-      call behaves the same as if the direction were `in`.  Such an
-      argument expression need not be a compile-time known value.
+    as an argument in a call must be a compile-time known value
+    (see Section [#sec-ct-constants]).
+  - For an action, a directionless parameter indicates that either:
+    - The parameter's value is provided in the P4 program.  In this
+      case, the parameter behaves as if the direction were `in`.  Such
+      an argument expression need not be a compile-time known value.
+    - The parameter's value is provided by the control plane software
+      when an entry is added to a table that uses that action.  See
+      Section [#sec-actions].
 
 Direction `out` parameters are always initialized at the beginning of
 execution of the portion of the program that has the `out` parameters,
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1500,7 +1500,7 @@ Each parameter may be labeled with a direction:
   entity the parameter is for:
   - For anything other than an action, e.g. a control, parser, or
     function, a directionless parameter means that the value supplied
-    as an argument in a call must be a compile time known value.
+    as an argument in a call must be a compile-time known value.
   - For an action, a directionless parameter indicates:
     - If the action appears in an `actions` table property (Section
       [#sec-table-action-list]), then the parameter's value must be
@@ -1509,7 +1509,7 @@ Each parameter may be labeled with a direction:
     - If the action is called in some other way, e.g. directly from
       within a control, then the caller must provide a value, and the
       call behaves the same as if the direction were `in`.  Such an
-      argument expression need not be a compile time known value.
+      argument expression need not be a compile-time known value.
 
 Direction `out` parameters are always initialized at the beginning of
 execution of the portion of the program that has the `out` parameters,
@@ -2086,7 +2086,7 @@ const int c = b - a + 3;
 Parameters with type `int` are not supported for actions.  Parameters
 with type `int` for other callable entities of a program,
 e.g. controls, parsers, or functions, must be directionless,
-indicating that all calls must provide a compile time known value as
+indicating that all calls must provide a compile-time known value as
 an argument for such a parameter.  See Section
 [#sec-calling-convention] for more details on directionless
 parameters.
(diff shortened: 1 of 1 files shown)


jafingerhut:
This PR is intended to be a clarification of the existing specification, not a change in the meaning of anything.  I am not expecting that it should be made before the next release.  I wrote it as a response to this p4c issue: https://github.com/p4lang/p4c/pull/2360

hesingh:
Please change "compile time" to "compile-time".

Maybe, you can mention a reason why an `int` is not support as parameter to an action?  P4Runtime allows only `bit<W>` and `int<W>`.

jafingerhut:
Done on the changes to 'compile-time'.  That is definitely more consistent with the rest of the spec.

Regarding this: "Maybe, you can mention a reason why an int is not support as parameter to an action?"

I believe the reason is that there is no way for actions to specify that one of their parameters must be a compile-time known value.  For controls, parsers, functions, etc. there is a way: make a parameter directionless.  For actions, directionless parameters mean something different.

That is explained in one of the parts of the text that is changed by this PR.  I believe those changes are just another way of saying what was already there, but hopefully more explicit.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1496,12 +1495,21 @@ Each parameter may be labeled with a direction:
...
+    function, a directionless parameter means that the value supplied
+    as an argument in a call must be a compile-time known value.
+  - For an action, a directionless parameter indicates:
+    - If the action appears in an `actions` table property (Section
+      [#sec-table-action-list]), then the parameter's value must be
+      provided by the control plane software when an entry is added to
+      the table that uses that action.
+    - If the action is called in some other way, e.g. directly from
+      within a control, then the caller must provide a value, and the
+      call behaves the same as if the direction were `in`.  Such an
```
mbudiu-vmw, line 1511: Actually I don't think that this is true, they should still be a compile-time constant.

jafingerhut: I do not see anything else in the spec that implies this about actions with directionless parameters.  p4c as of a very recent version accepts programs that call such actions with run-time variable expressions as action arguments.  I can compile attached try3.p4.txt program without error using this version (I may need to attach that program to a different comment -- Github is complaining about attaching it to this comment):
```
$ p4c --version
p4c 1.2.0 (SHA: 98a890767 BUILD: DEBUG)
```

jafingerhut:
Small test program demonstrating that latest p4c handles run-time variable expressions as arguments to correspond with an action parameter declared as directionless, when the action is called directly from a control.
[try3.p4.txt](https://github.com/p4lang/p4-spec/files/4652541/try3.p4.txt)

p4-16/spec/P4-16-spec.mdk
jafingerhut: I understand that this might be a case where p4c is doing more than is allowed by the spec.  The test program and p4c behavior is background that perhaps this new proposed spec text might be correct as written.  I do not know what other P4 compilers do with this yet.

mbudiu-vmw:
I think we had to do this because initially there were no functions, and some actions may be both inserted by the control-plane and called by other actions. But functions and actions are still different, e.g., you cannot call exit in functions, so we need to preserve this use case. Your clarifications are reasonable.

mbudiu-vmw:
Make sure this covers extern instances.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1496,12 +1495,21 @@ Each parameter may be labeled with a direction:
...
+- The meaning of parameters with no direction depends upon the kind of
+  entity the parameter is for:
+  - For anything other than an action, e.g. a control, parser, or
+    function, a directionless parameter means that the value supplied
+    as an argument in a call must be a compile-time known value.
+  - For an action, a directionless parameter indicates:
+    - If the action appears in an `actions` table property (Section
+      [#sec-table-action-list]), then the parameter's value must be
+      provided by the control plane software when an entry is added to
+      the table that uses that action.
```
jafingerhut, line 1508: @jafingerhut The phrase "must be provided by the control plane software ..." should be edited, perhaps to refer to some of the new text in this PR: https://github.com/p4lang/p4-spec/issues/914

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1496,12 +1495,21 @@ Each parameter may be labeled with a direction:
...
-	- a compile-time known value
-	- an action parameter that can only be set by the control plane
-	- an action parameter that can be set directly by another calling
-      action; in this case it behaves like an `in` parameter
+  an `inout` parameter must be an l-value.
+- The meaning of parameters with no direction depends upon the kind of
+  entity the parameter is for:
+  - For anything other than an action, e.g. a control, parser, or
+    function, a directionless parameter means that the value supplied
+    as an argument in a call must be a compile-time known value.
```
jafingerhut, line 1503: @jafingerhut Add a cross-reference to the existing spec section describing what "compile-time known value" is.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Updated the text to make it more consistent with #914.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Added cross-reference.

jafingerhut:
> Make sure this covers extern instances.

I added a cross-reference to the section on compile-time known values, which includes this as one of the cases: "Instances constructed by instance declarations (Section 10.3) and constructor invocations."

mbudiu-vmw:
Should we merge this?

jafingerhut:
I should probably do a small amount of editing first to make some of the text more consistent with this already-merged change: https://github.com/p4lang/p4-spec/commit/563e727ba238e0ffb4912d76e79d59ef58a83bbf
i.e. add some more words about where directionless parameters for an action can come from, and perhaps a cross section to the section added in the commit I just linked for a full list of possibilities.

jafingerhut:
@mbudiu-vmw Please take a look at the changes with my latest commit added on 2021-Jun-02, to see if it looks clear and correct.  It includes a mention of the term "action data", which was added in the last few months to the spec.

mbudiu-vmw:
I think that this can be merged at any time, since it is a clarification and not a change.

jfingerh:
@jnfoster @mbudiu-vmw Wanted to call attention to this PR in case we think it is worth any other discussion in a LDWG meeting, and/or whether we think it is ready to be merged into the spec.

jnfoster:
Thanks. We agreed to take this small clarification already. Merging it....

Issue #868, opened 2020-06-05
Update p4-16 spec to include serializable enums for select expressions in value sets (12.6)


amreshk:
## Personnel
- [ ] **Owner**: 
- [x] **Supporters**: @amreshk 

## Design
- [x] **Document**: see below

## Implementation
- [ ] **`p4-spec`**: 
- [x] **`p4c`**: https://github.com/p4lang/p4c/pull/2415

## Process
- [ ] **LDWG discussed**: 
- [ ] **LDWG approved**:
- [ ] **Merged into `p4-spec`**:
- [ ] **Merged into `p4c`**: 

====================

According to `p4-16v1.2.0` spec, `select` expressions on targets which support `parser value set` must be either `bit<>`, `tuple` or `struct` type.
https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-select

> Some targets may support parser value set, see Section 12.11. Given a type T for the type parameter of the value set, the type of the value set is set. The type of the value set must match to the type of all other keysetExpression in the same select expression. If there is a mismatch, the compiler must raise an error. The type of the values in the set must be one of bit<>, tuple, and struct.

This description should also include `serializable_enum` as a supported type.

Following example illustrates usage,

```
// Serializable enum type
enum bit<16> value_set_t {
    ABCD       = 0x8100,
    EFGH       = 0x0800
}

parser MyParser(packet_in b, out my_packet p, inout my_metadata m, inout standard_metadata_t s) {

    value_set<value_set_t>(4) pvs;

    state start {
        b.extract(p.data);
        transition select(p.data.da) {
            pvs   : accept;
            0x100 : foo;
        }
    }

    state foo {
        transition accept;
    }
}
```

Related PR - https://github.com/p4lang/p4c/pull/2415

hesingh:
Please add `int<>` to this sentence in the spec.

"The type of the values in the set must be one of bit<>, tuple, and struct."

The reason is because both `int<W>` and `bit<W>` map internally in p4c to `Type_Bits`.

mbudiu-vmw:
We should check whether the expansion of ranges of signed values into ternary matches is correct.

jnfoster:
In the interest of tidying up the set of active issues on the P4 specification repository, I'm marking this as "stalled" and closing it. Of course, we can always re-open it in the future if there is interest in resurrecting it.

PR #873, opened 2020-06-15
Update parser value set types

hesingh:
See https://github.com/p4lang/p4-spec/issues/868


jafingerhut:
This seems like a reasonable generalization to me.  I suspect that taking advantage of this generality in the language might require similar generalizations in the P4Runtime API and/or TDI, but I don't see a problem if the language spec "run ahead" of those other things.

jnfoster:
@jafingerhut and @jnfoster will investigate and propose a list of types allowed, consistent with extant implementations, P4Runtime, and as general as possible.

Issue #864, opened 2020-05-24
Tuples: no elimination form


jnfoster:
## Personnel
- [x] **Owner**: @jnfoster
- [x] **Supporters**: @mbudiu-vmw

## Design 
- [x] **Document**: see below

## Implementation
- [x] **`p4-spec`**: https://github.com/p4lang/p4-spec/pull/877
- [x] **`p4c`**: https://github.com/p4lang/p4c/pull/2655 this implementation names fields f[0], f[1].
Old version: https://github.com/p4lang/p4c/pull/2451 this implementation names fields f0, f1, f2.

## Process
- [x] **LDWG discussion**: https://docs.google.com/document/d/1XSgdXeG1UuF1FM_XAqxDrHeN4dHZWBnJPKVS6SnGNwM/edit#heading=h.lggdh8yt82ei, at the July LDWG
- [x] **LDWG approved**
- [x] **Merged into `p4-spec`**
- [x] **Merged into `p4c`**

Frank Pfenning from CMU is one of the world's experts on type systems. He has a set of beautiful design principles that provide guidance when developing new type systems. One of these principles, called [harmony](https://www.cs.cmu.edu/~fp/courses/15317-f09/lectures/03-harmony.pdf), governs how the introduction forms and elimination forms for a type interact. An introduction form is an expression that lets you create a value of a given type. An elimination form is an expression that lets you destruct or access the components of a value of a given type.

Unfortunately, unless I missed it, when we added tuples to the language, we forgot to add any elimination forms. So if you create a tuple, you no longer have a way to access its components. (The one exception is in parsers, where a `select` can be used to destruct a tuple and match it against one or more key sets; but this can't be done in controls.) 

```p4
control c();
package p(c c);

control myc() {
  tuple<bool, bit<8>> y;
  apply {
    y = { true, 0 };
    // y's components can't be accessed any more
  }

}

p(myc()) main;
```

To fix this, I think we need to add one of the following to the langauge:
* Casts and/or assignments between tuples and structs (acceptable option) with compatible fields (an OK option, but clunky, and implicitly treats structs as ordered, so I don't like it.)
* Projection operations on tuples (a good option; what syntax?)
* Extend `switch` to support pattern matching on tuples (a great option; already discussed as a possible language extension in an appendix).

(H/T to the Neptune team at @cornell-netlab who figured this out: @piercedouglis, @minmit, @rachitnigam)

mbudiu-vmw:
[...]

PR #877, opened 2020-06-24
Tuple fields

mbudiu-vmw:
This small PR introduces tuple field extraction. Initially I was thinking to use a syntax like `t.0` to extract a field, but using `t.f0` seems equally good and requires no changes to the grammar, and will not introduce ambiguities.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -7971,6 +7971,7 @@ The P4 compiler should provide:
 
 ## Summary of changes made in version 1.3.0
 
+* support for tuple field access (Section [#sec-tuple-exprs]).
 * Support for generic structures (Section [#sec-type-spec]).
 * Generalize `switch` statements to allow expressions with integer,
   `enum`, or `error` types (Section [#sec-switch-stmt]).
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3749,7 +3749,7 @@ list expressions.
 tuple<bit<32>, bool> x = { 10, false };
 ~ End P4Example
 
-The fields of a tuple can be accessed using array index sytnax `x[0]`,
+The fields of a tuple can be accessed using array index syntax `x[0]`,
 `x[1]`.  The array indexes *must* be compile-time constants, to enable
 the type-checker to identify the field types statically.
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2488,7 +2488,7 @@ struct Parsed_headers {
 ### Tuple types { #sec-tuple-types }
 
 A tuple is similar to a `struct`, in that it holds multiple
-values. Unlike a `struct` type, tuples have no named fields. The
+values. The fields of a tuple are named in order `f0`, `f1`, etc.. The
 type of tuples with n component types `T1`,...,`Tn` is written
 as
 
@@ -3750,6 +3750,9 @@ list expressions.
 ~ Begin P4Example
 tuple<bit<32>, bool> x = { 10, false };
 ~ End P4Example
+
+The fields of a tuple are named `f0`, `f1`, etc,; they can be
+extracted using the same syntax as structure fields: `x.f0`.
 
 ## Operations on lists { #sec-list-exprs }
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2487,10 +2487,8 @@ struct Parsed_headers {
 
 ### Tuple types { #sec-tuple-types }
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
Perhaps this is intended by design, but this proposal syntactically disallows the ability to have a run-time variable expression, or even a named compile-time constant (e.g. #define symbol, or `const int x = 2;`) as the index of the tuple element reference.

Using syntax like `my_tuple[0]` or similar, even if most or all implementations would likely limit the expression to a compile-time known value that is `int` (also `bit<W>` or `int<W>`), would enable code like:
```
#define MY_FAVORITE_INDEX 2
const int second_favorite_index = 5;

     a = my_tuple[MY_FAVORITE_INDEX];
     b = my_tuple[second_favorite_index];
```

mbudiu-vmw:
Yes, this is by design, with a dynamic index we cannot statically tell the field type.

jafingerhut:
So compile-time known value is needed, understood.  But an expression still might be useful for developers?  We allow compile-time known value expressions in bit slices today, and in place of the `W` in `bit<W>` types.

mbudiu-vmw:
This syntax won't work for expressions.

mbudiu-vmw:
Fixes #864

jnfoster:
I approve but let's discuss at the July LDWG.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3701,6 +3701,9 @@ list expressions.
 tuple<bit<32>, bool> x = { 10, false };
 ~ End P4Example
 
+The fields of a tuple are named `f0`, `f1`, etc,; they can be
```
jnfoster, line 3704: Should be "etc." not "etc,"

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2486,7 +2486,7 @@ struct Parsed_headers {
 ### Tuple types { #sec-tuple-types }
 
 A tuple is similar to a `struct`, in that it holds multiple
-values. Unlike a `struct` type, tuples have no named fields. The
+values. The fields of a tuple are named in order `f0`, `f1`, etc.. The
```
jnfoster, line 2489: Should be "etc." not "etc.."

liujed:
+1 for array indexing syntax, even if the indices are restricted to compile-time constant expressions.

jnfoster:
Actually, upon reflection, I think array indexing is much better than `f0`, `f1`, etc.

jnfoster (changes requested):

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: Let's see if can implement an alternative prototype for this.

vgurevich:
Would it make sense to introduce optional field names that can be defined by the programmer, while using `f0`, `f1`, etc. as a backup?

mbudiu-vmw:
@vgurevich can you elaborate on your proposal?

vgurevich:
@mbudiu-vmw -- something like:
```
result = Hash.get({src_ip = hdr.ipv4.src_addr, dst_ip=hdr.ipv4.dst_addr, potocol=hdr.ipv4.proto});
```

mbudiu-vmw:
This is an anonymous struct, not a tuple. We have an issue filed for supporting anonymous structs, and you are listed as an owner.

mbudiu-vmw:
The spec should say whether tuple fields can be left values.

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: Technically one dot is for the etc abbreviation, and the second is the sentence-ending dot.

p4-16/spec/P4-16-spec.mdk
liujed-intel: > Technically one dot is for the etc abbreviation, and the second is the sentence-ending dot.

That might seem logical, but English defies logic in many ways. https://english.stackexchange.com/questions/8382/when-etc-is-at-the-end-of-a-phrase-do-you-place-a-period-after-it/8385#8385

jnfoster: The Chicago Manual of Style agrees :-)

https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Punctuation/faq0040.html

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3746,6 +3744,15 @@ list expressions.
 tuple<bit<32>, bool> x = { 10, false };
 ~ End P4Example
 
+The fields of a tuple can be accessed using array index sytnax `x[0]`,
```
blp, line 3747: sytnax -> syntax

mbudiu-vmw:
@jnfoster : made the requested typo fix. If you approve we can merge this.

PR #883, opened 2020-07-28
Support for parsing varbit values

mihaibudiu:
This is done by just adding a new method to packet_in; this essentially converts a varbit value into a new packet_in which can be parsed using the existing mechanisms.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3569,6 +3569,10 @@ Variable-length bit-strings support a limited set of operations:
   Two `varbit` fields can be compared only if they have the same type.
   Two varbits are equal if they have the same dynamic width and all
   the bits up to the dynamic width are the same.
+- Initialization of a `packet_in` object from a varbit value, by using the
+  `initialize` method of a packet_in extern instance.
+  This allows using the traditional methods of `packet_in` to be used
+  for parsing the data in a varbit field.
 
 The following operations are _not_ supported directly on a value of
 type `varbit`, but instead on any type for which `extract` and `emit`
@@ -3585,6 +3589,9 @@ finding this information in a section dedicated to type `varbit`.
   contain a field with type `varbit`. Such an `emit` method call
   inserts a variable-sized bit-string with a known dynamic width into
   the packet being constructed.
+
+An example that uses a parser to extract data from a varbit field is given
+in Section [#sec-parsing-varbits].
 
 ## Casts { #sec-casts }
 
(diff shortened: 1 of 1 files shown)


jnfoster:
@jafingerhut please take a look and see if this is in harmony with Intel efforts.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5715,6 +5724,36 @@ the time budget allocated for parsing, the parser should transition to
...
+
+parser top(packet_in b, out Parsed_packet p) {
+   packet_in buffer;
+   parse_options options_parser();
+   // some states omitted
+
+   state parse_ipv4 {
+       b.extract(headers.ipv4);
+       b.extract(headers.ipv4options,
+                 (bit<32>)(((bit<16>)headers.ipv4.ihl - 5) * 32));
```
vgurevich, line 5746: Isn't that overly complicated (with two casts)? :)

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3569,6 +3569,10 @@ Variable-length bit-strings support a limited set of operations:
   Two `varbit` fields can be compared only if they have the same type.
   Two varbits are equal if they have the same dynamic width and all
   the bits up to the dynamic width are the same.
+- Initialization of a `packet_in` object from a varbit value, by using the
+  `initialize` method of a packet_in extern instance.
+  This allows using the traditional methods of `packet_in` to be used
+  for parsing the data in a varbit field.
```
vgurevich, line 3575: I am a little worried about a variety of cases, where 
1. `varbit` (or any other source field is not byte-aligned)
2. When `varbit` length (in bits) is not divisible by 8
3. What happens when not the whole field is parsed (esp. how will deparsing be defined)

mbudiu-vmw: These cases are not different from parsing a packet.
Deparsing is not defined in this proposal, but it could be defined by attaching a varbit to a packet_out object. 
One would have to specify what happens on varbit overflow.

vgurevich:
Just a thought: wouldn't you agree that if we allow parsing of varbits, then we should allow bit slices of varbits at least as rvalues?

vgurevich:
As long as we are OK with functions instead of methods or operations, then, yes, this would be the safest and least intrusive approach (although for parsing you would first need to create an extern to hold the state (such as the imaginary pointer to the current position for the parsing).

jnfoster:

Issue #885, opened 2020-08-06
Be consistent about `miss` in `apply_table(T)`


blp:
In `#sec-invoke-mau` and `#sec-mau-semantics`, the `miss` field in `apply_table(T)` is inconsistently included in the discussion. The attached patch should make what I mean clear (and possibly fix it).
[0001-P4-16-spec-Consistently-consider-the-miss-field-in-a.txt](https://github.com/p4lang/p4-spec/files/5037541/0001-P4-16-spec-Consistently-consider-the-miss-field-in-a.txt)

jnfoster:
Nicely spotted. Do you want to go ahead and make a proper PR on the p4-spec repository?

blp:
> Nicely spotted. Do you want to go ahead and make a proper PR on the p4-spec repository?

Done: pull #886

PR #886, opened 2020-08-07
P4-16-spec: Consistently consider the 'miss' field in apply_result(T).

blp:
Some places this is mentioned, otherwise it is not.

Issue #885.


jafingerhut (approved):
Looks good to me.

jnfoster:
This is a bug fix, so let's just merge it.

PR #887, opened 2020-08-24
Generalize switch statements to have expressions with other types

jafingerhut:


jafingerhut:
At the August 2020 meeting with advanced topics / new proposals for P4_16, I presented a much more incremental addition to P4_16: one way to add tagged unions to the language.

Part of that was generalizing the switch statement to allow other types of expressions, which seems useful in its own right, independent of tagged unions.  This PR includes _only_ changes for generalizing switch statements, and nothing about tagged unions.

While the diffs show lots of text changes in the section on switch statements, I have tried carefully not to change the behavior or restrictions of the switch statement that were there before when the switch expression is `t.apply().action_run`.  If you ignore the new possible types of the switch expression, the meaning should be exactly what it was before these changes.

jafingerhut:
Note that as far as implementation in p4c goes, I believe that several targets will be able to implement switch expressions more efficiently than the equivalent 'daisy chain' of if-then-else if statements, because of the guarantee that all switch cases are mutually exclusive, and have labels that are compile-time known values.  Thus it would be best if p4c preserved such new switch statements in p4c's IR, and not replace them with if-then-else if statements.

Mihai did a quick look at the IR during the meeting where I presented this, and he commented at that time that it appears that the current p4c IR for switch statements allows switch expressions to have other types.  If so, I hope that helps make it easier to implement p4c changes for this extension to the language, while preserving switch statements in the IR.

vgurevich:
Anything, that allows the programmer to express their **actual** intent more clearly and unambiguously is a good thing. As @jafingerhut said, one thing it will certainly do is to allow compiler to better deduce mutual exclusivity of certain code paths (if-then-elseif-elseif-else chains are OK either way, but sequences of if-then statements are more difficult to prove to be mutually exclusive).

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4946,21 +4946,69 @@ switchCase
...
+    | expression
     | DEFAULT
     ;
 ~ End P4Grammar
 
-The expression within the `switch` statement is restricted to be
-the result of a table invocation (See Section [#sec-invoke-mau]).
-
-If a switch label is not followed by a block statement it falls
+The expression within the `switch` statement must be one of:
```
mbudiu-vmw, line 4954: I would probably make two separate sub-sections for these, I think that the explanations would be easier to follow.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4946,21 +4946,69 @@ switchCase
...
+  - `enum`, either with or without an underlying representation specified
+  - `error`
+
+If the `switch` expression is of the form `t.apply().action_run`, then
+all switch labels must be names of actions of the table `t`.
+Otherwise, all switch labels must be expressions with compile-time
+known values, and have the same type as the type of the switch
+expression.
+
+TBD: What language to use above if we want to allow a switch
```
mbudiu-vmw, line 4971: you could say that they must have a type which can be implicitly cast to the switch expression type.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4946,21 +4946,69 @@ switchCase
...
+evaluated, and any side effects from evaluating this expression are
+visible to any `switch` case that is executed.  Among switch labels
+that are not `default`, at most one of them can equal the value of the
+switch expression. If one is equal, that switch case is executed.
+
+If no labels are equal to the `switch` expression, then:
+
+- if there is a `default` label, the case with the `default` label is
+  executed.
+- if there is no `default` label, then no switch case is executd, and
```
mbudiu-vmw, line 4997: typo

p4-16/spec/P4-16-spec.mdk
jafingerhut: Thanks for that.  Incorporated that terminology into the latest commit.

p4-16/spec/P4-16-spec.mdk
jafingerhut: fixed.  Thanks.

p4-16/spec/P4-16-spec.mdk
jafingerhut: I have made a try at following this suggestion, but I do not know if it is any easier to follow.  Take a look and see what you think.

jfingerh:
As some background details for questions that might come up in P4 language design work group meetings about the current behavior of the switch statement, I have made a list of test programs in p4lang/p4c that exercise various kinds of switch statements: https://github.com/jafingerhut/p4-guide/tree/master/switch-statement

jafingerhut:
Note that this sentence in this PR (after 4 commits it was still there, at least), is new, and perhaps controversial: "The last switch label must be followed by a block statement."

The current p4c implementation as of 2020-Sep-04 seems to treat a sequence of one or more labels at the end of a switch statement, with no following blocks statement, as if you had written `{ }` as the blocks statement after them.  I would recommend that the LDWG decide on either this as the explicitly defined meaning of such syntax, or declare it a compile-time error: https://github.com/p4lang/p4c/issues/2527

jafingerhut:
A suggestion for how the proposed generalized switch statement might be implemented, by transforming the new switch statements into the existing switch statements, is described here: https://github.com/jafingerhut/p4-guide/blob/master/switch-statement/implementing-generalized-switch-statements.md

It does not seem to me that this belongs as part of the P4_16 language specification itself, but if it seems useful to include it as a companion document in this repository, perhaps linked from the spec, I could arrange that.

mbudiu-vmw:
Would you like to try your hand at implementing this as a midend pass?

jafingerhut:
@mbudiu-vmw I am not sure, but I suspect it may be faster if I work at persuading someone at Intel to implement it, versus trying to implement it myself :-)

mbudiu-vmw:
You would have no problem persuading me to do it, but I thought this could be a very good exercise for learning more about the compiler internals. I can provide assistance. First, of course, we would have to add support for generalized switch statements, and then this elimination scheme.

jafingerhut:
At the 2020-Sep-14 language design work group meeting, several possible additional generalizations to switch statements were discussed.  I will create a separate issue for those, and a separate PR.  At least Mihai suggested that this PR might move forward as a first step towards an even more general switch statement, with the further generalizations done as a separate PR after this one was approved and merged.

I have made one additional commit after that meeting, which is to explicitly say that if the last switch label has no body after it, the behavior is the same as if there were an empty body `{ }` after it.  That is the behavior that p4c implements today.  There did not seem to be any strong opinions on whether this case should be legal or not, so this latest commit is the conservative approach of explicitly defining the current p4c behavior, rather than changing it to be illegal.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4946,21 +4946,20 @@ switchCase
 
 switchLabel
     : name
+    | expression
```
mbudiu-vmw, line 4949: I am trying to implement this, but the grammar becomes ambiguous, because now if you see a { after a colon you no longer know whether it's a list expression, part of the next label, or a block statement. So we may need to add some other disambiguating factor, like a 'case' label.

mbudiu-vmw:
Implemented here: https://github.com/p4lang/p4c/pull/2575

mbudiu-vmw:
I reworked the implementation to not require 'case'; the modified grammar is in the compiler pull request. If this PR is accepted we should amend the grammar as shown in the code changes.

jfingerh:
@mbudiu-vmw At the last P4 language design work group meeting, you asked about adding an entry to the list of changes section near the end.  Just a few minutes ago I pushed commit #8 on this PR that adds such a section.  There was not yet such a section for the next version, so I semi-arbitrarily named the section with version 1.2.2 -- that can be edited later if a different next version number is desired.

mbudiu-vmw:
@jafingerhut I think we agreed to merge this change in the spec after you apply some minor fixes. I would also suggest using version 1.3.0, since this is what I used for the `if in parsers` commit I just merged, so they end up in the same section, which we can rename later.

jafingerhut:
@mbudiu-vmw I am unaware of any fixes proposed to this PR except the most recent one, which was to add a change log, which has been done, and I just pushed a change to make the version number 1.3.0 as you suggested.

mbudiu-vmw:
I wasn't sure whether the note on the implementation stays external to the spec or not.

mbudiu-vmw:
If you are happy you can in principle merge this.

jafingerhut:
@mbudiu-vmw I will merge it in as it is now (with next spec version being 1.3.0 as you recommend).

The implementation note in a separate file is linked by URL from the spec, which seems like the best approach to me, given that it is only a description of one possible way to implement this language feature, and isn't a requirement of any kind.

PR #920, opened 2021-03-12
Add match_kind optional to PSA specification

jafingerhut:


jnfoster (approved):
Looks good!

mbudiu-vmw (approved):

konne88:
Do we also have to update this file?
https://github.com/p4lang/p4c/blob/master/p4include/psa.p4#L400

jafingerhut:
@konne88 To support this in p4c, yes that psa.p4 include file would need to be updated, too.

konne88:
Thanks Andy. Once we add it to the p4c include file, do you know if it would it just work, or would we also have to change the compiler? 

Would we submit this PR first, and then send the followup PR to the p4c repo?

jafingerhut:
I think submitting it for the p4c compiler would be reasonable now, perhaps with a note that it is probably best to wait for this PR to be merged before merging that one.

There is only an incomplete open source p4c back end that supports the PSA architecture for behavioral-model (a process called psa_switch, instead of simple_switch) today, and it fails to support so many other PSA features that if it fails to support match_kind optional for a while longer, I don't think anyone will notice.  PSA architecture implementations for other targets may very well beat that one to completion. (help welcome on the BMv2 PSA support, if you know of anyone interested)

jafingerhut:
There was discussion of this topic in the P4 LDWG meeting early April, with no objections for adding this to the PSA.  There was discussion of what a prerequisite might be for adding match_kind `optional` to core.p4, i.e. the base language specification, but that can be more fully discussed later.  Adding this to PSA now is useful and desirable.

PR #921, opened 2021-03-12
Update list of P4 architecture working group chairs

jafingerhut:


jafingerhut:
Just happened across this old contact info while looking at something else.  Let me know if this is the email address you prefer, Gordon.

Issue #915, opened 2021-02-26
What types can be used as indexes into header stacks?


jnfoster:
The spec doesn't say... for example, is `varbit` allowed?

jfingerh:
For compile-time known values, I would think that a reasonably conservative proposal would explicitly support only these types:
+ int
+ bit<W> for any W >= 1
+ int<W> for any W >= 1
+ Any enum type defined with an explicit storage representation that was one of the above
For header stack index expressions that are not compile-time known values, they could not be type int, of course.

jnfoster:
@jnfoster and @jafingerhut will take this design and flesh it out into a full-fledged PR.

mbudiu-vmw:
Fixed in https://github.com/p4lang/p4c/pull/2738

PR #923, opened 2021-03-14
Address spec issue #915 - Specify the types allowed for header stack indexes

jafingerhut:


jafingerhut:
@jnfoster Here is a proposed change for issue #915.

I have no strong preference for or against allowing serializable enum types as header stack indexes.  It seems reasonable to me to allow them, or not.

The rest seems fairly safe and conservative to me.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4111,6 +4111,17 @@ expressions are legal:
...
+  The `index` is an expression that must be one of the following types:
+
+  - `int` - an infinite-precision integer (section
+    [#sec-infinite-precision-integers])
+  - `bit<W>` - a `W`-bit unsigned integer where `W >= 1` (section
+    [#sec-unsigned-integers])
+  - `int<W>` - a `W`-bit signed integer where `W >= 1` (section
+    [#sec-signed-integers])
+  - a serializable `enum` with an underlying type that is `bit<W>` or
+    `int<W>` (section [#sec-enum-types]).
```
jafingerhut, line 4123: @jafingerhut AI Andy to create a small test program and see whether p4c accepts a header stack index with type of serializable enum is supported today.

mbudiu-vmw:
Check whether the implementation supports enums.
Probably this should be accepted.

p4-16/spec/P4-16-spec.mdk
jafingerhut: A simple test program shows that recent versions of p4c given an error when one attempts to use an expression of type `enum bit<8> ...` as the index of a header stack reference: https://github.com/p4lang/p4c/issues/2726

mbudiu-vmw (approved):
The compiler now supports all these cases: https://github.com/p4lang/p4c/pull/2738

jafingerhut:
@mbudiu-vmw I am pretty sure we agreed to merge this into the spec during the 2021-May-03 LDWG meeting.  It had a conflict (in the change log section, as expected) which should now be resolved.  I will let you push the merge button for this, in case I am mixed up on PRs.

mbudiu-vmw (approved):

Issue #918, opened 2021-03-08
Request to make license explicit for p4lang/p4-spec repository files


jafingerhut:
Someone asked me in a personal message about the license of the files in this repository.  It would make sense if it was intended to be Apache 2 like it is for other p4lang project code/files, but except for this mention in the README, and many of the PSA example programs, I don't see any mention of Apache in the files of this repository:
```
Only members of the P4.org community may propose changes to the P4 language specification, and all contributed changes will be governed by the Apache-style license specified in the P4.org membership agreement.
```
In particular the P4_16 spec grammar.mdk file has no mention of license inside of it.  Is it desired to have a separate mention of license in each such source file?  Rely on the top level README mention above?  Perhaps add a separate LICENSE file at the top level and have the README link to it?  Something else?

I'm happy to add an explicit license notice to the grammar.mdk file in a PR, if that seems reasonable, plus any other non-Madoko source-program-like files in this repository, if that is decided as a reasonable change to make.

mbudiu-vmw:
Often for content one uses Creative Commons licenses rather than Apache.

jnfoster:
Good question. Let me bring this to the TST.

PR #924, opened 2021-03-14
Add top level file LICENSE copied from p4lang/p4c/LICENSE

jafingerhut:
containing the Apache 2.0 license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,176 @@
+  Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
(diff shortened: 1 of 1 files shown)


jafingerhut:
This is one possible way to address #918 - Add a top level file LICENSE to this repository that contains the Apache 2.0 license.  @jnfoster plans to review the question of the license for the P4_16 language specification by the TST, so this should not be merged in unless that group approves of this plan.

mbudiu-vmw (approved):
What does this license apply to?
Spec? PSA spec? Source files?
Shouldn't we use a contents license?
https://en.wikipedia.org/wiki/List_of_free-content_licences

jafingerhut:
The same thing it applies to in this repository :-).  https://github.com/p4lang/p4runtime

In other words, I don't know.  One interpretation is it applies to every file in the repository.  As I mentioned in my comment above, Nate has taken the action item to decide whether some of the files, e.g. documentation, might better be distributed under a different license than Apache.  By this PR, I am not trying to make this decision unilaterally.  This PR can be closed if P4.org decides that some other changes should be made instead.

jfingerh:
@jnfoster @mbudiu-vmw Just a brief reminder of this issue.  I do not know the answers to Mihai's questions in a comment above about what license SHOULD be used for p4.org specifications, but if Apache 2.0 is what we want, then adding a LICENSE file to this repo is at least one way to go about making that explicit.  I am happy to add notes to the README pointing at this license file, and explaining which files it applies to, and which it does not, but it seems to me that it would be reasonable to apply Apache 2.0 to ALL files in this repo, documentation and code alike.  If we think specs and documentation should be under a different license than Apache 2.0, then who decides that, and how?

jnfoster:
@bocon13 was going to check with past contributors about this. Brian, are we clear to add a license? 

If so, I think Apache2 is best for uniformity with other p4lang projects, even considering the comments on Creative Commons licenses.

bocon13:
@jnfoster @jafingerhut, we notified all contributors last year about adding the Apache 2.0 license. There were no objections. **So, we are clear to add the license file to the repo.**

Under extremely conservative circumstances, we would need an explicit statement from each contributor. However, the P4 membership agreement clearly states that work contributed is intended to be open sourced with a permissive license:

> WHERAS, the Consortium intends to make P4 available to the public through a permissive open source license such as Apache 2.0 or BSD other substantially equivalent license;

It also included the Apache CLA, which provides that contributions are effectively licensed under the terms of the Apache 2.0 license.

Therefore, we've decided that a notification about the new license with the opportunity to object would be sufficient.

---

Regarding the question on which files to license, I think that the easiest thing to do here is to just license everything under Apache 2.0. It's a well understood license that is widely used in the P4 (and more broadly) ONF ecosystem. It covers copyright and patents. It also avoid the question of how to use code snippets that are contained in the spec docs. Technically, there may be a better license for the spec, but Apache 2.0 seems good enough.

bocon13 (approved):

jnfoster:
Thanks @bocon13 and @jfingerh!

PR #925, opened 2021-03-17
PSA: add packet length standard metadata

jafingerhut:


jafingerhut:
I do not consider this done, as there is a "TBD" in this PR as I write this comment for fleshing out more details of the value that the new packet_length field will contain for cloned packets.  I think the part that is written so far is useful as a starting point for discussion with those interested in the PSA specification and this field.

jafingerhut:
I am recommending that we consider PR #927 over this one, but I will leave them both open for now.

p4-16/psa/PSA.mdk
```diff
@@ -1039,6 +1041,32 @@ packet contents and metadata when a packet begins egress processing.
...
 
+For normal unicast and multicast packets, `packet_length` is the
+length in bytes of the packet as received by the ingress parser,
+_ignoring_ any changes in the packet length that might have been made
+during ingress processing and ingress deparsing.
+
+If the input port on which the packet arrived was an Ethernet port,
+this includes the 4 bytes of Ethernet Frame Check Sequence, and
+targets may restrict this value to be the minimum Ethernet frame
+length of 64 bytes.
```
blp, line 1052: I don't really like this, because virtual Ethernet packets and virtual packet processing exist and there's no FCS in that case.

I'd prefer something like:
```
If the input port on which the packet arrived was an Ethernet port,
this includes the 4 bytes of Ethernet Frame Check Sequence if and
only if the FCS is included in the bytes available to the parser.
Targets may restrict this value to be the miniimum Ethernet frame
length of 64 bytes (68 bytes including FCS).
```

blp:
Conversation in the LDWG meeting leaned toward this not being portable and therefore against including it in the spec.

p4-16/psa/PSA.mdk
jafingerhut: IMO, this kind of detail is another reason to prefer the PR #927 over this one -- getting multiple implementations to agree on this level of detail is not easy.

FYI, I am pretty sure that the minimum Ethernet frame size is 60 bytes of (Ethernet header + Ethernet payload) + 4 bytes of FCS, totaling 64 bytes (not counting those fun things like preamble and inter-frame gaps that appear on physical Ethernet cables, at least, and whatever WiFi does, which I've never learned in detail).

jafingerhut:
Merged in PR https://github.com/p4lang/p4-spec/pull/927 instead of this one.  Closing this PR.

PR #928, opened 2021-03-25
Possible updates/corrections to P4_16 spec grammar

jafingerhut:
I found a few differences between the latest p4lang/p4c P4_16 grammar, as compared to the grammar.mdk file in the P4_16 spec repository, when reviewing another issue on the spec recently.

I suspect that some of these differences might be due to experimental features in p4c, but some of them look like things we might want to update in the spec's grammar.mdk file.  Happy to get comments about which differences are in which category, and update this PR.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2636,22 +2636,14 @@ These declarations describe two objects:
 ### Extern types { #sec_extern }
 
 []{tex-cmd: "\indent"}
-P4 supports extern object declarations, extern function declarations, and forward extern declarations using the following syntax.
+P4 supports extern object declarations and extern function declarations using the following syntax.
 
 ~ Begin P4Grammar
 externDeclaration
     : optAnnotations EXTERN nonTypeName optTypeParameters '{' methodPrototypes '}'
     | optAnnotations EXTERN functionPrototype ';'
-    | optAnnotations EXTERN name ';'
     ;
 ~ End P4Grammar
-
-The last case is a forward extern declaration, intended for declaring
-the name of an extern object type.  Such a name can be used in later
-declarations, without first having to provide the full extern object
-declaration.  The full extern object declaration must appear later in
-the program.
-
 
 #### Extern functions
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2636,7 +2636,7 @@ These declarations describe two objects:
 ### Extern types { #sec_extern }
 
 []{tex-cmd: "\indent"}
-P4 supports extern object declarations and extern function declarations using the following syntax.
+P4 supports extern object declarations, extern function declarations, and forward extern declarations using the following syntax.
 
 ~ Begin P4Grammar
 externDeclaration
@@ -2645,6 +2645,13 @@ externDeclaration
     | optAnnotations EXTERN name ';'
     ;
 ~ End P4Grammar
+
+The last case is a forward extern declaration, intended for declaring
+the name of an extern object type.  Such a name can be used in later
+declarations, without first having to provide the full extern object
+declaration.  The full extern object declaration must appear later in
+the program.
+
 
 #### Extern functions
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2173,7 +2173,7 @@ An enumeration type is defined using the following syntax:
 ~ Begin P4Grammar
 enumDeclaration
     : optAnnotations ENUM name '{' identifierList '}'
-    | optAnnotations ENUM BIT '<' INTEGER '>' name '{' specifiedIdentifierList '}'
+    | optAnnotations ENUM typeRef name '{' specifiedIdentifierList '}'
     ;
 
 identifierList
@@ -2642,6 +2642,7 @@ P4 supports extern object declarations and extern function declarations using th
 externDeclaration
     : optAnnotations EXTERN nonTypeName optTypeParameters '{' methodPrototypes '}'
     | optAnnotations EXTERN functionPrototype ';'
+    | optAnnotations EXTERN name ';'
     ;
 ~ End P4Grammar
 
@@ -3057,6 +3058,7 @@ typeArg
     : DONTCARE
     | typeRef
     | nonTypeName
+    | VOID
     ;
 
 typeArgumentList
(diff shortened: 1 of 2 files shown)


p4-16/spec/grammar.mdk
```diff
@@ -253,6 +254,7 @@ functionPrototype
 methodPrototype
     : optAnnotations functionPrototype ';'
     | optAnnotations TYPE_IDENTIFIER '(' parameterList ')' ';'
+    | optAnnotations ABSTRACT functionPrototype ';'
```
mbudiu-vmw, line 257: this is clearly experimental, part of the abstract methods.

p4-16/spec/grammar.mdk
```diff
@@ -712,6 +720,7 @@ prefixedNonTypeName
 
 lvalue
     : prefixedNonTypeName
+    | THIS
```
mbudiu-vmw, line 723: part of abstract methods.

p4-16/spec/grammar.mdk
```diff
@@ -328,12 +331,15 @@ typeParameterList
...
     : DONTCARE
     | typeRef
+    | VOID
     ;
 
 typeArg
     : DONTCARE
     | typeRef
     | nonTypeName
+        // This is necessary because template arguments may introduce the return type
```
mbudiu-vmw, line 341: I believe that the comment applies to the previous line

mbudiu-vmw (approved):

p4-16/spec/grammar.mdk
jafingerhut: removed

p4-16/spec/grammar.mdk
jafingerhut: removed

p4-16/spec/grammar.mdk
jafingerhut: I just removed the comment.  I only left it in in case it helped explain the reason for the VOID case of typeArg on the next line.

jafingerhut:
@mbudiu-vmw So you believe the other occurrences of THIS added by this PR are not related to abstract methods, and should be included in spec grammar?

mbudiu-vmw:
this is only used with abstract methods.

jafingerhut:
@mbudiu-vmw The annotationToken had one production in grammar.mdk `annotationToken -> THIS` which I have removed in my latest commit, along with the two occurrences of `THIS` that earlier commits of this PR added.  Not sure if it should be in the `annotationToken` rule or not, in the spec.

mbudiu-vmw:
@mbudiu-vmw to check the "voids".

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2642,6 +2642,7 @@ P4 supports extern object declarations and extern function declarations using th
 externDeclaration
     : optAnnotations EXTERN nonTypeName optTypeParameters '{' methodPrototypes '}'
     | optAnnotations EXTERN functionPrototype ';'
+    | optAnnotations EXTERN name ';'
```
jafingerhut, line 2645: @jafingerhut Action item for me to propose adding a sentence or two to the spec for this.  The basic idea is that it enables giving a kind of forward declaration for an extern name, before it is fully defined.  Chris Dodd believes the motivation for this being in the grammar was for Register and RegisterAction definition in TNA.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3057,6 +3058,7 @@ typeArg
     : DONTCARE
     | typeRef
     | nonTypeName
+    | VOID
```
jafingerhut, line 3061: @mbudiu-vmw Mihai took the action item in LDWG meeting to think about why VOID is here, and also the occurrence in the production of the `realTypeArg` nonterminal of the grammar.

p4-16/spec/P4-16-spec.mdk
jafingerhut: I have added some proposed text for this, which should be reviewed.

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: Perhaps this may make sense for some extern objects that could be instantiated with "void" for some type parameters.

mbudiu-vmw (approved):

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2636,15 +2636,23 @@ These declarations describe two objects:
...
 
 ~ Begin P4Grammar
 externDeclaration
     : optAnnotations EXTERN nonTypeName optTypeParameters '{' methodPrototypes '}'
     | optAnnotations EXTERN functionPrototype ';'
+    | optAnnotations EXTERN name ';'
     ;
 ~ End P4Grammar
 
+The last case is a forward extern declaration, intended for declaring
```
mbudiu-vmw, line 2649: I guess this allows mutually-recursive declarations (instantiations). Otherwise this would be impossible.

jafingerhut: I do not know of an example, other than Chris Dodd's mention during the LDWG meeting that perhaps this was added because of the way that the TNA architecture defines the Register and RegisterAction externs.  He did not sound confident that this was the motivating reason, so it might be some other reason.  I do not see any occurrences of forward declarations of externs in this public file, for example: https://github.com/barefootnetworks/Open-Tofino/blob/master/share/p4c/p4include/tofino.p4

p4-16/spec/P4-16-spec.mdk
jafingerhut: During 2021-May-03 LDWG meeting it was decided to leave forward extern declarations out of the spec.  It will remain in the p4c implementation, which already includes several features marked experimental that are not in the spec.

AI Andy: Remove this part of this PR and update it today.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Following discussion at the 2021-May-03 LDWG meeting, I have removed all changes related to forward extern declarations from this PR.

jafingerhut:
@mbudiu-vmw I do not recall whether it was decided to merge this PR into the spec today, but if it was, note that I have removed all of the changes related to the forward declarations of externs from this PR, as suggested during today's LDWG meeting.

mbudiu-vmw (approved):
Yes, we have agreed to merge.

PR #945, opened 2021-06-22
Allow switch statements in actions

mbudiu-vmw:
Since we made switch statements more general we should allow them in actions.
By forbidding table applications we are more precise.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6169,9 +6169,10 @@ table entries (e.g., as specified by the control plane, the
 property).
 
 The body of an action consists of a sequence of statements and
-declarations. No `switch` statements are allowed within an
-action---the grammar permits them, but a semantic check should reject
-them. Some targets may impose additional restrictions on action
+declarations.  No `table`, `control`, or `parser` applications can
+appear within actions.
+
+Some targets may impose additional restrictions on action
 bodies---e.g., only allowing straight-line code, with no conditional
 statements or expressions.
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
The current implementation of switch statements in open source p4c that transforms them into ones where the switch expression include `table_name.apply().action_run` won't work for switch statements inside of actions, unless we also allow table applications in actions, which seems more controversial when I asked about it in the LDWG late in 2021.

Are you thinking that within actions, switch statements would be implemented without using a table apply() to do it?  It is certainly possible, of course.  The current implementation of switch statements is only one way to do it.

mbudiu-vmw:
The spec does not mandate how switch statements are implemented. Yes, the ones in actions would need to use a different implementation strategy - e.g., nested if/else statements.

mbudiu-vmw:
Unless, of course, they are supported natively by the target. Which is true for some targets like ebpf.

jafingerhut (approved):
This removing of a restriction seems reasonable to me.  As is already mentioned in the spec, some targets may not allow conditional statements inside of actions.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6169,9 +6169,10 @@ table entries (e.g., as specified by the control plane, the
 property).
 
 The body of an action consists of a sequence of statements and
-declarations. No `switch` statements are allowed within an
-action---the grammar permits them, but a semantic check should reject
-them. Some targets may impose additional restrictions on action
+declarations.  No `table`, `control`, or `parser` applications can
+appear within actions.
```
jnfoster, line 6173: This is related to #852!

PR #949, opened 2021-07-03
Make psa.p4 include file in spec repo explicitly not compilable

jafingerhut:
Before this change,the p4-16/psa/psa.p4 include file contained a mix
of one section intended to be an example for a BMv2 target-specific
psa.p4 include file, and also a section with 'unspecified' for several
bit widths intended for generating the PSA specification text.

With this change, that file is only for the purpose of generating the
PSA specification.  It is no longer compilable, and contains an #error
directive that should make this obvious to anyone who tries to include
it while compiling a P4 program.  It also has comments linking to two
compilable example target-specific psa.p4 include files in the
p4lang/p4c repository.

Add a copy of the compilable version of the BMv2-specific psa.p4
include file in p4-16/psa/examples/psa-for-bmv2.p4, simply so the
example PSA programs can include that version.  Keeping it separate in
this repository makes it straightforward to do experimental local
changes in this repo.

Remove include of core.p4 from psa.p4.  This should be unnecessary
given that all example programs for other architectures recommend
first including core.p4, then the architecture's header file.

Add @pure and @noSideEffect annotations to match the ones used by
psa.p4 include files in open source p4c compiler.

### DIFF
--- a/p4-16/psa/examples/psa-example-bridged-metadata.p4
+++ b/p4-16/psa/examples/psa-example-bridged-metadata.p4
@@ -15,7 +15,14 @@ limitations under the License.
 */
 
 #include <core.p4>
-#include "../psa.p4"
+/* In a normal PSA program the next line would be:
+
+#include <psa.p4>
+
+ * These examples use psa-for-bmv2.p4 instead so that it is convenient
+ * to test compiling these PSA example programs with local changes to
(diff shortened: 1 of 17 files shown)


gbrebner:
Okay with me.

hanw (approved):

PR #967, opened 2021-10-04
Extend minSizeInBits to cover more cases

mbudiu-vmw:
Fixes https://github.com/p4lang/p4-spec/issues/937

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4080,16 +4080,6 @@ of the header.
 - The method `setInvalid()` sets the header's validity bit to
 "false". It can only be applied to an l-value.
 
-The expression `h.minSizeInBits()` is defined for any value `h` that has a
-header type.  The expression is equal to the sum of the sizes of all
-of header `h`'s fields in bits, counting all `varbit` fields as length
-0.  An expression `h.minSizeInBits()` is a compile-time constant with type
-`int`.  See also Section [#sec-minsizeinbits].
-
-The expression `h.minSizeInBytes()` is similar to `h.minSizeInBits()`, except
-that it returns the total size of all of the header's fields in bytes.
-See also Section [#sec-minsizeinbits].
-
 Similar to a `struct`, a header object can be initialized with a list
 expression [#sec-list-exprs] --- the list fields are assigned to the
 header fields in the order they appear --- or with a structure initializer
@@ -4174,14 +4164,6 @@ expressions are legal:
   the stacks to have the same types and sizes. All components of `hs`
   are copied, including its elements and their validity bits,
   as well as `nextIndex`.
-
-- The expression `hs.minSizeInBits()` returns the stack size
-  multiplied by the `minSizeInBits` of an element.  See also the Section
-  [#sec-minsizeinbits].
-
-- The expression `hs.minSizeInBytes()` returns the stack size
-  multiplied by the `minSizeInBytes` of an element.  See also the
-  Section [#sec-minsizeinbits].
 
 To help programmers write parsers for header stacks, P4 also offers
 computations that automatically advance through the stack as elements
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -4762,6 +4768,36 @@ H h2 = { f2=5, ... };   // initialize h2 with a header that is valid, field f1 0
...
+to most expressions.  Each of these method calls evaluate to
+compile-time known values that return the minimum size in bits
+required to store the expression.
+
+The method `minSizeInBytes` always returns the result the result of
+`minSizeInBits` rounded up to the next whole number of bytes.  not a
+multiple of 8 bits long.  In other words, for any expression `e`,
+`e.minSizeInBytes()` is equal to `(e.minSizeInBits() + 7) >> 3`.
+
+The definition of `e.minSizeInBits` is given recursively on the type of `e` as
```
jnfoster, line 4783: Should this be `e.minSizeInBits()`?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4762,6 +4768,36 @@ H h2 = { f2=5, ... };   // initialize h2 with a header that is valid, field f1 0
...
+follows:
+
+* if the type of `e` is `varbits` the result is 0
+* if the type of `e` is `bit<N>` or `int<N>` the result is `N`
+* if the type of `e` is a `struct` or `header`, the result is the sum
+  of the `minSizeInBits` of all fields of the struct or header.
+* if the type of `e` is a `header_union` the result is the max of the
```
jnfoster, line 4790: Do we want the max/min semantics here?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4762,6 +4768,36 @@ H h2 = { f2=5, ... };   // initialize h2 with a header that is valid, field f1 0
...
+
+The method `minSizeInBytes` always returns the result the result of
+`minSizeInBits` rounded up to the next whole number of bytes.  not a
+multiple of 8 bits long.  In other words, for any expression `e`,
+`e.minSizeInBytes()` is equal to `(e.minSizeInBits() + 7) >> 3`.
+
+The definition of `e.minSizeInBits` is given recursively on the type of `e` as
+follows:
+
+* if the type of `e` is `varbits` the result is 0
```
jnfoster, line 4786: Do we want this semantics? And is it consistent with what we do for header unions and stacks below?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4762,6 +4768,36 @@ H h2 = { f2=5, ... };   // initialize h2 with a header that is valid, field f1 0
...
+* if the type of `e` is `bit<N>` or `int<N>` the result is `N`
+* if the type of `e` is a `struct` or `header`, the result is the sum
+  of the `minSizeInBits` of all fields of the struct or header.
+* if the type of `e` is a `header_union` the result is the max of the
+  `minSizeInBits` of all fields of the union
+* if the type of `e` is a header stack, the result is the size of the
+  stack multiplied with the `minSizeInBits` of an element
+* if the type of `e` is a serializable `enum`, the result is the
+  `minSizeInBits` of the underlying type
+* if the type of `e` is a type introduced by `newtype`, the result
```
jnfoster, line 4796: Do we want to enforce stricter information hiding for `newtype`?

jnfoster:
We discussed this issue at the October '21 LDWG. 

There is broad support for this proposal, but some questions about certain types (header stacks, unions, etc.). 

We resolved to remove this operation for those types.

Action items:
- @vgurevich should weigh in on the proposal
- @mbudiu-vmw will modify the proposal to implement the above changes

p4-16/spec/P4-16-spec.mdk
vgurevich: I can see the consistency with the definition of `minSizeInBits()` for the headers above (where we state that varbits are counted as 0).

However, for the practical purposes, it would be a lot more useful to have at least **some way** to return `N` for a `varbit<N>`.

One way to do that, might be to state that `e.minSizeInBits()` is really the minimum number of bits that should be allocated to store ANY value of the corresponding type. This, by the way, is consistent with what we do with header stacks that can also be only partially filled. This also works well with the idea that we allocate all resources statically and as a result there are no exceptions in the controls.

p4-16/spec/P4-16-spec.mdk
vgurevich: @jnfoster -- I do not think we need to hide the size of the `newtype`, becaue it still ahs to be visible via table keys, action parameters, etc. The only thing we really want to do (in that specific case where the type is used for port (re)mapping is to constrain some operations that might be dangerous (e.g. ternary matching, math, etc.). Note by the way that today the constraints are so severe that this construct is almost useless.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4177,6 +4175,14 @@ expressions are legal:
   are copied, including its elements and their validity bits,
   as well as `nextIndex`.
 
+- The expression `hs.minSizeInBits()` returns the stack size
```
mbudiu-vmw, line 4178: Proposal is for this to be 0.

jnfoster (approved):
Looks good. I like the way you converted the Google Doc notation.

Minor: maybe we should split out controls/parsers/packages?

mbudiu-vmw:
Was actually thinking to take out all the undefined ones and say that everything is is undefined...

jnfoster:
That could work too. Probably more robust that way.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -8081,13 +8123,18 @@ The P4 compiler should provide:
...
+|           |              | enumeration types, abstract methods, conditional and empty         |
+|           |              | statements in parsers, additional casts, and 0-width types;        |
 |           |              | clarified semantics of default actions, headers, empty types, and  |
-|           |              | action data; fixed typos and inconsistencies in the grammar.       | 
+|           |              | action data; fixed typos and inconsistencies in the grammar.       |
 |-----|-----|-----|
 
+## Summary of changes made in version 1.2.3
+
+* extended the of minSizeInBits, and minSizeInBytes to apply to apply to more
```
jafingerhut, line 8135: looks like some kind of typo here.  Perhaps remove "the of"?

jnfoster:
Any the double “to apply”

jfingerh:
Would it be reasonable to include at least a link to another file inside this same repo that gives the rationale that Nate recorded in his notes, for the cases like header stacks and header unions that one might consider "odd"?  Also the example Nate wrote that seems counter-intuitive when written out and demonstrated how minSizeInBits and maxSizeInBits behave on those data types?  I understand if we don't want to include such background material in the spec itself, but having a separate MarkDown file in the directory `p4-16/spec/docs` has been done for similar background details for other things in the spec, and seems useful as history of "why".

jnfoster:
I personally think that would be fine...

-N

jafingerhut:
I am happy adding the separate document with background details/rationale in a separate PR after this one is merged, if that helps this one make progress.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4762,6 +4750,59 @@ H h2 = { f2=5, ... };   // initialize h2 with a header that is valid, field f1 0
...
+# Compile-time size determination { #sec-minsizeinbits }
+
+The method calls `minSizeInBits`, `minSizeInBytes`, `maxSizeInBits`,
+and `maxSizeInBytes` can be applied to some expressions.  Each of
+these method calls evaluate to compile-time known values that return
+the minimum size in bits required to store the expression.  None of these
+methods evaluates the expression itself.  In consequence, the expression
+may be invalid (e.g., an out-of-bounds header stack access).
+
+The method `minSizeInBytes` always returns the result the result of
```
jafingerhut, line 4762: typo: duplicated "the result the result"

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4762,6 +4750,59 @@ H h2 = { f2=5, ... };   // initialize h2 with a header that is valid, field f1 0
...
+`minSizeInBits` rounded up to the next whole number of bytes.
+In other words, for any expression `e`,
+`e.minSizeInBytes()` is equal to `(e.minSizeInBits() + 7) >> 3`.
+
+The method `maxSizeInBytes` always returns the result the result of
```
jafingerhut, line 4767: Same typo "the result the result" here.

jafingerhut (approved):
Looks good to me, with only minor comments about a couple of typos.  I am fine if we wait for a later PR to add a separate file with background details/rationale.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -8081,13 +8123,18 @@ The P4 compiler should provide:
...
+|           |              | enumeration types, abstract methods, conditional and empty         |
+|           |              | statements in parsers, additional casts, and 0-width types;        |
 |           |              | clarified semantics of default actions, headers, empty types, and  |
-|           |              | action data; fixed typos and inconsistencies in the grammar.       | 
+|           |              | action data; fixed typos and inconsistencies in the grammar.       |
 |-----|-----|-----|
 
+## Summary of changes made in version 1.2.3
+
+* extended minSizeInBits, and minSizeInBytes to apply to more
```
jnfoster, line 8135: Capitalize the "e" here.

jnfoster:
We agreed to take this as written (modulo typos) at December '21 meeting.

PR #968, opened 2021-10-04
Clarify types allowed in select expressions

### LINKED ISSUE #954: What types are allowed in select expressions?
The spec is silent in this respect. Can you use a struct, or a varbit?
The implementation complains about expressions containing headers only.

mbudiu-vmw:
Fixes #954

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -5510,7 +5510,8 @@ selectCase
 ~ End P4Grammar
 
 In a `select` expression, if the `expressionList` has type `tuple<T>`,
-then each `keysetExpression` must have type `set<tuple<T>>`.
+then each `keysetExpression` must have type `set<tuple<T>>`.  Each
+expression in the `expressionList` must have a `bit<W>` or `int<W>` type.
 
 In terms of the `ParserModel`, the meaning of a select expression:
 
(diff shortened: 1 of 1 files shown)


jnfoster:
We discussed this at the October '21 LDWG.

Status:
* It's clear we should support `bit<n>` and `int<n>`, and serializable enums.
* It seems possible that we should also support `bool` and nested tuples.
* It's unclear whether we should also support headers, `struct`s, `varbit`s, etc.

We should be careful about introducing a non-backward compatible change that could break programs and backends.

Action items:
* @mbudiu-vmw will change the compiler implementation to enforce restrictions to numeric types
* @jafingerhut will check on whether this will break any programs or backends at Intel, including folks like @robertsoule-barefoot who may have novel in-network computing applications.
* Others should do the same...

jnfoster:
@jafingerhut and @jnfoster will investigate and propose a list of types allowed, consistent with extant implementations, P4Runtime, and as general as possible.

mbudiu-vmw:
Change the changelog

jafingerhut:
@ChrisDodd @hanw @vgurevich Can you think of any issues with the proposed restrictions in select expressions?  I am not thinking of any.

vgurevich:
In general I agree. I do not see why we can't support structs and headers, though. Structs are especially useful for the `value_set`s since that's the only way the API can generate reasonable names there.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5803,10 +5803,10 @@ selectCase
 ~ End P4Grammar
 
 In a `select` expression, if the `expressionList` has type `tuple<T>`,
-then each `keysetExpression` must have type `set<tuple<T>>`.  In
-particular, if a set is specified as a range or mask expression, the
-endpoints of the range and mask expression are implicitly cast to type
-`T` using the standard rules for casts.
+then each `keysetExpression` must have type `set<tuple<T>>`.  Each
```
mbudiu-vmw, line 5806: we should not delete the text on the left.

mbudiu-vmw:
@jafingerhut I have reinstanted the paragraph deleted by mistake

jafingerhut (approved):
LGTM

PR #970, opened 2021-10-07
Minor updates for Issue 955

### LINKED ISSUE #955: Index type of bit slicing
The data types allowed for the bit slicing index are unclear in sections 8.5 and 8.6, and it may be appreciated to be more explicit than saying “positive integers that are compile-time known values”. It would be nice to explicitly list the allowed types, similar to how the index of header stacks is defined in 8.17 (where int, bit<W>, int<W>, and serializable enum are allowed). For example, are bit<w> values permitted, or does the index have to be an int?

MollyDream:
- We clarified the types of indices for bit slicing in two places, for `bit<w>` and `int<w>`.
- We removed the requirements that indices be positive (it seems that zero should also be allowed.)

Fixes #955.


MollyDream:
Accidentally removed the commit. Sorry about that!

jfingerh (approved):
Looks reasonable to me.  My only minor comment, which need not result in any changes to this PR, is that if we could list the allowed types once, and refer to them from the other part of the spec, that is some factoring that if we ever amend the list of types, it need only be updated in one place rather than two.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3452,9 +3452,20 @@ to bit-strings of the same width:
 Bit-strings also support the following operations:
 
 - Extraction of a set of contiguous bits, also known as a slice, denoted by `[m:l]`,
-  where `m` and `l` must be positive integers
-  that are compile-time known values, and `m >= l`. The result is
-  a bit-string of width `m - l + 1`, including the bits numbered
+  where `m` and `l` must be expressions that evaluate to
```
jnfoster, line 3455: We should say that `m` and `l` are non-negative.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3531,9 +3542,20 @@ type. The result always has the same width as the left operand.
   - all result bits are zero when shifting right a positive value
   - all result bits are one when shifting right a negative value
 - Extraction of a set of contiguous bits, also known as a slice, denoted by `[m:l]`,
-  where `m` and `l` must be positive integers
-  that are compile-time known values, and `m >= l`. The result is
-  an unsigned bit-string of width `m - l + 1`, including the bits numbered
+  where `m` and `l` must be expressions that evaluate to
```
jnfoster, line 3545: Ditto...

jnfoster:
We agreed to take this at the December '21 LDWG meeting, modulo small typos.

jafingerhut (approved):
Approving with my jafingerhut Github account, which was requested.  Earlier I approved with a different Github account of mine instead -- apologies for any confusion or delay.

MollyDream:
@jnfoster I added "non-negative" with the recent commit. Ready for merging.

jnfoster (approved):

PR #980, opened 2021-12-01
Add concatenation for Issue 956

### LINKED ISSUE #956: Clarification of concatenation
Concatenation, as an allowed operation on the bit type, should be at least mentioned in the list of binary operations on the bit type in section [8.5](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-bit-ops). However, the spec instead describes it only in [a subsection of 8.6](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-concatenation), the operations on the fixed-width int type. This would be hard to read.
Also, concatenation should be included as another exception (in addition to shifts) to the requirement that the operands of all binary operations should have the same signedness and width in [8.5](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-bit-ops) and [8.6](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-int-ops) (the paragraph starting with “All binary operations...”.)

MollyDream:
- We clarified that both shifts and concatenation are the exceptions among the binary operations that require two operands to have the same type.
- We added concatenation to the list of allowed binary operations for bit-strings and fixed-width integers.
- We grouped shift, bit slicing, and concatenation in 8.6 (in a similar style to 8.5) to leave them outside the restriction of two operands must have the same types; we swapped shift and bit slicing in 8.5 since shift is the only item among the three that has a different description in 8.5 and 8.6.
- We added bitwise operations explicitly to the list of operations in 8.6 to make it more readable.
- We changed the title from "Concatenation" to "A note about concatenation" for 8.6.1 to make it parallel to 8.6.2.

Fixes  #956

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3408,11 +3408,11 @@ is `0` and the maximum value is `2^W-1`.
 The precedence of saturating addition and subtraction operations is the
 same as for modulo arithmetic addition and subtraction.
 
-All binary operations (except shifts) require both operands to have
+All binary operations (except shifts and concatenation) require both operands to have
 the same exact type and width; supplying operands with different
 widths produces an error at compile time. No implicit casts are
 inserted by the compiler to equalize the widths. There are no binary
-operations that combine signed and unsigned values (except shifts).
+operations that combine signed and unsigned values (except shifts and concatenation).
 The following operations are provided on bit-string expressions:
 
 - Test for equality between bit-strings of the same width, designated
@@ -3471,6 +3471,10 @@ Bit-strings also support the following operations:
   The result has the
   same type as the left operand. Shifting by an amount greater than
   the width of the input produces a result where all bits are zero.
+- Concatention of bit-strings and fixed-width signed integers, denoted by `++`.
+  The two operands must be either `bit<W>` or `int<W>`, and they can be of different signedness and width.
+  The result has the same signedness as the left operand and the width equal to the sum of the two operands' width. 
+  In concatenation, the left operand is placed as the most significant bits.
 
 ## Operations on fixed-width signed integers { #sec-int-ops }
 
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
This is great, we'll discuss this in the next language design meeting and merge it if the committee approves.
You are welcome to join the meeting if you are interested.

MollyDream:
@mbudiu-vmw I would be happy to join the meeting. When is that going to happen? Meanwhile, I will make more pull requests for the remaining issues I opened.

mbudiu-vmw:
The meetings happen over zoom every first Monday of each month at 1PM PT (unless there is some holiday).
The calendar from p4.org has a link: https://p4.org/working-groups/

MollyDream:
@mbudiu-vmw So the next meeting happens next Monday, right? Will attend :)

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3408,11 +3408,11 @@ is `0` and the maximum value is `2^W-1`.
 The precedence of saturating addition and subtraction operations is the
 same as for modulo arithmetic addition and subtraction.
 
-All binary operations (except shifts) require both operands to have
+All binary operations (except shifts and concatenation) require both operands to have
 the same exact type and width; supplying operands with different
 widths produces an error at compile time. No implicit casts are
 inserted by the compiler to equalize the widths. There are no binary
-operations that combine signed and unsigned values (except shifts).
+operations that combine signed and unsigned values (except shifts and concatenation).
```
jfingerh, line 3415: Do concatenations support combining signed and unsigned values, without casts?

jfingerh (approved):
If the consensus (and p4c implementation) support concatenation of one signed and one unsigned value, then I approve of these changes as written.

p4-16/spec/P4-16-spec.mdk
MollyDream: Yes, that is inferred in the original spec:
_8.6.1. Concatenation
Concatenation is applied to two bit-strings (signed or unsigned)... the sign of the result is taken from the left operand._

jnfoster:
We agreed to take this fix at the December '21 LDWG meeting.

PR #982, opened 2021-12-01
Clarify implicit casts of enum for Issue 958

### LINKED ISSUE #958: When can a serializable enum be implicitly cast?
The specification states that "implicit casting from [a serializable] enum to its fixed-width (signed or unsigned) integer type" is supported in [8.3](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-enum-exprs), but it is unclear when exactly that is doable. The spec shows that enum can be used as the left argument of bit shifts and the index of header stack access, so what about bit shift’s right argument, concatenation, unary operations, and bit slicing’s bitstring and indices? 
After this is made transparent with either explanation or case-by-case rules, it should be probably re-emphasized or referenced in section [8.9.2](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-implicit-casts) of implicit casts.

Example:
```
enum bit<8> E {
  e1 = 1,
  e2 = 2
}
E e = E.e1;
bits<8> x = e << 3;    // Allowed in the spec
bits<8> y = 8w0 << e;  // Unclear
bits<16> z = e ++ 8w0; // Unclear
bit<8> w = -e;         // Unclear
bit<4> u = e[3:0];     // Unclear
bit<4> v = 8w0[4:e];   // Unclear
```
Based on the spec, the expression in the initialization of `x` is allowed, but whether the expressions on e in the following initializations are legal is unclear.

MollyDream:
- We added a reference to the implicit cast section in the enum operations section.
- We re-asserted that the implicit casts from enum to its underlying types is always safe and allowed, with extra examples.
Fixes #958

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3195,7 +3195,7 @@ fixed-width unsigned integer representation for `E.e2`, `1`.  The variable `b` i
 symbolic value `E.e2`, which corresponds to the fixed-width unsigned integer value `1`.
 
 Because it is always safe to cast from an `enum` to its underlying fixed-width integer type,
-implicit casting from an `enum` to its fixed-width (signed or unsigned) integer type is also supported:
+implicit casting from an `enum` to its fixed-width (signed or unsigned) integer type is also supported (see Section [#sec-implicit-casts]):
 
 ~ Begin P4Example
 bit<8> x = E.e2; // sets x to 1 (E.e2 is automatically casted to bit<8>)
@@ -3710,7 +3710,9 @@ only implicitly casts from `int` to fixed-width types and from enums
 with an underlying type to the underlying type. In particular,
 applying a binary operation to an expression of type `int` and an
 expression with a fixed-width type will implicitly cast the `int`
-expression to the type of the other expression.
+expression to the type of the other expression. For enums with an underlying
+type, it can be implicitly cast to its underlying type whenever appropriate,
+including but not limited to in shift, concatenation, bit slicing, header stack indexing and other binary operations.
 
 For example, given the following declarations,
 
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):
Looks good to me.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3731,6 +3733,10 @@ the compiler will add implicit casts as follows:
 - `x << 13` becomes `0`; overflow warning
 - `x | 0xFFF` becomes `x | (bit<8>)0xFFF`; overflow warning
 - `x + E.a` becomes `x + (bit<8>)E.a`
+- `16w11 << E.a` becomes `16w11 << (bit<8>)E.a`
+- `E.a[3:0]` becomes `((bit<8>)E.a)[3:0]`
```
jnfoster, line 3737: We should remove this line, and explain it in a separate PR addressing slices, l-values, and serializable enums.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3710,7 +3710,9 @@ only implicitly casts from `int` to fixed-width types and from enums
 with an underlying type to the underlying type. In particular,
 applying a binary operation to an expression of type `int` and an
 expression with a fixed-width type will implicitly cast the `int`
-expression to the type of the other expression.
+expression to the type of the other expression. For enums with an underlying
+type, it can be implicitly cast to its underlying type whenever appropriate,
+including but not limited to in shift, concatenation, bit slicing, header stack indexing and other binary operations.
```
jnfoster, line 3715: Rewrite to remove implication that casts are inserted when a serializable enum is used as the base of a bit slice here.

jnfoster:
We agreed to take this at the December '21 LDWG, modulo two small fixes.

MollyDream:
@jnfoster I restricted the implicit casts of enums in bit slicing to only the indexes with the recent commit. Ready for final review and merging. Thanks!

jnfoster (approved):

p4-16/spec/P4-16-spec.mdk
jafingerhut: I have created this PR regarding this issue: https://github.com/p4lang/p4-spec/issues/992.  Please take a look at it and add corrections or clarifications to my initial attempt at describing it.

PR #983, opened 2021-12-01
Clarify shifting int by constant bit<w> for Issue 959

### LINKED ISSUE #959: Is integer >> bit<w> legal?
It is unclear from the manual if a bit shift of a variable-width integer by a fixed-width bit is allowed. It is listed as illegal with an example in section [8.9.3](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-illegal-arith) (“ `bit<8> x; 1 << x;` ” is illegal). However, the reasoning for the example being illegal (“width of 1 is unknown”) does not match up with the fact that integers can be shifted by integers ([8.7](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-varint-ops)). On the other hand, if this shift is legal, it will make the allowed operand types of shifts much cleaner, in that the left operand can be the bit\<w> / int\<w> / integer, and the right operand can be the bit\<w> / integer ([8.5](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-bit-ops) and [8.6](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-int-ops)). BMV2 also allows this shift.

Following this reasoning, the description of "each operand that participates in any of these operation must have type int" in [8.7](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-varint-ops) may be legacy content. This also erroneously appears in [8.6](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-int-ops) in a slightly paraphrased way ("all binary operations require both operands to have the exact same type"), contradicting the “All binary operations……” paragraph.

Example:
```
bit<8> a = 1;
int<8> b = 1;
integer c = 1;

bit<8> x = a << a;    // Legal
bit<8> y = a << c;    // Legal
int<8> z = b << a;    // Legal
int<8> w = b << c;    // Legal
integer u = c << a;   // Legal
integer v = c << c;   // Unclear
```
Based on the spec, it is unclear whether the expression in the last initialization is allowed though the first five are legal.

MollyDream:
- We listed explicitly the allowed "positive and constant" right operands for shifts on arbitrary-width integers.
- We moved the subsection of "A note about shifts" to after 8.7 and added a beginning paragraph to describe the general patterns of the allowed operands for shifts.
- We modified the reason why an arithmetic operation is illegal accordingly.

Fixes #959

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3467,7 +3467,7 @@ Bit-strings also support the following operations:
   unchanged.  A slice of an unsigned integer is an unsigned integer.
 - Logical shift left and right with a runtime known unsigned integer
   value, denoted by `<<` and `>>` respectively. In a shift, the left operand is unsigned, and right operand must be either an
-  expression of type `bit<S>` or a non-negative integer literal.
+  expression of type `bit<S>` or a non-negative integer constant.
   The result has the
   same type as the left operand. Shifting by an amount greater than
   the width of the input produces a result where all bits are zero.
@@ -3521,7 +3521,7 @@ type. The result always has the same width as the left operand.
 - Arithmetic shift left and right denoted by `<<` and `>>`.
   The left operand is signed and the right operand must be either an
   unsigned number of type `bit<S>` or a non-negative
-  integer literal. The result has the same type as the left operand.
+  integer constant. The result has the same type as the left operand.
   Shifting left produces the exact same bit pattern as a shift left
   of an unsigned value.  Shift left can thus overflow, when it leads to
   a change of the sign bit.
@@ -3570,7 +3570,7 @@ expressions of type `int` must be compile-time known values.  The type
 - Truncating integer division between positive values, denoted by `/`.
 - Modulo between positive values, denoted by `%`.
 - Arithmetic shift left and right denoted by `<<` and `>>`. These operations produce an `int` result.
-  The right operand must be either an unsigned constant of type `bit<S>` or a non-negative integer literal.
+  The right operand must be either an unsigned constant of type `bit<S>` or a non-negative integer constant.
   The expression `a << b` is equal to $a \times 2^b$ while `a >> b`
   is equal to $\lfloor{a / 2^b}\rfloor$.
 
@@ -3598,7 +3598,7 @@ Note: saturating arithmetic is not supported for arbitrary-precision integers.
 
 The left operand of shifts can be any one out of unsigned bit-strings, signed bit-strings,
 and arbitrary-precision integers, and the right operand of shifts must be either
-an expression of type `bit<S>` or a non-negative integer literal. The result has the
+an expression of type `bit<S>` or a non-negative integer constant. The result has the
 same type as the left operand.
 
 Shifts on signed and unsigned bit-strings deserve a special discussion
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -3604,14 +3570,14 @@ expressions of type `int` must be compile-time known values.  The type
 - Truncating integer division between positive values, denoted by `/`.
 - Modulo between positive values, denoted by `%`.
 - Arithmetic shift left and right denoted by `<<` and `>>`. These operations produce an `int` result.
-  The right operand must be constant and positive.
+  The right operand must be either an unsigned constant of type `bit<S>` or a non-negative integer literal.
```
QinshiWang, line 3573: Can it be a mathematical (unlimited-width) integer constant?

MollyDream: Good point. The shifts of bit<w> and int<w> uses the wording of "a non-negative integer literal", but indeed a integer constant is also "constant and positive".

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3467,7 +3467,7 @@ Bit-strings also support the following operations:
   unchanged.  A slice of an unsigned integer is an unsigned integer.
 - Logical shift left and right with a runtime known unsigned integer
   value, denoted by `<<` and `>>` respectively. In a shift, the left operand is unsigned, and right operand must be either an
-  expression of type `bit<S>` or a non-negative integer literal.
+  expression of type `bit<S>` or a non-negative integer constant.
```
jafingerhut, line 3470: @jnfoster @mbudiu-vmw Is this a case where you would prefer the phrase "compile time known value" instead of "constant"?  I do not have a strong preference, but wanted to see if you did.

mbudiu-vmw: "compile time known value" is very verbose, but I think we consistently used it in the spec... it's perhaps more precise than "constant"

jafingerhut (approved):
I left one question on wording for Nate & Mihai, but other than that, this all looks reasonable to me.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3766,7 +3772,7 @@ cannot guess the user intent, so P4 requires the user to disambiguate.
 | `x << z`            | RHS of shift cannot be signed |  `x << (bit<8>)z ` |
 | `x < z`             | Different signs |  `X < (bit<8>)z ` |
 |                         |                           |  `(int<8>)x < z ` |
-| `1 << x`            | Width of `1` is unknown |  `32w1 << x ` |
+| `1 << x`            | RHS of int shift must be compile-time known |  `32w1 << x ` |
```
jnfoster, line 3775: Suggested fixes:
- Combine the two reasons, as an "either or"
- Write "None" for the alternative

p4-16/spec/P4-16-spec.mdk
jnfoster: I agree with this...

jnfoster:
We agreed to take this change at the December '21 LDWG meeting (modulo small fixes).

MollyDream:
@jnfoster I changed the wording to "compile-time known value" and adapted the illegal arithmetic expression to accept two explanations in the recent commits. Ready for final review and merging.

jnfoster (approved):

PR #984, opened 2021-12-01
Add equality and inequality comparisons for Issue 960

### LINKED ISSUE #953: Are structure-valued expressions subject to implicit casts?
In the new section on structure-valued expressions, the spec says "The typeRef can be omitted if it can be inferred from context, e.g., when initializing a variable with a struct type." Is this an implicit cast or is it a separate kind of inference? If it's an implicit cast, it should be included in the section 8.9.2 on implicit casts.

### LINKED ISSUE #960: Comparisons of list expressions
The specification states that structure-valued expressions can be used in comparisons (== or !=) in section [8.12](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-structure-expressions) ("Structure-valued expressions can be used... in comparisons..."), but it does not say so for list expressions in section [8.11](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-list-exprs). Since those two kinds of expressions are very similar, this inconsistency here seems weird and probably needs some clarification.

Similarly, it probably should be clarified in section [8.10](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-tuple-exprs) whether tuples support comparisons (== or !=).

Lastly, the example of structure-valued expressions used in comparisons in [8.12](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-structure-expressions) is perhaps a bad example ( `bool b = s == (S) { a = 1, b = 2 }` ) since that essentially compares two structures instead of two structure-valued expressions.

Example:
```
struct S {
    bit<4> a;
}
bool x = {a = 1} == {a = 1};
bool y = {1} == {1};
```
Based on the spec the first comparison is allowed, but the second is not.

MollyDream:
- We added explicit equality and inequality comparisons for tuples, list expressions, and structure-valued expressions.
- We explained the recursive implicit casts applied when using list and structure-valued expressions. (Do we want to also mention this in the implicit cast section?)
- We elaborated the examples to illustrate these two points.
- We merged `8.15. Structure initializers` into `8.12. Structure-valued expressions` and `8.14. Operations on struct types` to reduce confusion and redundancy and to make the spec more organized.
- We changed the title of section `8.12. Structure-valued expressions` into `8.12. Operations on structure-valued expressions` to make the spec more organized.
- We renamed all usages of "structure initializer" with "structure-valued expressions" to be parallel to "list expressions" and to reduce confusion.

Fixes #960 
Fixes #953 

Note: There is a corner case not covered but this change. For more details, please see the issue report #960.


mbudiu-vmw:
Can you please delete .DS_Store?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4035,37 +4077,32 @@ field access, written using dot (".") notation---e.g., `s.field`. If
...
+with a structure-valued expression, as described in
+[#sec-structure-expressions].  Both these cases must initialize all
 fields of the structure.
 
-Two structs can be compared for equality (==) or inequality (!=) only
-if they have the same type and all of their fields can be recursively
-compared for equality.  Two structures are equal if and only if all
-their corresponding fields are equal.
-
-## Structure initializers { #sec-structure-initializers }
```
jfingerh, line 4047: To other reviewers: It is intentional that these changes remove the section "Structure initializers".  The intent is that they should be removed from the spec, since they are redundant with the addition of structure-valued expressions.  @MollyDream  please add another comment replying to this one if I have stated things incorrectly here.

MollyDream: That is exactly right Andy. "We merged `8.15. Structure initializers` into `8.12. Structure-valued expressions` and `8.14. Operations on struct types` to reduce confusion and redundancy and to make the spec more organized."

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3778,9 +3778,16 @@ arguments and returned from functions, and can be initialized with
...
 ~ Begin P4Example
-tuple<bit<32>, bool> x = { 10, false };
+tuple<bit<32>, bool> x = { 10, false }; // Initialization
+tuple<bit<32>, bool> y = x;             // Assignment
+bool b = x == y;                        // Equality comparison
 ~ End P4Example
 
+Two tuples can be compared for equality (==) or inequality (!=) only
+if they have the same type and all of their fields can be recursively
+compared for equality.  Two tuples are equal if and only if all
```
jnfoster, line 3788: "all of"?

jnfoster:
Example program from LDWG chat.
```
control C<G>(inout G G);
package P<G>(C<G> c);

header h {
  bit<8> f;
}

header g {
  bool b1;
  bool b2;
  bool b3;
  bit<5> padding;
}

struct s { 
  bit<8> f;
}

control MyC(inout g my_g) {
  h my_h;
  s my_s; 
  tuple<bit<8>> my_t; 
  apply {
     my_g.setValid();
     my_g.b1 = { 8w1 } == my_h;
     my_g.b2 = { 8w2 } == my_s;
     my_g.b3 = { 8w3 } == my_t;
  }
}

P(MyC()) main;
```

jnfoster:
* Let's refactor or split this pull request to eliminate structure initializers.
* Regarding the comparison operations, we need to revise the spec to explain the rewriting that the compiler is doing, converting list expressions into structure-value expressions... this needs some thought because we don't have casts for structs and headers.

MollyDream:
To summarize what we know at the current status, I am copying over the conclusion from [here](https://github.com/p4lang/p4c/issues/3057#issue-1116122059) and [here](https://github.com/p4lang/p4-spec/issues/960#issuecomment-984005608):

> it seems tempting to summarize the allowed comparisons as:
> - A list expression can be compared with a list expression, a structure-valued expression, a tuple, a struct, or a header for equality (==) or inequality (!=)
> - A structure-valued expression can be compared with a list expression, a structure-valued expression, a struct, or a header for quality (==) or inequality (!=)
> - Two tuples, two structs, and two headers can be compared for equality (==) or inequality (!=)
> 

These are _all_ true in the p4c compiler, except one corner case: A list expression can be compared with a structure-valued expression if and only if `'(' typeRef ')'` is explicit.
```
bool b11 = (S) { a = 1, b = 2 } == {1, 2}; // Good
// bool b12 = { a = 1, b = 2 } == {1, 2};  // Rejected by the BMV2 compiler
```
We should first examine if these comparisons allowed by p4c are reasonable.

If so, in order to explain the underlying "conversion" from a list expression to a structure-valued expression (which happens in p4c based on Nate's experiment in the last meeting), new types and casts need to be introduced. As a demonstration, in Petr4's type system, a list expression is of the _list_ type and a structure-valued expression is of the _record_ type. Such two types can only be created and used by the P4 compiler internally (similar to the function types in Section 7.2.8.2.)

Ideally, I would like to claim that casts from the list type to the record type, from the list type to the tuple type, and from the record type to the struct/header type are allowed. All nice and clean in explaning the comparisons and the initializations of tuples and structs/headers. But several changes need to happen in the p4c compiler and the spec:

- Since a structure-valued expression now has a well-defined type, and given that comparing two structure-valued expressions (without `'(' typeRef ')'`)are accepted in p4c, ` { a = 1, b = 2 } == {1, 2}` should also be allowed in p4c.
- In these three new types of casts, the first one (from the list type to the record type) can be only implicit casts since the record type is only internal to the compiler; for the second cast (from the list type to the tuple type), it seems all reasonable to allow explicit casts, e.g. `(tuple<bit<32>, bool) { 10, false }` should be allowed; for the last one (from the record type to the struct/header type), grammar needs to be adapted to recognize `'(' typeRef ')' expression` as a cast.

mbudiu-vmw:
I tried to implement this stuff in the compiler by desugaring these constructs by creating some new types for these expressions, but there are some difficulties. In particular, today struct types cannot have int fields, so the compiler has to infer appropriate widths for these integers. Also, the types for expressions that are compared to each other must be consistent. Not entirely trivial.

jnfoster:
@MollyDream do we close this in favor of @jafingerhut's more fine-grained PRs?

jfingerh:
I only created one other PR which addressed removing structure initializers.  My thinking was that this PR can now be updated to focus solely on any equality/inequality changes that might be desired.  I do not know the details of what changes might be desired for equality/inequality purposes.

jnfoster:
Closing this in favor of @jafingerhut's PRs.

PR #985, opened 2021-12-01
Fix casting wording for Issue 961

### LINKED ISSUE #961: Casting from list expressions & casting to headers
List expressions may be used to initialize structures (implicitly), but it is unclear from the spec whether they can be explicitly cast to structures. The spec should add more examples to [8.11](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-list-exprs) and [8.15](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-structure-initializers) if that is allowed, and add clarifications otherwise. The reason for this confusion is that structure-valued expressions may be used to initialize structures, both implicitly and explicitly ([8.12](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-structure-expressions)).

Example 1:
```
struct S {
    bit<4> a;
}
S x = {a=6};      // Allowed
S y = {6};        // Allowed
S z = (S) {a=6};  // Allowed
S w = (S) {6};    // Unclear
```
Based on the spec, it is vague if the last initialization of `w` is valid though the first three are clearly supported based on the spec.

Similarly, the explicit cast from a list expression / structure-valued expression to a header, whether allowed or not, should probably also be clarified in section [8.16](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-ops-on-hdrs) since a header is “similar to a struct”.

Example 2
```
header H {
    bit<4> a;
}
H x = {a=6};      // Allowed
H y = {6};        // Allowed
H z = (H) {a=6};  // Unclear
H w = (H) {6};    // Unclear
```
Based on the spec, it is vague if the last two initializations of `z` and `w` are valid though the first two are clearly supported based on the spec.

Similarly, the explicit cast from a list expression to a tuple, whether allowed or not, should probably also be clarified in section [8.10](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-tuple-exprs).

Litmus Test 3
```
tuple<bit<32>, bool> x = { 10, false };                         // Allowed
tuple<bit<32>, bool> y = (tuple<bit<32>, bool) { 10, false };   // Unclear
```
Based on the spec, it is vague if the second initialization of the tuple is allowed.

Following that, [8.9.1](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-explicit-casts) should probably list all the allowed casts to tuple, struct, or header from list/struct-valued expressions as legal explicit casts. Also, the beginning of [8.9](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-casts) seems outdated in stating that “casts are only permitted between base types" e.g. even enum is not a base type and "casts between an enum with an explicit type and its underlying type" are clearly allowed.

MollyDream:
- We added the derived types that allow casting.

Fixes #961

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3667,7 +3667,7 @@ finding this information in a section dedicated to type `varbit`.
 
 P4 provides a limited set of casts between types. A cast is written
 `(t) e`, where `t` is a type and `e` is an expression. Casts are only
-permitted between base types. While this design is arguably more onerous
+permitted on base types and derived types introduced by `typedef`, `type`, and `enum`. While this design is arguably more onerous
 for programmers, it has several benefits:
 
 - It makes user intent unambiguous.


jafingerhut (approved):
Looks good to me.

jnfoster (approved):
LGTM.

PR #986, opened 2021-12-01
Add control plane names for Issue 962

### LINKED ISSUE #962: Value set is missing documented in control plane objects
Parser value sets, described in Section 12.11 can be manipulated by the control plane, so they should have control plane names and be documented in Section 17.3. This documentation is missing now. According to @jnfoster, it is probably due to value sets were introduced relatively recently.

MollyDream:
- We added value sets to the control plane names with a new subsection following the existing format.

Fixes #962


mbudiu-vmw (approved):

jafingerhut (approved):
Looks reasonable to me.

PR #987, opened 2021-12-02
Add types with optional parameters for Issue 977

### LINKED ISSUE #977: Optional parameters in parser and control types
The Tofino architecture is inconsistent with the spec in that the spec states that “optional parameters can only appear for arguments of: packages, extern functions, extern methods, and extern object constructors” ([Section 6.7.2](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-optional-parameters)), but the [Tofino architecture](https://github.com/barefootnetworks/Open-Tofino/blob/master/share/p4c/p4include/tofino1arch.p4) apparently also includes the parser types and control types.

MollyDream:
- We added parser and control types to the types that allow optional parameters.

Fixes #977

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1697,7 +1697,7 @@ directions:
 A parameter that is annotated with the `@optional` annotation is
 optional: the user may omit the value for that parameter in an
 invocation.  Optional parameters can only appear for arguments of:
-packages, extern functions, extern methods, and extern object
+packages, parser types, control types, extern functions, extern methods, and extern object
 constructors.  Optional parameters cannot have default values.  If a
 procedure-like construct has both optional parameters and default values then it
 can only be called using named arguments.  It is recommended, but not
@@ -7888,7 +7888,7 @@ annotations named like `@X_annotation`
 
 ### Optional parameter annotations { #sec-optional-parameter-annotations }
 
-A parameter to a package, extern method, extern function or extern
+A parameter to a package, parser type, control type, extern method, extern function or extern
 object constructor can be annotated with `@optional` to indicate that
 the user does not need to provide a corresponding argument for that
 parameter.  The meaning of a parameter with no supplied value is


jnfoster:
We agreed that @jnfoster (and possibly others) would write some test cases to better understand how the compiler is type checking and simplifying parsers and controls with `@optional` parameters.

jnfoster:
Next step: we should write down the rules for when a `control` instance has a given `control` type. @jnfoster will draft.

jafingerhut:
I see this on the list of things to discuss at 2022-Apr-02 LDWG meeting.  There are a couple of action items mentioned in comments above from the last time it was discussed.  Should those things be prerequisites to discussing them in the LDWG again?  I am not claiming it should be a hard requirement -- just wondering if there is new information on those open questions.

jnfoster:
@mbudiu-vmw and I spent an hour reviewing these items and discussing the agenda. These will be ready to discuss on Monday.

jnfoster:
Here, the feature in question is already implemented in the open-source front end and the proprietary back-end from several vendors. So we believe this particular PR, which closes the gap between the spec and implementation, can be safely taken. Nevertheless, we will draft the rules mentioned as action items before Monday's meeting.

jnfoster:
Here's a small test program that shows the behavior of `p4c`:
```p4
control C();
control D(inout bit<8> x);
package P(C c, D d);

control MyY(@optional inout bit<8> x) { apply { } }
P(MyY(), MyY()) main;
```
As you can see, `MyY` is allowed to be used as the `C` and `D` parameter to `P`. 

I think we have a choice:
1. A control / parser with optional parameters has the type signature where any of those parameters are elided. Here, we allow `MyY` to be a control with `C`'s type (`control ()`) and `D`'s type (`control (inout bit<8>)`). This is what `p4c` appears to implement.
2. A control / parser with optional parameters has the type signature where none of those parameters are elided. Here we would only allow `MyY` to be a control with `D`'s type.

Option 1 has the most flexibility, but is arguably a bit odd, since a control instance can be used at multiple types. But I don't think it breaks anything...

Option 2 focuses on optional parameters as a way for callers to elide arguments. But it clearly doesn't change how lookups and matching works.

PR #989, opened 2021-12-02
Clarify multiple instantiations and invocations for Issue 926

### LINKED ISSUE #926: State and repeated invocations
Hi, I apologize if this has already been addressed in issues here. I'm trying to determine what's supposed to happen if a control instance `apply()` is called twice. Is the state of local variables persisted between the two calls, or is it like creating two separate instances and applying them one after the other? Is it even legal P4 to call a control (or parser) twice?

Here is an example: https://github.com/cornell-netlab/petr4/blob/36862d54cd037fa964e951db6883236c37485342/stf-test/custom-stf-tests/multiple-calls.p4

...and an STF test describing what I'd expect it to do if local variables persist: https://github.com/cornell-netlab/petr4/blob/36862d54cd037fa964e951db6883236c37485342/stf-test/custom-stf-tests/multiple-calls.stf

MollyDream:
- We clarified that local stateful instantiations are evaluated in every instantiation of controls and parsers
- We clarified that local variable declarations are evaluated in every invocation of controls and parsers
- We added reference to direct type invocation in the two sections of invoking parsers (12.10) and controls (13.4)

Fixes #926


mbudiu-vmw:
@MollyDream if you use `commit -s` with your commits you will make the cla bot happy.

jafingerhut (approved):
Except for one minor wording suggestion, this looks good to me.  I approve whether the wording changes are made, or not.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6007,6 +6007,19 @@ Figure [#fig-subparser] shows a diagram of this process.
...
+Thus, invoking a parser instance twice is different from invoking two parser instances each once, where the former
+accesses the same stateful constructs while the latter access two different copies.
+
+This exactly-once evaluation only applies to stateful instantiations. For local variable declarations,
+whether in `state`s or out, and whether with initializers or not, they are always evaluated
+when a parser instance is invoked. That is, local variables in a parser never persist across invocations. For variables
+declared outside `state`s, they are evaluated at the beginning of execution.
+
+When a parser with no constructor parameters is instantiated exactly once, 
+it is possible to invoke the parser directly (see Section [#sec-direct-invocation]).
```
jfingerh, line 6021: Suggested alternate wording for the last paragraph: "The above applies to all parser instantiations, regardless of whether they are explicit instantiations, or performed via direct type invocation (see Section [#sec-direct-invocation])."

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6834,6 +6847,19 @@ control Caller(inout Headers h) {
...
+Thus, invoking a control instance twice is different from invoking two control instances each once, where the former
+accesses the same stateful constructs while the latter access two different copies.
+
+This exactly-once evaluation only applies to stateful instantiations. For local variable declarations,
+whether in the `apply` block or out, and whether with initializers or not, they are always evaluated
+when a control instance is invoked. That is, local variables in a control never persist across invocations. For variables
+declared outside the `apply` block, they are evaluated at the beginning of execution.
+
+When a control with no constructor parameters is instantiated exactly once, 
+it is possible to invoke the control directly (see Section [#sec-direct-invocation]).
```
jfingerh, line 6861: Suggested alternate wording for the last paragraph: "The above applies to all control instantiations, regardless of whether they are explicit instantiations, or performed via direct type invocation (see Section [#sec-direct-invocation])."

MollyDream:
Ready for merging :)

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6007,6 +6007,19 @@ Figure [#fig-subparser] shows a diagram of this process.
 Note that since P4 requires declarations to precede uses, it is impossible to
 create recursive (or mutually recursive) parsers.
 
+When a parser is instantiated, all its local declarations of stateful instantiations are evaluated recursively.
+Each instantiation of a parser will have a unique set of value sets, extern objects, and inner parser instances.
```
jfingerh, line 6011: I will leave it to other LDWG members to decide whether the meaning is clear to them or not, but I wanted to point out that you can pass as compile-time parameters instances of objects like value sets, extern objects, and/or parser instances, and the instantiated parser can then use those pre-existing instances, as well as creating new instances of those things during the parser's instantiation.

Thus the statement "Each instantiation of a parser will have a unique set of value sets, extern objects, and inner parser instances." is true if all of those things are instantiated within the parser's declaration, but false if you pass pre-existing instances to the parser's instantiation.

The same applies to controls, too.

Similarly the statement below "invoking a parser instance twice is different from invoking two parser instances each one" can be true or false depending upon how the two parser instances are declared and instantiated.  For example, if there are NO local instantiations inside the parser declaration, and it ONLY invokes value sets, extern objects, and inner parser instances that were passed to it at instantiation time, you can have two different parser instances that DO update the same stateful objects, regardless of which parser you invoke at run time.

I think the "heart" of the motivation for these changes is that local variables behave differently than local instantiations of other things, and I'm perfectly fine with clarifying that behavior if it is unclear.  I am simply hoping we don't add other statements that might be confusing or subtly wrong in the process.

MollyDream: Hi Andy! Thanks for the review. I totally agree with your concern that no more confusion should be created during the process of solving one. Just my opinion on this specific issue: I think it should be clear by saying it is the "local declarations of stateful instantiations" that are evaluated recursively. The order of evaluation is explained in section [17.2. Compile-time Evaluation](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-ct-eval), which separates  constructor arguments from local instantiation. But that is just my opinion, and we can tweak the wording here to make things more clear.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6007,6 +6007,18 @@ Figure [#fig-subparser] shows a diagram of this process.
...
 
+When a parser is instantiated, all its local declarations of stateful instantiations are evaluated recursively.
+Each instantiation of a parser will have a unique set of local value sets, extern objects, and inner parser instances.
+Thus, invoking a parser instance twice is different from invoking two parser instances each once, where the former
+accesses the same local stateful constructs while the latter access two different copies.
+
+The exactly-once evaluation only applies to local stateful instantiations. For local variable declarations,
+whether in `state`s or out, and whether with initializers or not, they are always evaluated
+when a parser instance is invoked. That is, local variables in a parser never persist across invocations. For variables
+declared outside `state`s, they are evaluated at the beginning of execution.
```
jnfoster, line 6018: Is "evaluated" the right term here?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6834,6 +6846,18 @@ control Caller(inout Headers h) {
...
 
+When a control is instantiated, all its local declarations of stateful instantiations are evaluated recursively.
+Each instantiation of a control will have a unique set of local tables, extern objects, and inner control instances.
+Thus, invoking a control instance twice is different from invoking two control instances each once, where the former
+accesses the same local stateful constructs while the latter access two different copies.
+
+The exactly-once evaluation only applies to local stateful instantiations. For local variable declarations,
+whether in the `apply` block or out, and whether with initializers or not, they are always evaluated
+when a control instance is invoked. That is, local variables in a control never persist across invocations. For variables
+declared outside the `apply` block, they are evaluated at the beginning of execution.
```
jnfoster, line 6857: Ditto.

jnfoster:
We agreed to take this change at the April 2022 LDWG, after one more polishing pass.

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: In fact, I think the text could work fine without the second paragraph, both here and below.

mbudiu-vmw: @MollyDream if you can fix this before the next meeting we should discuss it.

jnfoster:
Subsumed by #1243

PR #990, opened 2021-12-03
Clarify rules for extern functions (#972)

### LINKED ISSUE #972: Are extern functions methods or functions
In Appendix F in the "can be a run-time parameter..." table, there's a column for functions and methods but there isn't a column for extern functions. Should they be classified under functions or methods, or can the methods column be renamed to externs to clear that up?

hackedy:
Fixes #972


jafingerhut (approved):
Looks good to me.

mbudiu-vmw:
I think we can merge this one without a discussion

hackedy:
Thank you!

Issue #1001, opened 2021-12-18
Constructor parameters are not explicitly listed as compile-time-known values


vgurevich:
The full discussion can be found in https://github.com/p4lang/p4c/issues/2995

The [Section 17.1 Compile-time known values](https://p4.org/p4-spec/docs/P4-16-v1.2.1.html#sec-ct-constants) omits constructor parameters from the list.

It is fairly clear from the other places in the spec that they should be there:

First, Section 6.7.1 says:

All constructor parameters are evaluated at compilation-time, and in consequence they must all be directionless (they cannot be in, out, or inout); this applies to package, control, parser, and extern objects. Values for these parameters must be specified at compile-time, and must evaluate to compile-time known values. See Section 14 for further details.

Section 14 says this, indeed:

Constructor parameters must be directionless (i.e., they cannot be in, out, or inout) and when the parser is instantiated, it must be possible to fully evaluate the expressions supplied for these parameters at compilation time.

So, it's pretty clear that the contructor parameters are compile-time known values.

vgurevich:
Good discussion happened on the corresponding pull request (https://github.com/p4lang/p4-spec/pull/1002) and the reference to the issue #932 has been made.

I understand the arguments. 

Practically speaking, there are the following problems:

1. If we literally interpret the spec the way it is written today, the following things that are **very common** in coding practice will be simply outlawed, specifically: 
1.a. The ability to use constructor parameters to instantiate nested parsers/controls and externs (the spec requires constructor parameters to be compile-time known values)
1.c. The ability to use constructor parameters to specify table sizes (the spec requires those to be compile-time known values)
2. There is a distinct need to be able to create parsers/controls parameterized on the types. At the very minimum, one should be able to specify the width of a type like `bit<X>. Currently, the only workaround is to use preprocessor, which is ugly.

One way or another we need to work on this. Perhaps, the same outcome can be achieved by allowing controls and parsers to have type parameters, but I am afraid that in any case we'll have to do type checking recursively.

QinshiWang:
An earlier issue: https://github.com/verified-network-toolchain/petr4/issues/218.

QinshiWang:
@vgurevich That's also the reason that I think this is useful. Not only for control and parsers, functions that are parameterized on width are also useful.

QinshiWang:
For the convenience of tracking, I also link this issue [#932] here (already mentioned in the PR).

apinski-cavium:
[...]

Issue #932, opened 2021-04-17
"Compile-time known" versus "statically known" versus "instantiation-time known"


QinshiWang:
I suggest distinguishing two kinds of "compile-time known values", namely "statically known values" and "instantiation-time known values", instead of coarsely saying "compile-time known values".
#### Phase distinction
In P4, there are actually three phases of evaluation: static evaluation, instantiation-time evaluation, and dynamic evaluation. Static evaluation happens when typechecking (which is a single pass). Instantiation-time evaluation happens when generating instances and going through the instance body to allocate resources. Dynamic evaluation happens when processing packets.

The first two phases both happen at compile time, so the values evaluated in both phases are literally compile-time known, but we need to be careful with the difference between them.
#### Examples
```
const bit<32> x = 1; /* statically known */
control MyC()(bit<32> v) { /* v is instantiation-time known */
  const bit<32> y = 2; /* statically known */
  const bit<32> c = v; /* instantiation-time known, because v is instantiation-time known */
  apply {
  }
}
```
#### Statically known values versus instantiation-time known values
It makes much sense to allow passing a constructor parameter (a instantiation-time known value) as constructor parameter of another constructor invocation. For example,
```
control C()(bit<32> v){ ... }

control MyC()(bit<32> v) {
  C(v) c;
  register<bit<8>>(v) reg;
  apply {
  }
}
```
But we don't want to allow using instantiation-time known values in typechecking. For example, this program should not be allowed:
```
control MyC()(bit<32> v) {
  bit<32> x;
  apply {
    f(x[v:10]); /* This is not allowed, because we don't know the type of x[v:10] during typechecking. */
  }
}
```
#### Problem with the current spec
1. It's confusing that constructor parameters are not included in "compile-time known values".
2. It looks useful to allow passing constructor parameters as constructor parameters.
#### Fix
We should clearly state the following in the spec:
1. Constants without constructor parameters are statically known values.
2. Constructor parameters and their derivatives are instantiation-time known values.
3. Both statically known values and instantiation-time known values are compile-time known.
4. Only statically known values can be used in typechecking.
5. Constructor parameters and directionless parameters (excluding action data) only need to be compile-time known, including constructor parameters.

QinshiWang:
I realize there are actually four, instead of three, phases. To be precise, constructor parameters and directionless parameters need to be distinguished, too. For example,
```
/* This makes sense. */
control MyC()(bit<32> size) {
  register<bit<8>>(size) reg;
[...]

PR #1002, opened 2021-12-19
Issue #1001: Add constructor parameters to the list of compile-time known values

vgurevich:
Added constructor parameters to the list of the compile-time known values. This was an unfortunate omission in the spec.


QinshiWang:
This is not absolutely true. Sometimes we want values that only dependents on the declaration but not instantiation. See my issue report for more information: https://github.com/p4lang/p4-spec/issues/932.

jnfoster:
Ahh, I'd forgotten about @QinshiWang's old writeup. Thanks for reminding me. Indeed, if we make this change, we would certainly not be able to do modular type checking. Perhaps there is a multi-pass, whole-program analysis that would be sound, but it'd be super complex and we've never written it down. I think there are good reasons to avoid it. 

So we need to discuss this...

mbudiu-vmw (approved):

jnfoster:
We discussed this again at the April '22 meeting. @QinshiWang and @jnfoster will discuss and come back to the group with a proposal for going forward...

mbudiu-vmw:
This will be superseded by #1213

mbudiu-vmw:
The main question is whether we merge this until we finalize #1213

jnfoster:

PR #1006, opened 2022-01-03
Remove instantiation from block for Issue 975

### LINKED ISSUE #975: Instantiation in a block
The [spec](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-block-stmt) allows instantiations to be a member of a block of statements, but there is no example of that allowed by p4c. Since in the “can be instantiated in this place” table ([Appendix F](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-calling-restrictions)) only parsers and controls have blocks, if an instantiation in a block is valid, it must appear in the parsers state or the control apply block. I tested all the possible instantiations following that table in p4c (parser in parser, extern in parser, value-set in parser, control in control, extern in control, table in control), and none of them are allowed in the block. Similar results are derived from testing with the Tofino compiler. I think it is worth discussing if an `instantiation` should be a member of `statementOrDeclaration` in the P4 grammar.
Litmus Test
```
extern register<T> {
    register(bit<32> size);
    void read(out T result, in bit<32> index);
    void write(in bit<32> index, in T value);
}

control compute2(inout bit<16> x) {
    apply { x = x + 1; }
}

control compute(inout hdr h) {
    register(1) r1; // valid
    compute2() c1; // valid
    apply {
        // register(1) r2; // invalid
        // compute2() c2; // invalid
    }
}
```
Based on the spec `r2` and `c2` should be instantiable in the control apply block, but p4c rejects them. One side note is that even though one may regard the direction invocation of a table as a hidden instantiation in a control apply block, the intrinsic instantiation of a table actually is likely to happen in the control local declaration rather than the apply block since that is also the case for direct type invocation, which is showed by an example in Section [14.1](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-direct-type-invocation).

MollyDream:
- We removed `instantiation` from `statementOrDeclaration` in grammar

Fixes #975

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -5130,7 +5130,7 @@ emptyStatement
 
 A block statement is denoted by curly braces. It contains a
 sequence of statements and declarations, which are executed
-sequentially. The variables, constants, and instantiations within a
+sequentially. The variables and constants within a
 block statement are only visible within the block.
 
 ~ Begin P4Grammar
@@ -5147,7 +5147,6 @@ statementOrDeclaration
     : variableDeclaration
     | constantDeclaration
     | statement
-    | instantiation
     ;
 ~ End P4Grammar
 
--- a/p4-16/spec/grammar.mdk
+++ b/p4-16/spec/grammar.mdk
@@ -518,7 +518,6 @@ statementOrDeclaration
     : variableDeclaration
     | constantDeclaration
     | statement
-    | instantiation
     ;
 
 /************ TABLES *************/


mbudiu-vmw (commented):
We should make this change in the compiler sources and see if nothing breaks.

jafingerhut:
Here is a PR for p4c that makes this change to the parser.  It appears that the only test cases that fail are two that check for almost nothing except whether an instantiation inside of a block is allowed or not, and they could easily be updated for this change if we think it is a good spec change: https://github.com/p4lang/p4c/pull/3014

mbudiu-vmw:
That is true, but the error messages you get are much better if you don't make the change.
So we should probably make the change in the spec but leave the compiler unchanged and add a comment explaining why the grammar is more permissive.

mbudiu-vmw (approved):

mbudiu-vmw:
I think we can merge this right away.

jafingerhut:
Created this p4c issue to add a comment to its grammar on why this unnecessary grammar production rule "statementOrDeclaration -> instantiation" is useful to keep in the compiler, i.e. for a clearer error message when an instantiation is done inside of a block: https://github.com/p4lang/p4c/pull/3016

jnfoster (approved):
Looks good to me...

PR #1007, opened 2022-01-03
Change semantics for setInvalid for Issue 995

### LINKED ISSUE #995: Assigning invalid header to union member vs Assigning to fields of invalid union member
In [8.18](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-expr-hu), the spec illustrates with an example that assigning an invalid header to a union member is the same as setting the union member to invalid.
```
u.hi = e
// is equivalent to
u.hi.setInvalid(); // when e is invalid
```
Further, it explains that `u.hi.setInvalid()` means that "if the valid bit for any member header of u is true then sets it to false, which implies that reading any member header of u will return an unspecified value."

Adding these together, writing an invalid header to a union member invalidates all the union members.

In contrast, in [8.22](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-uninitialized-values-and-writing-invalid-headers), the spec states that "if a write is attempted to a field of the invalid member header u1.h2, then any or all of the fields of the valid member header u1.h1 may change as a result. Such a write must not change the validity of any member headers of u1, nor any other state that is currently defined in the system, whether it is defined state in header fields or anywhere else." 

This means that writing to a field of an invalid member header does not invalidate other member headers. 

We are wondering if there is any particular reason for such a discrepancy. Is it motivated by the behavior of existing switches?

We tested with the v1model, but it does not invalidate other header members in both situations.
```
header H1 {
  bit<8> f;
}
header H2 {
  bit<16> g;
}
header_union U {
  H1 h1;
  H2 h2;
}

{
  U u;
  H1 my_h1 = { 8w0 }; // my_h1 is valid
  u.h1 = my_h1;

  H2 my_h2; // my_h2 is invalid
  u.h2 = my_h2;
  if (u.h1.isValid()) {
    h.output = 0xDCBA;
  } else {
    h.output = 0x0101;
  }
}
```
With the above P4 code snippet, the v1model writes `0xDCBA` to `h.output`. The behavior is the same if `u.h2 = my_h2;` is changed to `u.h2.g = 1;`

MollyDream:
- We changed the definition of `setInvalid` for header union members.
- We added the definition of `setInvalid` for header unions.

Fixes #995

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4348,9 +4348,9 @@ operations can be used to manipulate `u`:
   sets the valid bit for all other headers to `false`, which implies
   that reading these headers will return an unspecified value.
 
-- `u.hi.setInvalid()`: if the valid bit for any member header of `u`
-  is `true` then sets it to `false`, which implies that reading any
-  member header of `u` will return an unspecified value.
+- `u.hi.setInvalid()`: sets the valid bit for header `hi` to `false` and
+  keeps the valid bit for all other headers untouched, which implies
+  that reading `u.hi` will return an unspecified value.
 
 We can understand an assignment to a union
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
Sorry to be so noisy on this issue, but a couple of comments:

(a) I agree that header stacks are not a frequently used feature of the language today, and have no fundamental objections to this change.
(b) IF we make this change, are we willing to actually follow semantic versioning and call the next version of the P4_16 language spec version 2.0.0?  If not, why not?

jfingerh:
If we do make this change, and if this comment: https://github.com/p4lang/p4-spec/issues/995#issuecomment-996201592

applies after this change, but not before this change, then I think it would be good to add some kind of note about that requirement of how implementations must behave, similar to the example in the comment linked above.

mbudiu-vmw:
Can we close this PR? It is a significant change in the language semantics.

jnfoster:
This semantics seems more natural, so let's leave this open for now...

jnfoster:

PR #1008, opened 2022-01-03
Improve and move operations on sets for Issue 969

### LINKED ISSUE #969: Fix wording of set operations in spec
There are some inconsistencies about which types are supported (`bit<n>`, `int<n>`, serializable enums etc.)

MollyDream:
- We added a section to define _numeric types_, which is later referred to when defining the argument types of sets and indexes.
- We clarified the implicit casts of both set types and set operations.
- We added examples of implicit casts for set operations in implicit casts.
- We moved the section of _Operations on sets_ before the section of _Casts_.
- We adapted an existing example to generate a set of `int<W>`.

Fixes #969

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3700,153 +3700,6 @@ finding this information in a section dedicated to type `varbit`.
   inserts a variable-sized bit-string with a known dynamic width into
   the packet being constructed.
 
-## Operations on sets { #sec-set-exprs }
-
-Some P4 expressions denote sets of values (`set<T>`, for some numeric type `T`;
-see Section [#sec-set-types]). These expressions can
-appear only in a few contexts---parsers and constant table
-entries.  In both cases,
-implicit casts are applied recursively to every element in the set when needed.
-For example, the `select` expression (Section
-[#sec-select]) has the following structure:
-
-~ Begin P4Example
-select (expression) {
-   set1: state1;
-   set2: state2;
-   // More labels omitted
-}
-~ End P4Example
-
-Here the expressions `set1, set2`, etc. evaluate to sets of values
-and the `select` expression tests whether `expression` belongs to
-the sets used as labels.
-
-~ Begin P4Grammar
-keysetExpression
-    : tupleKeysetExpression
-    | simpleKeysetExpression
-    ;
-
-tupleKeysetExpression
-    : "(" simpleKeysetExpression "," simpleExpressionList ")"
-    | "(" reducedSimpleKeysetExpression ")"
-    ;
-
-simpleExpressionList
-    : simpleKeysetExpression
-    | simpleExpressionList ',' simpleKeysetExpression
-    ;
-
-reducedSimpleKeysetExpression
-    : expression "&&&" expression
-    | expression ".." expression
-    | DEFAULT
-    | "_"
-    ;
-
-simpleKeysetExpression
-    : expression
-    | DEFAULT
-    | DONTCARE
-    | expression MASK expression
-    | expression RANGE expression
-    ;
-~ End P4Grammar
-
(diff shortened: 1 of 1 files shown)


MollyDream:
When merging, the description of bit-slicing should be shortened by referring to the "numeric types".

MollyDream:
As @QinshiWang suggests, maybe we want to rename the "numeric types" to "integral types".

QinshiWang:
This PR contains two parts, (1) adding the concept of numeric types and cleaning the spec, and (2) adding explanations of set<T>. If not much labor is needed to separate them, we should do so.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4022,15 +4019,15 @@ select (hdr.ipv4.version) {
 
 ### Masks { #sec-cubes }
 
-The infix operator `&&&` takes two arguments of type `bit<W>` or
-serializable `enum`, and creates a value of type `set<ltype>`, where
-ltype is the type of the left argument.  The right value is used as a
+The infix operator `&&&` takes two arguments of the same numeric type (section [#sec-numeric-values]),
+and creates a value of the same type. 
+The right value is used as a
 "mask", where each bit set to `0` in the mask indicates a "don't care"
```
jafingerhut, line 4025: 2022-Mar-07 LDWG meeting discussed this section of the PR for several minutes, in particular regarding whether type `int` should be allowed as operands to a `&&&` operator, and if not, then when open source p4c today accepts surface syntax like `5 &&& 7` today, there is implicit casting occurrding, probably to some `bit<W>` type, and it seems that the bit width must be coming from the context such as `select` expression or P4 table definition of key fields.

jafingerhut: Suggestion from Nate: Dump intermediate output after every step of open source p4c to see what it is doing in its current implementation, in terms of perhaps adding implicit casts, and what those casts are, and where did those cast types come from?  This would at least better inform what changes to the spec we may want to make in this area.

mbudiu-vmw: In general the compiler treats `int` as a type variable and tries to unify with with other types. Often this will give a unique solution for the `int`, which will insert an (implicit) cast. Sometimes there will be no solution, e.g., in an expression like `2 == 3`, and then sometimes the compiler will complain (ideally if the width is necessary at runtime).

mbudiu-vmw (approved):

p4-16/spec/P4-16-spec.mdk
QinshiWang: `2 == 3` is a comparison between `int` (unbounded integer/mathematical integer). I think it should be evaluated at compile-time. The P4 standard allows compilers to reject some programs. Only accepted programs need to execute according to the manual.

mbudiu-vmw: This is just an example, there are other expressions involving ints that cannot be evaluated statically. An example is calling an external function.

QinshiWang: The type system should guarantee that `int` values are constants as they cannot be evaluated at runtime.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -2596,7 +2596,7 @@ types.
 
 #### Set types { #sec-set-types }
 
-The type `set<T>` describes _sets_ of values of type `T`. Set
+The type `set<T>` describes _sets_ of values of some numeric type `T` (section [#sec-numeric-values]). Set
```
mbudiu-vmw, line 2599: You can have sets of any other types.

jnfoster: @MollyDream can you address this? We're blocked on merging this until it's fixed.

jnfoster:
Subsumed by #1244.

Issue #1004, opened 2021-12-19
Name style for push_front and all other names defined in language spec


QinshiWang:
Other built-in functions are named in camel style, e.g. `isValid`, `setValid`, `setInvalid`. Maybe we want to make the change.

@jafingerhut edit to first comment to give a link directly to this comment below, which gives a much more full picture of all names that have snake case, camel case, and other styles in the current language spec.  The three names isValid, setValid, and setInvalid are just the tip of the ice berg: https://github.com/p4lang/p4-spec/issues/1004#issuecomment-1008338434

jafingerhut:
I know there isn't a mountain of P4_16 code out there, but there is a value to maintaining backwards compatibility where it isn't difficult to do so.

Adding new names for things is less disruptive to existing programs, versus removing existing names.  Even adding new names in this case seems fairly cosmetic to me.

vgurevich:
You can also look at #762 

But I have to admit that most probably the train has left the station.

jafingerhut:
In general, I'm terrible at coming up with great names for functions/methods/externs/etc.  But as LDWG members, maintaining backwards compatibility is one good reason to try to come up with reasonable names we are willing to live with long term (i.e. indefinitely, as long as the P4 language is used) when new names are added.  Even deprecating existing names causes churn in developer code that some could consider unprofessional conduct.  Just because some other language/library environments create such code churn, does not mean we should do so.

QinshiWang:
Thanks for the explanation. In my opinion, we'd better keep the old name but advocate a new name that's consistent with the general naming style in the spec.

jnfoster:
+1

I'd like to see is support a uniform style (camel case I guess) and deprecate the old names over time.

jafingerhut:
I promise to get off my soapbox after one more comment on this issue, but I simply cannot understand how deprecating currently defined names in the P4_16 language specification, not because they are functionally broken, nor insecure, nor any other kind of critical reason that one might consider removing a language feature for, but because "we wish we would have named them differently 5 years ago", is possibly consistent with this statement in the P4_16 language specification:

"One important goal of the P4_16 language revision is to provide a stable language definition. In other words, we strive to ensure that all programs written in P4_16 will remain syntactically correct and behave identically when treated as programs for future versions of the language."

In my opinion, part of backing up that statement with actions is sometimes saying, "Yeah, I wish we had done things differently, but it isn't actually broken, so we will keep some of the things the same that we wish we would have done differently."

[...]

PR #1009, opened 2022-01-06
Add new names noAction, pushFront, popFront, and actionRun

jafingerhut:
The original names are mentioned as still supported, for backwards
compatibility reasons.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -993,8 +993,8 @@ control TopPipe(inout Parsed_packet headers,
       */
      table check_ttl {
          key = { headers.ip.ttl: exact; }
-         actions = { Send_to_cpu; NoAction; }
-         const default_action = NoAction; // defined in core.p4
+         actions = { Send_to_cpu; noAction; }
+         const default_action = noAction; // defined in core.p4
      }
 
      /**
@@ -6356,7 +6356,7 @@ In addition, the tables may optionally define the following properties,
   lookup table fails to find a match for the key used.
 - `size`: an integer specifying the desired size of the table.
 
-The compiler must set the `default_action` to `NoAction` (and also
+The compiler must set the `default_action` to `noAction` (and also
 insert it into the list of `actions`) for tables that do not define
 the `default_action` property. This is consistent with the semantics
 given in Section [#sec-default-action]. Hence, all tables can be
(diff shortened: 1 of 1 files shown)


jafingerhut:
These proposed changes are not yet implemented in p4c.  I plan to create p4c issues with test programs that are the same as existing ones, but use the new names.

As far as I know, only push_front, pop_front, and action_run use the snake-case style names today.  I can add others to this PR if I missed any.

This PR addresses the snake_case to camelCase additions of new names portion of this issue: https://github.com/p4lang/p4-spec/issues/1004.  It does not attempt to address the method vs. field name questions raised in that issue.  If people want to do that, I intend that they be done in a separate PR from this one.  I am not personally interested in taking on the editing of those changes at this time.

jafingerhut:
Question: Should there be a new table property `defaultAction`, as the new alternative for the current `default_action`?

jnfoster:
While we're resolving this holy war, I think we should also mandate that P4 source code should be edited in Emacs :-P

I think I'd be in favor of gracefully shifting to using camelCase for everything (by adding, not deleting identifiers). 

But I could also see making some kind of distinction where snake_case is used for certain terms in the language (e.g., built-in types and types in the core library) provided we are consistent.

jafingerhut:
I am not in the holy war, except to say let us add names, not delete names :-)

Tell me what new names you want, and I can make the spec changes for it.  Mihai may have some preferences on whether we add `matchKind` as another alternative to `match_kind` in the p4c implementation, I would guess, but maybe that is easy, too.

jnfoster:
Continuing the religious theme, we may end up with a Tower of Babel...

I guess technically we are grabbing an identifier that may have been previously used. So adding these names would be a breaking change, unlike adding new operations `pushFront` and `popFront` for header stacks. It seems unlikely that someone would name their identifier `matchKind` given the existence of the built-in type, but you never know.

jafingerhut:
Depending on the precise form of the namespace/module proposal and whether it moves forward, name conflicts can be managed.

jnfoster:
When the namespace proposal lands, we will all enter the proverbial land of milk and honey...

jnfoster:
Also, this branch should be named `addCamelcaseForSeveralThings` :-)

jafingerhut:
Big laugh out loud for the "bug report" on the branch name.  My Lisp/Clojure affinities shine through there.

jafingerhut:
Hmm.  Thinking about the addition of a new action named `noAction`, that is currently not merely another name for `NoAction`, but a _different_ action which happens to have the same effect (i.e. no-op).  Using this name would be visible in the auto-generated control plane APIs for things like the P4Runtime API and TDI, and probably others.  There is likely controller software that has `NoAction` baked into it, and by adding the name `noAction` we would be effectively requiring them to also recognize the new name `noAction` specially, too.

jfingerh:
I am closing this PR, since there are potentially many more names than the few changed by this PR, depending upon what rules we want for consistent naming in the spec.  All of these questions are still in this open issue for future reference: https://github.com/p4lang/p4-spec/issues/1004#issuecomment-1008338434

PR #1012, opened 2022-01-13
Support for `typeof` keyword

mbudiu-vmw:
## Personnel
- [x] **Owner**: @mbudiu
- [x] **Supporters**: @vgurevich 

## Design
- [x] **Document**: this PR

## Implementation
- [x] **`p4-spec`**: this PR
- [x] **`p4c`**: https://github.com/p4lang/p4c/pull/3017

## Process
- [ ] **LDWG discussed**: 
- [ ] **LDWG approved**:
- [ ] **Merged into `p4-spec`**:
- [ ] **Merged into `p4c`**:

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3005,6 +3005,32 @@ datapath (e.g., a target specific tool-chain could include software
 that automatically converts values of type `PortId_t` to a different
 representation when exchanged with the control-plane software).
 
+## Using the type of an expression
+
+The `typeof` keyword can be used in a `typedef` declaration to obtain the
+type of an expression:
+
+~ Begin P4Grammar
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
This would be only really useful in conjunction with #1011. If #1011 is not accepted, a different design should be used for this feature. The reason is that `typedef` declarations being global confines the use of `typeof` only for constant expressions, and P4 currently does not have constant expressions of some types, such as header stacks.

jnfoster:
We discussed this feature at the April '22 LDWG meeting.
* Folks should try this feature out, as well as #1011 and come with feedback for the May meeting.
* @mbudiu-vmw will also prototype a version that allows `typeof` anywhere a type can be indicated... if Bison allows it.

jnfoster:
@vgurevich especially, please take a look!

PR #1016, opened 2022-02-07
Add bit-slicing on an infinite-precision integer for issue 1015

### LINKED ISSUE #1015: Slicing arbitrary-width integers
p4c compiler allows the statement `bit<4> x = 16[3:0]`. This suggests variable-width integers should support slices, but the spec does not list bit slices among the supported operations for arbitrary-width integers.

Based on the p4lang/p4c#3058, the spec should add this operation to section 8.7.

MollyDream:
- Add description for bit-slicing on integers. Note that the operand bound is different, L-value is impossible, and the evaluation equation is added.
- If "numeric types" are accepted, the type operand part can be shortened. ( relates to #1008 )

Fixes #1015

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3610,6 +3610,29 @@ expressions of type `int` must be compile-time known values.  The type
   The right operand must be either an unsigned constant of type `bit<S>` or a non-negative integer compile-time known value.
   The expression `a << b` is equal to $a \times 2^b$ while `a >> b`
   is equal to $\lfloor{a / 2^b}\rfloor$.
+- Extraction of a set of contiguous bits, also known as a slice, denoted by `[m:l]`,
+  where `m` and `l` must be expressions that evaluate to
+  non-negative compile-time known values, and `m >= l`. The types of `m` and `l` (which do not have to be identical)
+  must be one of the following:
+
+  - `int` - an infinite-precision integer (section
+    [#sec-infinite-precision-integers])
+  - `bit<W>` - a `W`-bit unsigned integer where `W >= 0` (section
+    [#sec-unsigned-integers])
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -3610,6 +3610,29 @@ expressions of type `int` must be compile-time known values.  The type
...
+  - `bit<W>` - a `W`-bit unsigned integer where `W >= 0` (section
+    [#sec-unsigned-integers])
+  - `int<W>` - a `W`-bit signed integer where `W >= 1` (section
+    [#sec-signed-integers])
+  - a serializable `enum` with an underlying type that is `bit<W>` or
+    `int<W>` (section [#sec-enum-types]).
+
+  The result is an unsigned bit-string of width `m - l + 1`, including the bits numbered
+  from `l` (which becomes the least significant bit of the result) to `m` (the
+  most significant bit of the result) from the source operand.
```
mbudiu-vmw, line 3629: Technically for int types we don't mandate a representation, so it is not clear what the bits of the source operand are.

jafingerhut: You will probably hear this multiple times from multiple sources, but Section 7.1 of the spec says "Fixed-width signed integers represented using two's complement int<>"

jafingerhut: I agree that doesn't specify for `int` types, only `int<W>`, but I'm all in favor of specifying it as two's complement for `int` as well, if that helps move the ball along.

mbudiu-vmw: Yes, that's my point. We don't mandate such a representation for int, but when we apply a slice we compute the int representation as a m-wide number using 2's complement. I think that should be part of the text.

mbudiu-vmw:
Superseded by #1149

PR #1017, opened 2022-02-08
Correct Madoko markup for table defining maxSizeInBits

jafingerhut:


jafingerhut:
Only the line involving `H[N]` seemed to be causing the warning from Madoko, but once I fixed that, I decided to also continue making the table look visually better, e.g. by making several of the long cell contents spread over two lines instead of one long line.

mbudiu-vmw (approved):

PR #1020, opened 2022-02-26
match_kind is a base type

mbudiu-vmw:

See https://github.com/p4lang/p4c/pull/3099 for several P4 programs that cannot be compiled without this change.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1828,6 +1828,7 @@ P4 supports the following built-in base types:
 baseType
     : BOOL
     | ERROR
+    | MATCH_KIND
     | BIT
     | INT
     | STRING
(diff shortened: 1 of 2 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -1828,6 +1828,7 @@ P4 supports the following built-in base types:
 baseType
     : BOOL
     | ERROR
+    | MATCH_KIND
```
jafingerhut, line 1831: Minor nit-pick, but should we go ahead and make the relative order of the other alternatives `BIT`, `INT`, and `STRING` be in the same order as each other between the grammar.mdk and spec files, too?

I see that `match_kind` was already mentioned in the text of the spec in Section 7.1, so this does look like it was simply an omission in the grammar.

jafingerhut (approved):
LGTM

jnfoster (approved):
I agree. This was just an omission.

PR #1023, opened 2022-03-01
Migrate CI from Travis to GitHub Actions

rst0git:
Since June 2021, the building on travis-ci.org has been disabled.
This commit replaces the Travis CI script with a GitHub Action workflow.

I have tested these changes in https://github.com/rst0git/p4-spec/tree/gh-pages


rst0git:
cc: @jnfoster @jafingerhut

jfingerh (approved):
As far as I can tell this looks fine.  The list of packages installed is similar to, but not identical to, what this script does: https://github.com/p4lang/p4-spec/blob/main/p4-16/spec/setup-for-ubuntu-linux.sh

I am guessing you have tried with these packages on an Ubuntu 20.04 system, and if it works, it all looks fine to me.  I will approve, with the understanding that while I can recognize shell commands, I know next to nothing about what is allowed in Github Actions definitions.

jfingerh:
More proof of my ignorance when it comes to Github actions: Is there a way to tell this is working before we merge it in?

What evidence/logs are left when these Github actions are merged in, so we can see if they are working as expected?

rst0git:
>  Is there a way to tell this is working before we merge it in?

Yes, I tested this in my fork: https://github.com/rst0git/p4-spec/tree/gh-pages/docs

> What evidence/logs are left when these Github actions are merged in, so we can see if they are working as expected?

The logs are available here: https://github.com/rst0git/p4-spec/runs/5383622687

jafingerhut:
@jnfoster OK with you to merge this in and see if it gets us closer to the goal?

jnfoster:
Let's do it!

Issue #1021, opened 2022-02-27
Literal syntax to represent the value of "an invalid header"


jafingerhut:
## Personnel
- [x] **Owner**: @jafingerhut 
- [x] **Supporters**: @mbudiu-vmw 

## Design
- [x] **Document**: See rest of this comment below

## Implementation
- [x] **`p4-spec`**: https://github.com/p4lang/p4-spec/pull/1184
- [x] **`p4c`**: https://github.com/p4lang/p4c/pull/3667
## Process
- [x] **LDWG discussed**: April 2021
- [x] **LDWG approved**: April 2023
- [x] **Merged into `p4-spec`**: April 2023
- [x] **Merged into `p4c`**: https://github.com/p4lang/p4c/pull/3667 (as experimental PR)

=======================================

On 2022-Nov-10 this experimental change to p4c was merged into its code base, making the syntax `(H) {#}` mean "the value that is an invalid header with type `H`", where `H` must be the name of a header type: 

I am not sure, but I think that an answer to this question _might_ be blocking the completion of an implementation of structure-valued expressions, which has been part of the language spec since May 2021, but has no implementation yet in p4c.  It has only a PR with a partial implementation here: https://github.com/p4lang/p4c/pull/2368

Related issue that raised this question in 2021-Apr: https://github.com/p4lang/p4-spec/issues/341 but this detailed part of it on a literal syntax to mean "a value that is an invalid header" remains unresolved so far.

See these sections of the spec:

+ "Structure-valued expressions" https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-structure-expressions (Section 8.12 in version 1.2.2)
+ "Initializing with default values" https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-initializing-with-default-values (Section 8.23 in version 1.2.2)

 (note that these links are to version 1.2.2 of the spec, not the latest main revision, so look for the sections with these names in the latest main revision to be completely up-to-date, in case they have changed)

Notes:

The section "Structure-valued expressions" says: "For a structure-valued expression typeRef is the name of a struct or header type."  It does not explicitly say whether the resulting header value is valid or invalid, but the intent was probably that a structure-valued expression with a header type always represents a header value that is valid.  If that is correct, it might be nice to have a phrase that says so explicitly in that section.

Aside, for which a separate issue has been created here (https://github.com/p4lang/p4-spec/issues/1032): Should structure-valued expressions allow `...` inside of them?

[...]

PR #1031, opened 2022-03-07
Issue #1021 - clarifications for structure-valued expressions

jafingerhut:


p4-16/spec/P4-16-spec.mdk
```diff
@@ -4763,7 +4772,7 @@ they do not have to be explicitly initialized to have a legal value.
 
 (Header types with no fields always have a validity bit.)
 
-## Initializing with default values
+## Initializing with default values { #sec-initializing-with-default-values }
```
mbudiu-vmw, line 4775: so perhaps this section should also go away. Structure-valued expressions generalize initializers.

jafingerhut: What about initializing with default values for all of the types that are neither struct nor header, e.g. enums, `bit<W>`, etc.?

mbudiu-vmw: You're right.

mbudiu-vmw (approved):

mbudiu-vmw:
Can we merge this?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3958,8 +3961,14 @@ assignments, in comparisons, in field selection expressions, and as
 arguments to functions, method or actions.  Structure-valued
 expressions are not left values.
 
-The compiler must raise an error if a field name appears more than
-once in the same structure-valued expression.
+Structure-valued expressions must provide a value for every member of
+the struct or header type to which it evaluates, either by doing so
+explicitly with a literal expression of the appropriate type, or by
```
mbudiu-vmw, line 3966: this could be slightly clearer

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3958,8 +3961,14 @@ assignments, in comparisons, in field selection expressions, and as
...
+using the `...` syntax for initializing with default values (see
```
jafingerhut, line 3967: AI Andy: Update this sentence to make it clearer that the 'or' does not imply that one must either specify ALL fields explicitly, or NO fields explicitly and use `...`.  The P4 developer is free to write any subset of field values, and follow them with `...`.

p4-16/spec/P4-16-spec.mdk
jafingerhut: @mbudiu-vmw @jnfoster I have replaced this unclear paragraph with 3 shorter paragraphs that are hopefully clearer.

mbudiu-vmw (approved):

jafingerhut:
Sorry for any confusion this might create, but it seemed more predictable in results if I created a fresh PR based on the latest main version of the spec, that had (almost) the same diffs.  See this PR: https://github.com/p4lang/p4-spec/pull/1251

The only difference is that I fixed a misspelling, and preserved a sentence that was in the latest main but not in latest main when this PR was originally created (the one saying "The order of the fields of the `struct` or `header` type does not need
to match the order of the values of the structure-valued expression.")

PR #1039, opened 2022-04-04
?: can be a compile-time constant

### LINKED ISSUE #1036: Section 17.1 compile-time know values vs ?: expressions
In section 8.4.1 Conditional operator, it is implied that ?: operator will be considered a compile time constant if the first operand is a compile time constant (when dealing with infinite precision integers). But when reading section 17.1 compile time know values, it is not listed as an expression which is will have a "compile-time known value".

mbudiu-vmw:
Fixes #1036


jnfoster (approved):
LGTM. Let's vet with the committee today.

PR #1040, opened 2022-04-04
One cannot have variables with type int

### LINKED ISSUE #1037: 10.2 variables vs 7.1.6.5 Infinite precision integers
10.2 variables has the following:
```
The language places few restrictions on the types of the variables: most P4 types that
can be written explicitly can be used (e.g., base types, struct, header, header stack, tuple).
```
while 7.1.6.5 Infinite-precision integers has the following:
```
This type is reserved for integer literals and expressions that involve only literals.
```

It would be useful is the 10.2 lists Infinite-precision integers explicitly as a restriction on a variable's type. So you don't need to go back and forth between the two sections.

mbudiu-vmw:

Fixes #1037


jnfoster (approved):
This is a clarification, so I think it's okay to take it. Or we can flash it up for the LDWG today.

PR #1041, opened 2022-04-04
Add bit-slice to enum for issue 992

### LINKED ISSUE #992: Specify that serializable_enum[hi:lo] is a valid L-value expression
At 2021-Dec-06 LDWG meeting during discussion of PR https://github.com/p4lang/p4-spec/pull/982, it was noticed that adding an explicit cast to an expression like `serializable_enum[3:0]` to `((bit<W>) serializable_enum)[3:0]` where `bit<W>` is the underlying type of the type of the serializable enum is not a valid L-value expression, so it seemed best if the spec did not define such an L-value as a case where an implicit cast could be added, at least not if the meaning of an implicit cast was the same as adding an explicit cast in the source code.

An alternate approach discussed then was to make a special case for expressions like `serializable_enum[3:0]` when used as L-values, I believe somehow defining such an expression as one that could be assigned a value of type `bit<4>` (or something that can be implicitly cast to `bit<4>`).

I may have misinterpreted the desired approach here, but wanted to create an issue that should remind others of the issue, and they can clarify in follow-up comments if this original description is incorrect.

MollyDream:
- Add the bit-slice operator to enum.
- Clarify when bit-slice can produce an l-value.
Fixes #992

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1462,9 +1462,9 @@ lvalue
   (using the dot notation).
 - References to elements within header stacks (see Section
   [#sec-expr-hs]): indexing, and references to `last` and `next`.
-- The result of a bit-slice operator `[m:l]`.
+- The result of a bit-slice operator on a l-value `[m:l]`.
 
-The following is a legal l-value: `headers.stack[4].field`.  Note
+The following is a legal l-value: `headers.stack[4].field`, but not `32w8[3:0]`. Note
 that method and function calls cannot return l-values.
 
 ## Calling convention: call by copy in/copy out { #sec-calling-convention }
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -3194,8 +3194,35 @@ casts `a`, which was initialized to `E.e2` to a `bit<8>`, using the specified
 fixed-width unsigned integer representation for `E.e2`, `1`.  The variable `b` is then set to the
 symbolic value `E.e2`, which corresponds to the fixed-width unsigned integer value `1`.
 
-Because it is always safe to cast from an `enum` to its underlying fixed-width integer type,
-implicit casting from an `enum` to its fixed-width (signed or unsigned) integer type is also supported (see Section [#sec-implicit-casts]):
+We can apply a bit-slice operator on an `enum` with an underlying type
```
mbudiu-vmw, line 3197: this is a long description, wouldn't it be shorter to say that this is equivalent to casting the enum to the underlying type and then applying the slice?

mbudiu-vmw (changes requested):

jnfoster:

PR #1042, opened 2022-04-04
Revert "Update parser value set types"

jnfoster:
Reverts p4lang/p4-spec#873


mbudiu-vmw:
Why should we revert the value set types?

apinski-cavium:
Yes I don't see why can't signed types be valid here? (the bitwise operators are valid for fixed signed integer types even).

For serializable enum case, should it decay to the underlying type instead or does it keep it as the enum type?

jnfoster:
I'd merged this prematurely at the last LDWG...

mbudiu-vmw:
Can we close this PR?

mbudiu-vmw:
I think it's too late now to consider this, we should just close it.

PR #1051, opened 2022-04-09
Small fixes suggested by Cristian Dumitrescu

jafingerhut:
The changes to the code snippet comments are because in the generated
HTML version those comments extend outside of a shaded background box.
By breaking the lines as suggested, I have verified that the text lies
within that shaded background box in both PDF and HTML versions.


jafingerhut:
I suspect these can be merged in without discussion in a work group meeting, but I will let Mihai and/or Nate at least have a chance to look them over before doing so.

mbudiu-vmw (approved):

PR #1063, opened 2022-04-21
Clearify structured-value expressions and their fields

apinski-cavium:
This clearifies a few things
* the order of the initialization does not matter
* All of the fields of the struct/header types need be initialized

These might seem implicit but when writing a specifications like this,
they should be explicit.


mbudiu-vmw (approved):

mbudiu-vmw:
I think this is a clarification and thus can be merged without a full discussion in the LDWG.

Issue #1068, opened 2022-04-27
Direct type invocation vs generic types


apinski-cavium:
I can't tell if this is a reference compiler issue or a spec issue or both.
Take:
```
parser MyParser<t>(t tt) {
    state start {
        transition accept;
    }
}
parser MyParser1() {
    state start {
        MyParser<bit>.apply(1);
        transition accept;
    }
}
```
I would assume this would work but instead the reference compiler gives a parser error:
```
t.p4(8):syntax error, unexpected .
 MyParser<bit>.
              ^
[--Werror=overlimit] error: 1 errors encountered, aborting compilation
```
If we do instead:
```
parser MyParser<t>(t tt) {
    state start {
        transition accept;
    }
}
parser MyParser1() {
    state start {
        MyParser.apply(1);
        transition accept;
    }
}
```
Which I had assume would figure out the generic type (but does not), we get:
```
t.p4(8): [--Werror=type-error] error: MyParser_inst.apply
 MyParser.apply(1);
 ^^^^^^^^^^^^^^^^^
  ---- Actual error:
  t.p4(1): 'int' type can only be unified with 'int', 'bit<>', or 'signed<>' types, not with 't'
  parser MyParser<t>(t tt) {
                  ^
  ---- Originating from:
  t.p4(8): Type of argument '1' (int) does not match type of parameter 'tt' (t)
   MyParser.apply(1);
                  ^
  t.p4(1)
  parser MyParser<t>(t tt) {
                       ^^
  t.p4(1)
  parser MyParser<t>(t tt) {
                  ^
  ---- Originating from:
  t.p4(8): Function type 'MyParser' does not match invocation type '<Method call>'
   MyParser.apply(1);
   ^^^^^^^^^^^^^^^^^
```
Which is a mouthfull for saying using a generic type without specializing it in this location.

apinski-cavium:
And if do this, thinking maybe using a typedef will workaround the issue, still no luck:
```
parser MyParser<t>(t tt) {
    state start {
        transition accept;
    }
}
typedef MyParser<bit> mytype;
parser MyParser1() {
    state start {
        mytype.apply(1);
        transition accept;
    }
}
```
Error message:
```
t.p4(9): [--Werror=type-error] error: Cannot extract field apply from mytype which has type Type(parser MyParser<...>)
 mytype.apply(1);
        ^^^^^
t.p4(9)
 mytype.apply(1);
 ^^^^^^
```

[...]

PR #1069, opened 2022-04-27
Allow direct invocation of generic controls/parsers 

mbudiu-vmw:
I could argue that this is a bug, but perhaps it should be discussed in the design working group.


mbudiu-vmw:
Fixes #1068

jnfoster:
Even for bug fixes, I think we should at least notify the LDWG in a meeting for something that is a material change like this.

mbudiu-vmw:
This should add an example.

mbudiu-vmw:
Have added an example, as suggested during LDWG meeting.

jnfoster (approved):

PR #1072, opened 2022-05-02
Explain overload resolution and type inference for constructors

### LINKED ISSUE #1067: constructors and generic types and instantiations and constructors
I didn't see anywhere in the spec which takes about figuring out the type for types while doing instantiations.

A simple example is:
```
parser parsertype<t>(t tt);
package p<t>(parsertype<t> a);
parser MyParser1<t>(t tt) (t y) {
    state start {
        transition accept;
    }
}
p(MyParser1(true)) m;
```
There should be a sentence or two added to the instantiation case where it says constructors are treated like method calls when figuring out the argument types.
If I get time next week, I will submit a patch (but I don't think I will have time).

mbudiu-vmw:
Fixes #1067

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4622,6 +4622,8 @@ constructors can also be called using named arguments.  Constructors
 are evaluated entirely at compilation-time (see Section
 [#sec-p4-abstract-mach]). In consequence, all constructor arguments
 must also be expressions that can be evaluated at compilation time.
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
@apinski-cavium is this precise enough?

apinski-cavium:
> @apinski-cavium is this precise enough?

Yes this is basically what I was going to add when I got some time.

mbudiu-vmw (commented):
Add a comma.

PR #1078, opened 2022-05-02
Explicitly disallow overloading parsers, controls, packages

### LINKED ISSUE #1055: Does parser overload check argument types
Take:
```
parser p( bit y ) {
    state start {
        transition accept;
    }
}

parser p(bit u, bit t) {
    state start {
        transition accept;
    }
}
parser p1(bit u, bit p2) {
    p() pp;
    state start {
        pp(u,p2);
        pp(u, 1);
        transition accept;
    }
}
```

The reference compiler errors out at the point of instantation:
```
t8.p4(13): [--Werror=duplicate] error: p: multiple matching declarations
    p() pp;
    ^
t8.p4(1): [--Werror=duplicate] error: Candidate: parser p
parser p( bit y ) {
       ^
t8.p4(7): [--Werror=duplicate] error: Candidate: parser p
parser p(bit u, bit t) {
       ^
```
I think it is right to do. The specification should be clarified to make this explictly invalid code. 

As an aside should it be at the point of definition rather than at the point of instantation?

mbudiu-vmw:

Fixes #1055

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2752,6 +2752,19 @@ either by the number of arguments or by the names of the arguments,
 when calls are specifying argument names.  Argument type information
 is not used in disambiguating calls.
 
+Notice that overloading of parsers, controls, or packages is not allowed:
+
+~ Begin P4
+parser p(packet_in p, out bit<32> value) {
+  ...
+}
+
+// The following will cause an error about a duplicate declaration
+//parser p(packet_in p, out Headers headers) {
+//   ...
+//}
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (commented):
In this example there are no constructor arguments at all, so overloading based on constructor arguments would not work.

mbudiu-vmw:
@vgurevich do you know if there are examples of people overloading such constructs in real programs?

rst0git (approved):
LGTM

jnfoster:
@mbudiu-vmw should we also add a ChangeLog for this?

PR #1079, opened 2022-05-02
isValid is a valid header field name

### LINKED ISSUE #1060: method and fields for header types (and scopes)
Take:
```
header h
{
  bit<2> isValid;
}
bool f(in h t)
{
  return t.isValid();
}
bit<2> f1(in h t)
{
  return t.isValid;
}
```
I would assume this is invalid code as isValid is (re)defined as a field rather than a method.
Are there two different scopes, one for methods and one for fields for header types?
or is this just a reference compiler bug?
The specification is not clear if a field could be named the same as a method; I had assume it could not.

mbudiu-vmw:
Fixes #1060

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2418,6 +2418,9 @@ variable-length fields is to define two headers -- one fixed length,
 one containing a `varbit` field -- and extract each part in separate
 parsing steps.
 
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
Tested in https://github.com/p4lang/p4c/pull/3275

jafingerhut (approved):
LGTM, in the sense of no objections to this change to the language spec.  I would object vehemently if any P4 developer tried to write this in clear maintainable P4 programs :-)

PR #1081, opened 2022-05-03
making more corrections [batch #2] to the P4₁₆ spec. based on reading version 1.2.2

Abe149:
Making more corrections to the P4₁₆ spec. based on reading version 1.2.2 [batch number 2].

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4689,7 +4689,7 @@ Calling the `isValid()` method on an element of a header stack, where
 the index is out of range, returns an undefined boolean value, i.e.,
 it is either `true` or `false`, but the specification does not require
 one or the other, nor that a consistent value is returned across
-multiple such calls.  Assigning an out of range header stack element
+multiple such calls.  Assigning an out-of-range header stack element
 to another header variable `h` leads to a state where `h` is undefined
 in all of its field values, and its validity is also undefined.
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3753,7 +3753,7 @@ The following casts are legal in P4:
 - `bit<W> → int<W>`: preserves all bits unchanged and reinterprets values whose most-significant bit is `1` as negative values
 - `bit<W> → bit<X>`: truncates the value if `W > X`, and otherwise (i.e., if  `W <= X`) pads the value with zero bits.
 - `int<W> → int<X>`: truncates the value if `W > X`, and otherwise (i.e., if `W < X`) extends it with the sign bit.
-- `bit<W> → int`: preserves the value unchanged but converts it to an unlimited-precision integer; the result is always positive
+- `bit<W> → int`: preserves the value unchanged but converts it to an unlimited-precision integer; the result is always non-negative
 - `int<W> → int`: preserves the value unchanged but converts it to an unlimited-precision integer; the result may be negative
 - `int → bit<W>`: converts the integer value into a sufficiently
   large two's complement bit string to avoid information loss,
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3900,7 +3900,7 @@ lp.learn( { hdr.ethernet.srcAddr, hdr.ipv4.src } );
 
 A list may be used to initialize a structure if the list has the same
 number of elements as fields in the structure. The effect of such an
-initializer is to assign to the i-th element of the list to the i-th
+initializer is to assign the nᵗʰ element of the list to the nᵗʰ
 field in the structure:
 
 ~ Begin P4Example
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4617,8 +4617,8 @@ Allocation of such objects can be performed in two ways:
 - Using instantiations, described in Section [#sec-instantiations].
 
 The syntax for a constructor invocation is similar to a function call;
-constructors can also be called using named arguments.  Constructors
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -3732,39 +3732,39 @@ finding this information in a section dedicated to type `varbit`.
...
-for programmers, it has several benefits:
+permitted on base types and derived types introduced by `typedef`, `type`, and `enum`.
+While this design is arguably more onerous for programmers, it has several benefits:
 
 - It makes user intent unambiguous.
 - It makes the costs associated with converting numeric values
-  explicit. Implementing certain casts involve sign-extensions, and
-  thus can require significant computational resources on some
-  targets.
+  explicit. Implementing certain casts involve sign extensions,
```
mbudiu-vmw, line 3740: Why not "involves"?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3732,39 +3732,39 @@ finding this information in a section dedicated to type `varbit`.
...
   specification. Some targets may not support all casts.
 
 ### Explicit casts
 
 The following casts are legal in P4:
 
-- `bit<1> <-> bool`: converts the value `0` to `false`, the value `1` to `true`, and vice versa.
-- `int -> bool`: only if the `int` value is `0` (converted to `false`) or `1` (converted to `true`)
-- `int<W> -> bit<W>`: preserves all bits unchanged and reinterprets negative
+- `bit<1> ↔ bool`: converts the value `0` to `false`, the value `1` to `true`, and vice versa.
```
mbudiu-vmw, line 3749: This arrow is unreadable at least in the markup file. Also, latex is not necessarily Unicode-friendly, did you check to see how it looks in the pdf?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4156,9 +4156,9 @@ of the header.
 "false". It can only be applied to an l-value.
 
 Similar to a `struct`, a header object can be initialized with a list
-expression [#sec-list-exprs] --- the list fields are assigned to the
+expression (see section [#sec-list-exprs]) --- the list fields are assigned to the
```
mbudiu-vmw, line 4159: Isn't "Section" supposed to be capitalized as well?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4210,11 +4210,11 @@ Given a header stack value `hs` of size `n`, the following
 expressions are legal:
 
 - `hs[index]`: produces a reference to the header at the
-  specified position within the stack; if `hs` is an l-value, the
-  result is also an l-value. The header may be invalid. Some implementations
+  specified position within the stack; if `hs` is an l-value,
+  the result is also an l-value. The header may be invalid. Some implementations
```
mbudiu-vmw, line 4214: I would appreciate refraining from making just space changes, they add to the review burden significantly.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4210,11 +4210,11 @@ Given a header stack value `hs` of size `n`, the following
...
   may impose the constraint that the index expression evaluates to a
-  compile-time known value. A P4 compiler must give an error if an
-  index value that is a compile-time constant is out of range.
+  value that is known at compile time. A P4 compiler must give an error
```
mbudiu-vmw, line 4216: Actually this is a term that we tried to use consistently through the document: "compile-time known value".

mbudiu-vmw (commented):
I will not continue to review this document. We had debates in the working group about the meaning of "compile-time known values" which are not settled. These changes cannot be approved.

If you would submit smaller PRs we could perhaps take them individually.

p4-16/spec/P4-16-spec.mdk
Abe149: AFAIK this spec is not written in the LᴀTᴇX language, but rather in the “Madoko” language.

mbudiu-vmw: Yes, but it produces two consumables: one in pdf, through latex, and one in html.
The question is about how these look.

p4-16/spec/P4-16-spec.mdk
Abe149: Yes, but it`s poor English.

I`m guessing nobody other than myself would use “{compilation time}-known value” or “compileTime-known value”, so I tried to come up with a semantically-equivalent phrase that is good English.

Abe149:
> If you would submit smaller PRs we could perhaps take them individually.

Understood.

p4-16/spec/P4-16-spec.mdk
Abe149: Fine with me.

Abe149: https://github.com/p4lang/p4-spec/pull/1081/commits/43f0d41e6a791c682ff802771f731346ca770cc7

p4-16/spec/P4-16-spec.mdk
Abe149: A pre-Abe bug I did _not_ catch right next to a bug I _did_ catch.  Thanks.  Adding a fix in my private repo. (should be included in my next PR).

p4-16/spec/P4-16-spec.mdk
Abe149: After some testing [using <https://hub.docker.com/r/p4lang/madoko-debian>], I found that Unicode support was broken by default.

After extensive experimentation, I found that only 1 out of 4 different ways of trying to fix it actually _works_.

image

image

Abe149: ### Before (from r1.2.2)
image

### After (from my latest WIP)
image

PR #1082, opened 2022-05-04
making more corrections to the P4₁₆ spec. based on reading version 1.2.2 ... based on branch 002 but eliminating the parts Mihai disliked

Abe149:


p4-16/spec/P4-16-spec.mdk
```diff
@@ -5276,8 +5266,8 @@ following behavior after the expression evaluation is interrupted.
 ## Conditional statement { #sec-cond-stmt }
 
 The conditional statement uses standard syntax and semantics familiar from many programming languages.
-However, the condition expression in P4 is required to be a Boolean (and not
-an integer). Conditional statements may not be used within a `parser`.
+However, the condition expression in P4 is required to be a Boolean (and not an integer).
+Conditional statements cannot be used within a `parser`.
```
ChrisDodd, line 5270: At one point I vaguely recall a discussion about adding a pass that will convert conditionals (or at least ifs) in parsers into additional states.  Regardless, this seems like something that should be architecture-dependent -- some architectures may have no problem with conditionals in parser while others may not be able to deal with them (or require converting them to additional states).  Did will ever add language about "may not" referring to architecture restrictions, while "shall not" (or something stronger) would refer to language restrictions?

I think there's some discussion of this in the architecture section, but probably not enough

ChrisDodd (approved):

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: Conditionals are currently fully supported in the parser. They have been since P4 1.2.2, in fact.

Abe149: Interesting.  I didn`t know about this, or at least not _much_.  I was just trying to improve the use of English grammar.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3900,7 +3900,7 @@ lp.learn( { hdr.ethernet.srcAddr, hdr.ipv4.src } );
 
 A list may be used to initialize a structure if the list has the same
 number of elements as fields in the structure. The effect of such an
-initializer is to assign to the i-th element of the list to the i-th
+initializer is to assign the nᵗʰ element of the list to the nᵗʰ
```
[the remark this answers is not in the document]
Abe149: AFAIK, this code is not HTML.  It`s some weird thing call "Madoko" [<http://madoko.org/reference.html>].

However, your implied point is well-taken that checking the spec. of the language in question isn`t a bad idea.  Here is what Madoko has to say about superscripts & subscripts:  http://madoko.org/reference.html#sec-sub--and-super-script

Checking my self-compiled PDF, I found that Madoko did not "like" what I did here with "nᵗʰ".  Using Unicode® for this may have worked if I had written it as e.g. `n&#`_something_`;&#`_some_other_thing_`;` [<http://madoko.org/reference.html#sec-unicode>], but that`s even _less_ readable than doing it the Madoko way IMO.

Thanks for the bug catch.  I will change this around soon-ish and re-upload [or maybe just pushing to my relevant branch will update this PR].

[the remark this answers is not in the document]
Abe149: Fixed in my WIP.

image

p4-16/spec/P4-16-spec.mdk
Abe149: @ChrisDodd & @mbudiu-vmw : is it OK to resolve this "conversation"?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5335,8 +5325,9 @@ following behavior after the expression evaluation is interrupted.
 ## Conditional statement { #sec-cond-stmt }
 
 The conditional statement uses standard syntax and semantics familiar from many programming languages.
+
 However, the condition expression in P4 is required to be a Boolean (and not
-an integer).
+an integ
```
mbudiu-vmw, line 5330: at least this one looks like a bug

Abe149: PR: [https://github.com/p4lang/p4-spec/pull/1098](https://github.com/p4lang/p4-spec/pull/1098)

PR #1086, opened 2022-05-06
Fix small typo in example in sec-implicit-casts section

apinski-cavium:
The ';' is wrong inside the enum and should be removed. Inside enums, ','
is the seperator but since there is only one value, it can be removed.


jnfoster (approved):

jnfoster:
Merging as this is a trivial typo.

PR #1125, opened 2022-08-05
Replace 'infinite' with 'arbitrary'

### LINKED ISSUE #1084: style (minor) issue, infinite-precision integer vs arbitrary-precision integer
I noticed in the specifications, there are places where arbitrary-precision integers is used and some places where infinite-precision integer. It would be a good idea to chose one or the other.
for an example
```
#### Infinite-precision integers {#sec-infinite-precision-integers}
```
and
```
## Operations on arbitrary-precision integers { #sec-varint-ops }
```

mbudiu-vmw:
Fixes #1084

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -40,7 +40,7 @@ pre, code {
   font-family: LuxiMono;
   font-size: 75%;
 }
-table { 
+table {
   breakable: true;
 }
 }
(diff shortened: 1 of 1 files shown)


jafingerhut:
2022-Aug-08 LDWG meeting comment: We should ensure that there is at least one occurrence of "infinite precision integer" type, and that it is the former name of what is now called "arbitrary precision integer", for anyone coming to the spec and searching for the older name.  I @jafingerhut recommend merging this PR, and I will create another PR on top of this one to add a new line in the change log with such a sentence.

jafingerhut (approved):

Issue #1113, opened 2022-06-08
Is auto default action constant or mutable?


QinshiWang:
When a table does not have a default action, we say the compiler must set `NoAction()` as the default action. In this case, is the default action mutable or constant?

jafingerhut:
I believe that many current P4 implementations default to mutable by the control plane if there is no explicit default_action table property, i.e. if you want it to be immutable, you must explicitly declare `const default_action = some_action;`

vgurevich:
I re-read the spec and it is a little vague there. In fact, it does not really discuss whether it is possible to change the default action via a control plane API :) Neither, by the way does it discuss what other things about a table can be changed. I think this is actually a good thing, because it allows for more possibilities.

I would state that if something is not clearly stated in the program, the implementation has a lot of latitude. For example, if it doesn't say `const`, there is no reason to prohibit something from being changed at runtime. Whether or not this will be possible is another matter. 

Now, this creates an interesting question: do we need to specify if other table attributes are mutable or immutable? What about the key? The list of actions? The size? Just to name a few that we typically think of being immutable, but that do not have to be that way on all the targets.

vgurevich:
@mbudiu-vmw On some targets, run-time API can offer a possibility to use another set of field(s) as a table key. We do not currently think of this and neither it is supported by P4Runtime, but it is possible and then the key becomes mutable.

QinshiWang:
@vgurevich I think it is pretty strange to allow the set of keys or the set of actions to be mutable. Both of them sound like modifying a function (action) at runtime.

mbudiu-vmw:
Indeed, what happens to the existing mapping key-actions in the table when the key fields change?

vgurevich:
@QinshiWang -- why do you find this "strange"? Computers write code all the time. Think about how the compiler creates the code for the P4 program: it creates some code for the table 1, table 2, etc. Depending on the target, it can do it at runtime too. There are devices (non-P4-programmable even, mind you) that do allow table keys to be specified at runtime. Why do you want to artificially limit the possibilities? 

Again, I am not saying that we need to mandate anything -- it is quite fortunate that P4 specification as it is today does not prohibit us from doing these and many more things like these.

@mbudiu-vmw -- such an operation will almost certainly require the table to be empty in order to be possible, but this is OK and is often preferred to program recompilation.

QinshiWang:
[...]

PR #1126, opened 2022-08-05
const default default_action

mbudiu-vmw:


mbudiu-vmw:
Fixes #1113

mbudiu-vmw:
This is a very delicate change, because it can break user programs. We should think carefully what to do about it.
The alternative is to change the spec in a backwards-incompatible way... we'll have to break something one way or the other.

mbudiu-vmw:
@jnfoster will submit a PR against the spec.

mbudiu-vmw:
I have updated this PR as discussed during the last LDWG. We now specify that the compiler-inserted default action is not const.

jnfoster (approved):
Thanks for taking care of this. LGTM

PR #1127, opened 2022-08-05
Clarifications on range operators

### LINKED ISSUE #1103: range operator (..) and serializable enums
Currently the specifications says:
```
The infix operator `..` takes two arguments of the same type `T`,
where `T` is either `bit<W>` or `int<W>`
```
But p4c accepts serializable enums on them and does some implict type conversions and even then it seems not very useful.
For an example:
```
enum bit<3> t
{
  value0 = 0,
  value1 = 1
}
parser MyParser1(in t tt1) {
  state start {
    transition select(tt1) {
     t.value0..t.value1: accept;
     t.value0..3w1: accept;
     //3w0..t.value1: accept;
     //3w0..3w1: accept;
     _: reject;
   }
 }
}
```
The commented out ones are not accepted.
It looks like following hold up for types:
enum..bit -> `set<enum>`
bit..enum -> `set<bit>`
bit..bit -> `set<bit>`
And there is no implicit coversion from `set<bit>` to `set<enum>` while doing the select either; there is one for `set<enum>` to `set<bit>` though.
(note I used bit here as a short cut for the underlying type of the enum).

Note Mask operator (&&&) has the following in the specifications which seems like where P4 is getting the types rules for range from too:
```
The infix operator `&&&` takes two arguments of type `bit<W>` or
serializable `enum`, and creates a value of type `set<ltype>`, where
ltype is the type of the left argument. 
```

### LINKED ISSUE #1104: range operator semantics is not clear if left > right
The current specification has this:
```
The set contains all values numerically between the
first and the second, inclusively.
```
and an example of `4w5 .. 4w8` but it does mention the semantics of what `4w8 .. 4w5` would be?
Is it an empty set or the "anti" set of `4w6 .. 4w7` that is the set that does not contain `4w6` or `4w7` but contains the two sets `4w0..4w5` and `4w0x8..4w0xff`?

mbudiu-vmw:
Fixes #1104 
Fixes #1103

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -40,7 +40,7 @@ pre, code {
   font-family: LuxiMono;
   font-size: 75%;
 }
-table { 
+table {
   breakable: true;
 }
 }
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):
Looks reasonable to me.

jafingerhut (approved):
merged PR also looks good

rst0git (approved):
LGTM

PR #1128, opened 2022-08-05
Specify result type for size* methods

### LINKED ISSUE #1096: the types for the size method calls is not specified
I noticed that the type for the size method calls (`minSizeInBits`, `minSizeInBytes`, `maxSizeInBits`,
and `maxSizeInBytes`) is not specified.
I suspect it should be an Infinite-precision integer type but this is inconstaint with the header stack `size` field which is bit<32>

mbudiu-vmw:
Fixes #1096


jnfoster (approved):
LGTM

PR #1129, opened 2022-08-08
static_assert definition

### LINKED ISSUE #1123: static_assert built-in function in core.p4
## Personnel
- [x] **Owner**: @mbudiu-vmw
- [x] **Supporters**: @vgurevich  

## Design
- [x] **Document**: See below

## Implementation
- [x] **`p4-spec`**: #1129 
- [x] **`p4c`**: https://github.com/p4lang/p4c/pull/3196
## Process
- [x] **LDWG discussed**: August 2022
- [x] **LDWG approved**: August 2022
- [ ] **Merged into `p4-spec`**:
- [x] **Merged into `p4c`**: https://github.com/p4lang/p4c/pull/3196

=======================================

It could also be in a separate header file, e.g., to preserve 100% backward-compatibility.

```
#ifndef _STATIC_ASSERT_P4_
#define _STATIC_ASSERT_P4_

/// Static assert evaluates a boolean expression
/// at compilation time.  If the expression evaluates to
/// false, compilation is stopped and the corresponding message is printed.
/// The function returns a boolean, so that it can be used
/// as a global constant value in a program, e.g.:
/// const version = static_assert(V1MODEL_VERSION > 20180000, "Expected a v1 model version >= 20180000");
extern bool static_assert(bool check, string message);

/// Like the above but using a default message.
extern bool static_assert(bool check);

#endif
```

Here is an example using it:

```
#include <v1model.p4>
#include <static_assert.p4>

const bool test = static_assert(V1MODEL_VERSION >= 20200000, "V1 model version is not >= 20200000");

control c() {
    apply {
        static_assert(2 != 1, "Arithmetic is broken");
        static_assert(3 == 1);
    }
}
```

Since the arguments are not "in", the function is evaluated at compilation-time.
I made it return a result - this way we don't need to add a new syntactic entity in P4 at the top-level (we don't have statements at the toplevel). But it needs to be used in a dummy constant as an expression.

mbudiu-vmw:
Fixes #1123

Compiler changes: https://github.com/p4lang/p4c/pull/3196

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -8377,6 +8377,7 @@ The P4 compiler should provide:
 ## Summary of changes made in version 1.2.4
 
 * Clarified that no direct invocation is possible for objects that have constructor arguments ([#sec-direct-type-invocation]).
+* Added `static_assert` function (Section [#sec-static-assert]).
 * Clarified semantics of ranges where the start is bigger than the end (Section [#sec-ranges]).
 * Allow ranges to be specified by serializable enums (Section [#sec-ranges]).
 * Specified type produced by the `*sizeInB*` methods (Section [#sec-minsizeinbits]).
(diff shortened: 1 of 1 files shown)


smolkaj:
This is a nice addition!

mbudiu-vmw:
Hopefully we can merge this at the next LDWG meeting.

jafingerhut (approved):
LGTM

jnfoster:
```p4
const bool pr_is_approved = static_assert("heck yeah" != "");
```

PR #1130, opened 2022-08-08
Clarify direct type invocation

### LINKED ISSUE #1068: Direct type invocation vs generic types

mbudiu-vmw:
Fixes #1068

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -7154,7 +7154,7 @@ the `udpSupport` parameter, as in the following example:
 GenericParser(false) topParser;
 ~ End P4Example
 
-## Direct type invocation
+## Direct type invocation { #sec-direct-type-invocation }
 
 Controls and parsers are often instantiated exactly once.  As a light
 syntactic sugar, control and parser declarations with no constructor
@@ -7220,6 +7220,10 @@ than once is defined as follows.
 
 See Section [#sec-name-annotations] for details of `@name`
 annotations.
+
+No direct invocation is possible for controls or parsers that require
+constructor arguments.  These need to be instantiated before they are
+invoked.
 
 # Deparsing { #sec-deparse }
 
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):
LGTM

Issue #1120, opened 2022-07-14
Tables with empty keys and without keys


QinshiWang:
The manual says as follows:
> If a table has no key property, then it contains no look-up table, just a default action—i.e., the associated lookup table is always the empty map.

I'm wondering if it also includes `key = { }`. I think literally it doesn't. It only means tables without the key property.

mbudiu-vmw:
These should be equivalent

QinshiWang:
@mbudiu-vmw Okay. Then ley's say it in the spec.

jnfoster:
Just to state the obvious, it could instead be that a table with `key = {}` can be in three configurations: 
- A single entry whose key is the empty tuple,
- No entries and no default action defined in the program or set by the control plane (so implicitly NoAction),
- No entries and a default action defined in the program or set by the control plane.

QinshiWang:
Currently, the syntax doesn't allow defining an entry with empty tuple.

jnfoster:
Except for tables with `const entries`, this is a P4Runtime question, not a P4 Language question.

QinshiWang:
If it's said there's no look-up table, then there are no entries, for both `const` and non-`const`. But if we don't say it, I agree that the control plane can have entries for `key = { }`

jnfoster:
I think we can go either way:
- Decide that `key = {}` is the same as no `key` property.
- Decide that `key = {}` is a lookup table with only one possible entry.
Either one is coherent. I have a slight preference for the latter, as it seems more coherent.

(I don't think this is a big deal in practical programs...)

QinshiWang:
It matters for practical programs. Practical programs do use `key = { }` or keyless (I'm not sure which one should be preferred) to encode a "direct action" table, so the name of the table is given by the program instead of the compiler.
```
    table tbl_clear_index {
        actions = {
            act_clear_index();
        }
        default_action = act_clear_index();
        size = 1;
    }
```

jnfoster:
Of course. What I meant is that I don't know if having both no-key and empty-key tables (with different semantics) is important in practical programs.

jafingerhut:
It seems difficult to imagine that allowing `key = {}` tables to have a single entry that is an "empty key" provides any additional power in the language.  Whether you can add such an entry or not, as long as the `default` action is not declared `const`, the runtime can change the behavior of a miss on the table any time it wants to, just as it could for a hit entry.

vgurevich:
[...]

PR #1132, opened 2022-08-13
Specify tables with empty key list to be the same as those with no key

jafingerhut:
Intended to resolve this issue https://github.com/p4lang/p4-spec/issues/1120

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6585,9 +6585,9 @@ result either an action or a "miss" indication, as described in
 Section [#sec-mau-semantics].
 
 If a table has no `key` property, or if the value of its `key`
-property is the empty list, then it contains no look-up table, just a
-default action---i.e., the associated lookup table is always the empty
-map.
+property is the empty list, i.e. `key = {}`, then it contains no
+look-up table, just a default action---i.e., the associated lookup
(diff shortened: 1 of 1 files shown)


jnfoster (approved):
This reflects the consensus opinion of the August LDWG discussion, so looks good to me!

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6584,9 +6584,10 @@ table is an abstract finite map that is given a key and produces as a
 result either an action or a "miss" indication, as described in
 Section [#sec-mau-semantics].
 
-If a table has no `key` property, then it contains no look-up
-table, just a default action---i.e., the associated lookup table is
-always the empty map.
+If a table has no `key` property, or if the value of its `key`
+property is the empty list, then it contains no look-up table, just a
```
rst0git, line 6588: Would it be possible to add an example of "empty list" as in P4 this is not obvious?

jnfoster: Good idea. Maybe just a parenthetical?

mbudiu-vmw (approved):

rst0git (approved):
LGTM

mbudiu-vmw:
We will have to make sure the compiler does this too.

PR #1133, opened 2022-08-13
Add note to revision history giving former name of arbitrary-precisio…

jafingerhut:
…n integers


jafingerhut:
Follow-up to PR https://github.com/p4lang/p4-spec/pull/1125, which I merged, since I could not quickly figure out how to add my changes to that PR.

PR #1134, opened 2022-08-18
Clarify semantics for table entries without key

rst0git:
This pull request clarifies that table entries cannot be specified for a table with no `key` property (or when the value of its `key` is empty list).


mbudiu-vmw (approved):

mbudiu-vmw:
I think we should be able to merge this without much process.

PR #1143, opened 2022-09-08
push_front and pop_front should only be usable from parser

apinski-cavium:
Just like how next, last and lastIndex are usable from parsers only, push_front and pop_front should only be usable from parsers.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4316,14 +4316,14 @@ manipulate the elements at the front and back of the stack:
   elements in the stack are discarded. The `hs.nextIndex` counter
   is incremented by `count`. The `count` argument must be a
   positive integer that is a compile-time known value. The return type
-  is `void`.
+  is `void`. May only be used in a `parser`.
 
 - `hs.pop_front(int count)`: shifts `hs` "left" by `count`
   (i.e., element with index `count` is copied in stack at index `0`).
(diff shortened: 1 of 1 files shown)


jafingerhut:
I will double-check, but I am fairly certain that there are P4 implementations with real hardware targets that implement push_front and pop_front in controls, by design.  These operations are useful to call from controls in order to implement features like MPLS label push/pop, for example, but also other features.

apinski-cavium:
But with next or last or lastIndex not being able to be used outside of a parser, there is no way to use those I suspect. Plus there are no loops, so there you would need to do `if(!hs[0].isValid()) if(!hs[1].isValid()) ....` to find the index which was invalidated or the new index.  So either we should remove the constraints on next, last, and lastIndex or add the constraints on these two.

apinski-cavium:
I see the bmv2 backend does support these in action's which does mean it is supported from the controls and there is a testcase p4_16_samples/header-stack-ops-bmv2.p4 where it tests to make sure it works.


So maybe this one is ok after all. lastIndex being accepted outside of parser is still a bug though.

jafingerhut:
push_front and pop_front can be used safely in a control, if you independently keep track of how many elements are valid before calling them.  That is typically how they are used, I believe.

apinski-cavium:
So here is the really odd thing, p4c (and p4test) rejects push_front and pop_front from the parsers ... I guess my patch was wrong for other reasons and even backwards from the current practice.

PR #1144, opened 2022-09-09
Add definition for compound assignment operators

rst0git:
This pull request adds definition for compound-assignment operators to the P4 language specification. These operators provide a shorter syntax for assigning the result of an arithmetic or bitwise operator. The proposed definition of compound assignment operators is similar the [C99 specification](https://en.cppreference.com/w/c/language/operator_assignment).


p4-16/spec/P4-16-spec.mdk
```diff
@@ -5228,6 +5228,36 @@ types (e.g. `structs`) are copied recursively, and all components
...
+| `&=`        | bitwise AND assignment           | `a &= b`  | `a = a & b`   |
+| `|=`        | bitwise OR assignment            | `a |= b`  | `a = a | b`   |
+| `^=`        | bitwise XOR assignment           | `a ^= b`  | `a = a ^ b`   |
+| `<<=`       | bitwise left shift assignment    | `a <<= b` | `a = a << b`  |
+| `>>=`       | bitwise right shift assignment   | `a >>= b` | `a = a >> b`  |
+|-------------|----------------------------------|-----------|---------------|
+
+The behavior of every builtin compound-assignment expression with the form `E1 op= E2`,
+where `E1` is a modifiable lvalue expression and `E2` is an rvalue expression,
+is exactly the same as the behavior of the expression `E1 = E1 op E2`.
```
mbudiu-vmw, line 5258: there's the point of side-effects in E1. Do they happen once or twice?
You could also specify that the operations are defined for all types where the expanded form is defined.

jafingerhut: I believe someone else mentioned this in another issue or PR, but explicitly saying that expression E1 is evaluated exactly once seems far preferable to saying that it is evaluated twice, or leaving the number of evaluations unspecified.

vgurevich:
Did we intentionally forget about `|+|` and `|-|` operations? :)

rst0git:
> Did we intentionally forget about `|+|` and `|-|` operations? :)

Thank you Vladimir for pointing this out! I have revised the pull request to include saturating arithmetic operations.

p4-16/spec/P4-16-spec.mdk
rst0git: I have added the following paragraph (similar to the [notes for assignment in C](https://en.cppreference.com/w/c/language/operator_assignment#Notes)) to describe side-effects in `E1`:

> The side effect of updating `E1` is sequenced after the value computations, but not the side effects of `E1` and `E2` themselves and the evaluations of the operands are, as usual, unsequenced relative to each other.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5228,6 +5228,41 @@ types (e.g. `structs`) are copied recursively, and all components
...
+| `|-|=`   | saturating subtraction assignment | `a |-|= b` | `a = a |-| b` |
+|----------|-----------------------------------|----------- |---------------|
+
+The behavior of every builtin compound-assignment expression with the form `E1 op= E2`,
+where `E1` is a modifiable l-value expression and `E2` is an r-value expression,
+is exactly the same as the behavior of the expression `E1 = E1 op E2`.
+
+The side effect of updating `E1` is sequenced after the value computations, but not
+the side effects of `E1` and `E2` themselves and the evaluations of the operands are,
+as usual, unsequenced relative to each other.
```
jafingerhut, line 5264: The current spec actually defines left-to-right evaluation of side effects, in Section 8.1 "Expression evaluation order": https://p4.org/p4-spec/docs/P4-16-v1.2.3.html#sec-expr-eval-order

apinski-cavium:
Are the "+=" operators two tokens or just one?
That is can you do:
a + = b;
or does it need to be without whitespaces?
a += b;

The earlier version only needs parser changes while the later needs both lexer and parser changes.
Also it would nice to update the grammar in the Appendix for these too.

mbudiu-vmw:
Not only in the appendix.
We should also make sure that this won't introduce conflicts in the bison parser (mostly likely not).
The implementation has some subtle points - do we want to eliminate these early, in a midend pass, or not at all?
If we eliminate them early in the frontend most passes won't need to learn about them.
Otherwise we have to add code for handling them in many other places.

apinski-cavium:
> We should also make sure that this won't introduce conflicts in the bison parser (mostly likely not).

It won't because statements are defined as:
lvalue = expression
lvalue ( ...

And lvalue does not have any of those tokens in it.
If it was a single token it would make my parser easier though because I actually handle
expression = expression 
as a way to handle things better for error reasons.

p4-16/spec/P4-16-spec.mdk
rst0git: Thank Andy! I have updated the pull request to mention about the description of compound assignment expressions in Section 8.1 "Expression evaluation order".

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5254,6 +5256,42 @@ types (e.g. `structs`) are copied recursively, and all components
...
+| `>>=`    | bitwise right shift assignment    | `a >>= b`  | `a = a >> b`  |
+| `|+|=`   | saturating addition assignment    | `a |+|= b` | `a = a |+| b` |
+| `|-|=`   | saturating subtraction assignment | `a |-|= b` | `a = a |-| b` |
+|----------|-----------------------------------|----------- |---------------|
+
+The behavior of every builtin compound assignment expression with the form `E1 op= E2`,
+where `E1` is a modifiable l-value expression and `E2` is an r-value expression,
+is exactly the same as the behavior of the expression `E1 = E1 op E2`, except that
+`E1` is evaluated only once. Unlike normal assignments, compound assignments evaluate
+the left-hand side before evaluating the right-hand side.
```
jafingerhut, line 5293: "Unlike normal assignments": Is this true?  Do you have any reference within the spec that defines the evaluation order of `E1` and `E2` in a normal assignment `E1 = E2`?

The sentence "All other expressions are evaluated left-to-right as they appear in the source program." in the section "Expression evaluation order" could be taken to not apply to `E1 = E2`, because that is a statement, not an expression.  If this strict reading is taken, and no other place in the spec mandates an order of evaluation between `E1` and `E2`, then I would agree that the current spec does not mandate this order.

It does seem a little odd to me to define left-to-right evaluation order within expressions, but NOT mandate that `E1` must be evaluated before `E2` in `E1 = E2`.

mbudiu-vmw: assignments are evaluated left to right in the compiler. the spec should say that.
it is one of the trickiest parts of the compiler, BTW, because the evaluation of the lhs can have side-effects, e.g.
```
void f(inout bit i);
a[f(i)] = a[f(i)];
```

rst0git: I used as a reference the description of augmented assignment statements in Python:
https://docs.python.org/3/reference/simple_stmts.html#augmented-assignment-statements

> Unlike normal assignments, augmented assignments evaluate the left-hand side before evaluating the right-hand side. For example, a[i] += f(x) first looks-up a[i], then it evaluates f(x) and performs the addition, and lastly, it writes the result back to a[i].

jafingerhut: @rst0git It sounds like from Mihai's comment above that Python should not be used as a reference here.  He believes that even normal assignments should have their left-hand side evaluated first.

rst0git: Thank you Andy and Mihai! I have updated the pull request accordingly.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5254,6 +5256,41 @@ types (e.g. `structs`) are copied recursively, and all components
...
+| `<<=`    | bitwise left shift assignment     | `a <<= b`  | `a = a << b`  |
+| `>>=`    | bitwise right shift assignment    | `a >>= b`  | `a = a >> b`  |
+| `|+|=`   | saturating addition assignment    | `a |+|= b` | `a = a |+| b` |
+| `|-|=`   | saturating subtraction assignment | `a |-|= b` | `a = a |-| b` |
+|----------|-----------------------------------|----------- |---------------|
+
+The behavior of every builtin compound assignment expression with the form `E1 op= E2`,
+where `E1` is a modifiable l-value expression and `E2` is an r-value expression,
+is exactly the same as the behavior of the expression `E1 = E1 op E2`, except that
+`E1` is evaluated only once.
```
jfingerh, line 5292: Maybe this is redundant given other text in the spec, but we may want to consider adding a sentence like this here: "Like other P4 expressions where subexpressions are evaluated left to right, `E1` is evaluated before `E2`."

jafingerhut (approved):
LGTM

jnfoster:

jafingerhut:
Reopening due to new interest in implementing this in p4c.

PR #1145, opened 2022-09-09
Specify core.p4 match_kind semantics

### LINKED ISSUE #1090: The core.p4 match kinds are not described in the P416 spec (but they are in PSA!)
This is a relatively minor issue with the P416 specification, but one that I found confusing.

The match kinds in core.p4 (exact, ternary, lpm) are not really explained in the P416 spec, apart from some very terse comments in Appendix D:
https://p4.org/p4-spec/docs/P4-16-working-spec.html#sec-p4-core-lib
No big deal; it's not hard to infer what these mean. 

But, I was surprised to find the complete explanations of the core.p4 match kinds in the latest version of the PSA spec, here:
https://p4.org/p4-spec/docs/PSA.html#sec-match-kinds

Please consider moving the descriptions of exact, ternary, and lpm match from the PSA spec to Appendix D of the P416 spec.

mbudiu-vmw:
Fixes #1090 
To be discussed at the LDWG


apinski-cavium:
+1 from me

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6583,6 +6583,31 @@ table is an abstract finite map that is given a key and produces as a
 result either an action or a "miss" indication, as described in
 Section [#sec-mau-semantics].
 
+The expected meaning of these values is as follows:
+
+- an `exact` match kind on a key field means that the value of the
+  field in the table specifies exactly all the bits for the key field.
+  This is applicable for key fields of types `bit<>`, `int<>`,
+  serializable enums, and structs or headers whose fields have such
+  types.
```
jafingerhut, line 6592: Would it be reasonable to generalize this to "all types for which equality is defined" ?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6583,6 +6583,31 @@ table is an abstract finite map that is given a key and produces as a
...
+
+- a `ternary` match kind on a key field means that the field in the
+  table specifies a set of values for the key field using a priority
+  and a mask.  A priority is a numeric value which is used to break
+  ties when a particular key belongs to multiple sets.  When table
+  entries are specified in the P4 program the priorities are generated
+  by the compiler; then entries are specified by the control-plane,
```
jafingerhut, line 6599: "then" -> "when" ?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6583,6 +6583,31 @@ table is an abstract finite map that is given a key and produces as a
...
+  the priority may need to be explicitly specified.  Entries with
+  higher priority are matched first.
+
+- a `lpm` (longest prefix match) match kind on a key field is a
+  specific type of `ternary` match where the mask is required to have
+  a form in binary that is a contiguous set of 1 bits followed by a
+  contiguous set of 0 bits.  Masks with more 1 bits have automatically
+  higher priorities.
+
+An example specifying entires for a table is given in Section [#sec-entries]
```
jafingerhut, line 6609: typoe: "entires" -> "entries"

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: Does the control plane support such types for keys? 
Perhaps the control-plane is the scope of a separate document, though, and this spec should be very liberal.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6583,6 +6583,30 @@ table is an abstract finite map that is given a key and produces as a
...
+- a `ternary` match kind on a key field means that the field in the
+  table specifies a set of values for the key field using a priority
+  and a mask.  A priority is a numeric value which is used to break
+  ties when a particular key belongs to multiple sets.  When table
+  entries are specified in the P4 program the priorities are generated
+  by the compiler; when entries are specified by the control-plane,
+  the priority may need to be explicitly specified.  Entries with
+  higher priority are matched first.
+
+- a `lpm` (longest prefix match) match kind on a key field is a
```
mbudiu-vmw, line 6602: a mask with only zeros is valid

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6583,6 +6583,30 @@ table is an abstract finite map that is given a key and produces as a
...
+  which have an equality operation.
+
+- a `ternary` match kind on a key field means that the field in the
+  table specifies a set of values for the key field using a priority
+  and a mask.  A priority is a numeric value which is used to break
+  ties when a particular key belongs to multiple sets.  When table
+  entries are specified in the P4 program the priorities are generated
+  by the compiler; when entries are specified by the control-plane,
+  the priority may need to be explicitly specified.  Entries with
+  higher priority are matched first.
```
mbudiu-vmw, line 6600: "higher" is implementation-dependent

jafingerhut:
@jafingerhut AI for Andy: check whether anywhere in the spec already says that when a table has multiple keys, an apply() operation only matches an entry if ALL the individual field matching criteria are true.  That is, logically there is an AND between the matching conditions between multiple keys of a table.

mbudiu-vmw:
Can we get some more comments on this PR? We would like to merge this in the next design meeting.
I will apply the suggested changes.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6620,6 +6620,33 @@ table is an abstract finite map that is given a key and produces as a
...
+  table specifies a set of values for the key field using a priority
+  and a mask.  A priority is a numeric value which is used to break
+  ties when a particular key belongs to multiple sets.  When table
+  entries are specified in the P4 program the priorities are generated
+  by the compiler; when entries are specified by the control-plane,
+  the priority may need to be explicitly specified.  Entries with
+  higher priority are matched first.  This specification does not
+  mandate whether "higher" priorities are represented by higher or
+  lower numeric values; this choice is left to the target
+  implementation.
```
jafingerhut, line 6640: "this choice is left to the target implementation".  It seems more precise to say that "this choice is determined by the run time API used"

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6796,12 +6823,14 @@ order, stopping at the first matching entry. Architectures should
...
 
-Depending on the `match_kind` of the keys, key set expressions may define
-one or multiple entries. The compiler will synthesize the correct number of
-entries to be installed in the table. Target constraints may further restrict
-the ability of synthesizing entries. For example, if the number of synthesized
-entries exceeds the table size, the compiler implementation may choose to issue
-a warning or an error, depending on target capabilities.
+Depending on the `match_kind` of the keys, key set expressions may
+define one or multiple entries, as described in Section
+[#sec-table-keys]. The compiler will synthesize the correct number of
```
jafingerhut, line 6828: I am not fundamentally against adding this phrase "as described in Section [#sec-table-keys}" here, but I do not see anywhere in that section that specifies how one key set expression may turn into one or multiple entries in a table.  It seems reasonable to me that the language specification NOT try to specify that level of detail.

Basically, the addition of this phrase seems confusing to me.  If I go where it says to go, it does not explain what the sentence before it claims it will.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6620,6 +6620,33 @@ table is an abstract finite map that is given a key and produces as a
...
+  lower numeric values; this choice is left to the target
+  implementation.
+
+- a `lpm` (longest prefix match) match kind on a key field is a
+  specific type of `ternary` match where the mask is required to have
+  a form in binary that is a contiguous set of 1 bits followed by a
+  contiguous set of 0 bits.  Masks with more 1 bits have automatically
+  higher priorities.  A mask with all bits 0 is legal.
+
+An example specifying entries for a table is given in Section [#sec-entries]
```
jafingerhut, line 6648: Missing period at end of sentence.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6620,6 +6620,33 @@ table is an abstract finite map that is given a key and produces as a
...
+
```
jafingerhut, line 6649: I do not know if this is the best place in the spec to say this, but it seems worth saying at least once somewhere in the spec something like the following:

"The text above defines the condition when one key lookup field matches one field within an entry added to the table.  If a table has multiple key field defined, then a lookup key only matches an entry if the match condition for all fields is true simultaneously for a single table entry."

I am betting there is probably a clearer way to state that.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6620,6 +6620,33 @@ table is an abstract finite map that is given a key and produces as a
...
+The expected meaning of these values is as follows:
+
+- an `exact` match kind on a key field means that the value of the
+  field in the table specifies exactly all the bits for the key field.
+  This is applicable for key fields of all types supported by the target
+  which have an equality operation.
+
+- a `ternary` match kind on a key field means that the field in the
+  table specifies a set of values for the key field using a priority
+  and a mask.  A priority is a numeric value which is used to break
```
jnfoster, line 6632: Is this a sufficient description of mask? Or should we expand?

jafingerhut:
@jafingerhut @vgurevich Pinging us to remind us to review the latest version of this PR, last updated on 2022-Nov-04.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6637,6 +6637,39 @@ table is an abstract finite map that is given a key and produces as a
...
+
+- an `exact` match kind on a key field means that the value of the
+  field in the table specifies exactly all the bits for the key field.
+  This is applicable for key fields of all types supported by the target
+  which have an equality operation.
+
+- a `ternary` match kind on a key field means that the field in the
+  table specifies a set of values for the key field using a value, a
+  mask, and a priority.  The meaning of the `(value, mask)` pair is
+  similar to P4 mask expressions, as described in Section { #sec-cubes
```
mbudiu-vmw, line 6650: this needs square braces

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6639,6 +6639,39 @@ table is an abstract finite map that is given a key and produces as a
...
+The expected meaning of these values is as follows:
+
+- an `exact` match kind on a key field means that the value of the
+  field in the table specifies exactly all the bits for the key field.
+  This is applicable for key fields of all types supported by the target
+  which have an equality operation.
+
+- a `ternary` match kind on a key field means that the field in the
+  table specifies a set of values for the key field using a value, a
+  mask, and a priority.  The meaning of the `(value, mask)` pair is
```
jafingerhut, line 6651: Comment on "and a priority".  One might get the impression from reading this that if the table has 3 ternary fields, then there are 3 priorities.  There are not.  There is only one priority value for the entire table entry.

One way to avoid causing such confusion would be not to mention priorities except in relation to an entire table entry, later.  When describing the match condition for a single field, omit any mention of priority.

Thus this bullet item could be replaced with:
```
- a `ternary` match kind on a key field means that the field in the
  table specifies a set of values for the key field using a value and
  a mask.  The meaning of the `(value, mask)` pair is
  similar to the P4 mask expressions, as described in Section [#sec-cubes]:
  a key field `k` matches the table entry when `k & mask == value & mask`.
```

Then later after all of the bullet points that describe exact, ternary, and lpm, only THEN bring up the issue that some table entries require a priority (those with at least one `ternary` field), and some table entries do not require a priority (those with no `ternary` fields).  The priority is not used to determine which entries match, only which of multiple matching entries "wins", and its action is executed.

p4-16/spec/P4-16-spec.mdk
jafingerhut: There is a sentence in the current PR addressing this comment, so I am resolving this.

mbudiu-vmw:
@jafingerhut we are supposed to merge this but we need an approval from you

jafingerhut:
I did add a couple of review comments on 2022-Dec-05, and at least one of them I consider serious enough NOT to approve this PR in its current form (the one about the implied priority-per-field, which is untrue).

I would prefer if changes could be made to this PR to resolve those comments before I approve, but if others want to approve this as it is, I won't stop them.

mbudiu-vmw:
Sorry, lost track which comments have been addressed.
Any others I should be heeding?

mbudiu-vmw:
@jafingerhut hopefully this is better. I had some conflicts when rebasing, hopefully I didn't break anything.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6292,6 +6292,38 @@ table is an abstract finite map that is given a key and produces as a
...
+  mask expressions, as described in Section [#sec-cubes]: a key field
+  `k` matches the table entry when `k & mask == value & mask`.
+
+- a `lpm` (longest prefix match) match kind on a key field is a
+  specific type of `ternary` match where the mask is required to have
+  a form in binary that is a contiguous set of 1 bits followed by a
+  contiguous set of 0 bits.  Masks with more 1 bits have automatically
+  higher priorities.  A mask with all bits 0 is legal.
+
+Some table entries, in particular the ones with at least one `exact`
```
jafingerhut, line 6314: `exact` -> `ternary`

p4-16/spec/P4-16-spec.mdk
jafingerhut: My previous comment still seems not addressed.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Still missing period at end of sentence.

mbudiu-vmw:
@jafingerhut maybe this is closer to a resolution

p4-16/spec/P4-16-spec.mdk
jafingerhut: The cross-reference has been removed, addressing my comment.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Updated text in commit 2 resolves this comment.

p4-16/spec/P4-16-spec.mdk
jafingerhut: confirmed that this has been fixed.

jafingerhut (approved):
LGTM.  I believe all of my (numerous) comments have been resolved.

PR #1146, opened 2022-09-12
Clarify inout parameters semantics

### LINKED ISSUE #1142: Unclear woring of `inout` parameters vs. initialization
I had a discussion with @bleibig regarding behavior of `inout` arguments and initialization. In [sec 6.7](https://p4.org/wp-content/uploads/2022/07/P4-16-spec.html#sec-calling-convention) it is mentioned that

> Direction `out` parameters are always initialized at the beginning of execution of the portion of the program that has the `out` parameters, e.g. control, parser, action, function, etc.  This initialization is not performed for parameters with any direction that 
is not `out`.

this suggests that initialization only applies to `out`.

However, there is also:

> `inout` parameters are both `in` and `out`. An argument passed as an `inout` parameter must be an l-value.

which suggests that `inout` has properties of both `in` and `out`, which could be interpreted to include the initialization mentioned before.

I would assume that `inout` arguments are not intialized (as if they were, they would be functionally the same as `out`), but this wording is inconclusive in my opinion.

mbudiu-vmw:
Fixes #1142

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1497,8 +1497,10 @@ Each parameter may be labeled with a direction:
   parameter must be an l-value; after the execution of the
   call, the value of the parameter is copied to the corresponding storage location
   for that l-value.
-- `inout` parameters are both `in` and `out`. An argument passed as
-  an `inout` parameter must be an l-value.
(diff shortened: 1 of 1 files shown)


jnfoster:
Oops. You beat me to it. Either this wording or #1150 (both small tweaks on @vlstill's text) is fine with me.

jafingerhut (approved):
LGTM

PR #1147, opened 2022-09-12
Clarified implicit casts in select expressions

### LINKED ISSUE #1106: range and mask operator and implicit casts
Take:
```
parser MyParser1(in bit<8> v) {
  state start {
    transition select(v) {
     (0x100)..(0x1ff): accept;
     (0x00)..(0x101): accept;
     _: reject;
   }
 }
}
```
it is not obvious what the range should do here? I think it should be saturation cast rather than a truncation cast because that would make better sense really. This is not mentioned anywhere in the specifications of what should happen here.

The specifications has the following for cast (sec-implicit-casts):
```
To keep the language simple and avoid introducing hidden costs, P4
only implicitly casts from `int` to fixed-width types and from enums
with an underlying type to the underlying type. In particular,
applying a binary operation (except shifts and concatenation) to an expression of type `int` and an
expression with a fixed-width type will implicitly cast the `int`
expression to the type of the other expression. 
```
But in this case we have int...int and set<int> as the type which is being casted to set<bit<8>> which is not defined here or anywhere really.

mask has a similar issue if the non mask part has any bits which are outside the number of bits;
that is `0x1ff&&&0x3ff` which for bit<8> should be always false.

From what I can tell p4c just truncates both operands of mask and range but that seems wrong.

mbudiu-vmw:
Fixes #1106

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3834,6 +3834,14 @@ the compiler will add implicit casts as follows:
 - `x[E.a:0]` becomes `x[(bit<8>)E.a:0]`
 - `E.a ++ 8w0` becomes `(bit<8>)E.a ++ 8w0`
 
+The compiler also adds implicit casts when types of different
+expressions need to ``match''; for example, as described in Section
+[#sec-select], since select labels are compared againt the selected
+expression, the compiler will insert implicit casts for the select
+labels when they have `int` types.  Similarly, when assigning a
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
> Shouldn't we also briefly mention this in sec-implicit-casts too, since it adds to the set of such casts?

Added a paragraph doing this.

jfingerh (approved):
These changes look reasonable to me, but I don't claim to be an authority on the nitty-gritty details of implicit casts in P4.

PR #1152, opened 2022-09-15
Reference C++11/C23 attributes rather than Java or C#

apinski-cavium:
Since more people will more likely know about C++11 attributes than C# or Java, make reference to that instead.
C23 adds the same kind of attributes as C++11 had added so reference C23 too.
C# and Java are less likely to be known to many developers.
Remove reference to C's `#pragma`s since C++11/C23 attributes are more similar and `#pragma` are less likely to be used now.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -7986,13 +7986,13 @@ error.
 
 # Annotations { #sec-annotations }
 
-Annotations are similar to C# attributes and Java annotations. They are a simple
-mechanism for extending the P4 language to some limited degree without changing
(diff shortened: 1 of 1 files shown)


jnfoster:
I personally think we should not define P4 language features by analogy in the spec. When it was first published, I removed many such references. I would propose that we do the same here, rather than edit Java to C.

apinski-cavium:
> I personally think we should not define P4 language features by analogy in the spec. When it was first published, I removed many such references. I would propose that we do the same here, rather than edit Java to C.

Agreed. made that change too.

jfingerh (approved):
LGTM

mbudiu-vmw:
This will be merged.

rst0git (approved):
LGTM

PR #1155, opened 2022-09-27
Clarify the meaning of 'const entries'

jafingerhut:


jafingerhut:
This PR is intended only to clarify the existing practice of P4 compilers and run times regarding their behavior when a table's entries are declared using `const entries` in the source code.  I am not aware of any P4 implementations that differ from the description in these changes.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4054,7 +4054,7 @@ the `struct` or `header` types do need to be initialized though.
 
 Some P4 expressions denote sets of values (`set<T>`, for some type `T`;
 see Section [#sec-set-types]). These expressions can
-appear only in a few contexts---parsers and constant table
```
jafingerhut, line 4057: This and the next small change in this PR are not strictly necessary, but the table property name is technically `entries`, not `const entries`, although currently the `const` is required whenever `entries` appears.  You can also consider these minor changes as prelude to defining new semantics for `entries` without `const`, which I hope to present to the LDWG next Monday (first Monday of October), if there is time available in the meeting's agenda.

mbudiu-vmw, line 4057: we may change our mind and allow non-const entries too.

mbudiu-vmw:
You can paste that item in the agenda for next week.

jnfoster (approved):
These seem like helpful clarifications. I approve.

jfingerh:
@mbudiu-vmw "You can paste that item in the agenda for next week." Sorry for my ignorance, but I don't know where to go to do such pasting.  If you do not mind adding it to the agenda, that would be much appreciated.

jnfoster:
Here:
https://docs.google.com/document/d/1XSgdXeG1UuF1FM_XAqxDrHeN4dHZWBnJPKVS6SnGNwM/edit

mbudiu-vmw:
This link is on the front page of the spec repository.

mbudiu-vmw:
To add a changelog entry

mbudiu-vmw (approved):
I think we agreed to merge this

PR #1156, opened 2022-09-28
Change list expression to tuple expression

jafingerhut:


jafingerhut:
This has not been discussed, except in a discussion thread on https://github.com/p4lang/p4c/pull/3520, in particular here: https://github.com/p4lang/p4c/pull/3520#issuecomment-1251748379

Mihai, is your main concern here with all of the class and method names in p4c, and comments, and perhaps also some documentation about p4c internals, that use the phrase "list" and "list expressions" that would become out of date with the spec if these changes were made?

Or do you have concerns independent of that?  (I am not claiming that such concerns are unimportant, by any means.  I am mainly curious to be clear on the concerns.)

mbudiu-vmw (approved):
We should discuss this in the LDWG.

jnfoster (approved):
LGTM! Agree we should vet with the LDWG.

jafingerhut:
@chrisdodd @hanw @vgurevich This is "only" a change in terminology in the spec, primarily replacing occurrences of "list expression" with "tuple expression".  Just wanted to give you a heads up on this potential change, as it would potentially change terminology that one might want to use in training material and documentation (not an urgent change, but some day), and Mihai also pointed out that the p4c front end IR data structures would still have many occurrences of "list" in the names of some IR node types that would then no longer match the spec terminology, which Mihai is OK with, and would not want to change, because of the disruption that would cause in P4 compiler back ends using the p4c front/mid end.

mbudiu-vmw:
This needs a changelog

jfingerh:
Is there any chance I could convince you to merge this in if you approve of it, under the strict promise that I will add a revision history item as a separate PR after this is merged in?  My local git clone is in a state where that would probably be significantly easier for me.

PR #1163, opened 2022-10-06
Add a note about duplicated table properties

### LINKED ISSUE #1162: Duplicated table properties
I didn't see anything in the specifications which says that duplicated properties is invalid but p4c rejects it.
I agree this should be invalid even and not last one wins situtation even.

The specifications does say this:
```
Syntactically a table is defined in terms of a set of key-value properties.
```
But that could mean last one wins, errors out on duplicate or even merge them. It would be best if it was explict of what happens here.

test example:
```
action a() {}
control c() {
    table t {
        actions = { .a; }
        actions = { .a; } // { dg-error "uplicates" }
        size = 100;
        size = 100; // { dg-error "uplicates" }
        default_action = a;
        default_action = a; // { dg-error "uplicates" }
    }
    apply {}
}
```

apinski-cavium:
Even though the spec says table properties is
implemented as key-value pair, it could have been
the case where the last pair would have won.
This makes it explict that the if there is a
duplicated property, then the compiler should
reject it.

Fixes #1162

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6520,7 +6520,8 @@ necessary to disambiguate.
 Syntactically a table is defined in terms of a set of key-value
 properties. Some of these properties are "standard" properties, but
 the set of properties can be extended by target-specific compilers as
-needed.
+needed. Note duplicated properties are invalid and the compiler should
+reject them.
 
 ~ Begin P4Grammar
 tableDeclaration


mbudiu-vmw (approved):

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6520,7 +6520,8 @@ necessary to disambiguate.
 Syntactically a table is defined in terms of a set of key-value
 properties. Some of these properties are "standard" properties, but
 the set of properties can be extended by target-specific compilers as
-needed.
+needed. Note duplicated properties are invalid and the compiler should
```
jnfoster, line 6523: Suggest rephrasing:

> Note that each property may be listed at most once in a given table declaration. The compiler should reject table declarations containing multiple definitions of the same property.

jnfoster (approved):

jafingerhut (approved):
LGTM

mbudiu-vmw:
Is @apinski-cavium still supposed to improve the wording?

PR #1171, opened 2022-10-25
Add assert and assume to PSA include file to match p4c, plus minor changes

jfingerh:


jfingerh:
These changes bring the version of the psa.p4 include file in the p4-spec repo up to date with the latest versions in the p4c/p4include directory.  The biggest change is adding definitions of assert  and assume, which have been part of the p4include repo versions for some time now (added to psa.p4 include file in p4c repo shortly after they were added to v1model architecture include file in p4c).

jafingerhut:
FYI @mariobaldi please consider reviewing.  Happy to talk live about this if there is anything that seems off.

mariobaldi (approved):

apinski-cavium:
Hmm, this just adds it to the include file but I would assume the psa specifications would have these documented too?

apinski-cavium:
Also it has a p4c specific command line option. Will other implementations of PSA will require --ndebug option or can there be another option name for these?

jfingerh:
I am happy to leave out the `-ndebug` information if you think it doesn't belong there, but "how can I make all asserts into no-ops" at compile time seems like such a commonly wished thing that at least mentioning the possibility, and where to learn more, seems useful to have in these comments somehow.  Any recommendations there?  Perhaps "If you use a P4 compiler based on the p4c front end, then ..."?

jfingerh:
@apinski-cavium And I agree that it would be best to add text to the body of the PSA specification that document assert and assume, if they are in the include file.  I have just created this issue to track that: https://github.com/p4lang/p4-spec/issues/1172

apinski-cavium:
> I am happy to leave out the `-ndebug` information if you think it doesn't belong there, but "how can I make all asserts into no-ops" at compile time seems like such a commonly wished thing that at least mentioning the possibility, and where to learn more, seems useful to have in these comments somehow. Any recommendations there? Perhaps "If you use a P4 compiler based on the p4c front end, then ..."?

How about this:
The compiler should provide a way to the asserts into a no-op; for an example if using a P4 compiler based on the p4c front-end, the option `--ndebug` is the way to turn asserts into a no-op.

I don't know if the word `should` should be that or `must` there.

mariobaldi:
This text looks good to me.

It seems to me that “should” is the right choice of wording here. I don’t think we need to mandate that compilers support turning asserts in no-ops.

 

Thanks,

---

Mario

jfingerh:
@apinski-cavium I made some changes to the wording of the comments regarding p4c and the `--ndebug` option.  Please take a look and see if it looks acceptable.

apinski-cavium:
> @apinski-cavium I made some changes to the wording of the comments regarding p4c and the `--ndebug` option. Please take a look and see if it looks acceptable.

Perfect.

PR #1178, opened 2022-11-04
Specify that comparison is legal for tuples

### LINKED ISSUE #1094: eq and ne is currently not allowed on tuple expressions
I don't see why '==' and '!=' are not mentioned in #sec-tuple-exprs as operations on it. It is very similar to struct and should have the same behavior as struct really.

The other thing is in #sec-ops-on-structs, "all of their fields can be recursively compared for equality." is not needed as all of the types allowed after tuples are added to the list of allowed for equality.

I will submit a pull request for both of these issues over the weekend. I just wanted to file this issue now so I don't forget.

mbudiu-vmw:
Fixes #1094


apinski-cavium:
+1

jafingerhut (approved):
LGTM

PR #1179, opened 2022-11-04
Narrow the legal types that can be introduced by `type`

### LINKED ISSUE #1044: type keyword only applies to base types?
I didn't know if I should filed this as a specification bug or as a bug in the reference compiler but I think it is just a specification issue.
Take:
```
header t
{
 bit y;
}
type t tyu;
```
The specifications in section 7.5 there is no requirement for type keyword to only handle with a base type (or a type which is type or a base type) but the reference compiler rejects the code:
```
t2.p4(6): [--Werror=type-error] error: tyu: `type' can only be applied to base types
type t tyu;
       ^^^
```

mbudiu-vmw:
Fixes #1044 
To be discussed by the LDWG.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3029,7 +3029,6 @@ new type.
 
 ~ Begin P4Grammar
     | optAnnotations TYPE typeRef name
-    | optAnnotations TYPE derivedTypeDeclaration name
 ~ End P4Grammar
 
 ~ Begin P4Example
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
See also https://github.com/p4lang/p4c/pull/3664

apinski-cavium:
+1 from me.

jafingerhut (approved):
LGTM

jafingerhut (approved):
LGTM

Issue #1159, opened 2022-10-02
Define "initial entries" for P4 tables, in addition to currently defined "const entries"


jafingerhut:
## Personnel
- [x] **Owner**: @jafingerhut 
- [x] **Supporters**: @thantry 

## Design
- [x] **Document**: https://docs.google.com/presentation/d/1GW-9FDtyLuwqV_mdGeDh_qqDiey0VkQ1/edit?usp=sharing&ouid=113165450274035133953&rtpof=true&sd=true

## Implementation
- [x] **`p4-spec`**: There are at least two different proposed syntaxes in two different PRs (1) https://github.com/p4lang/p4-spec/pull/1180 (2) https://github.com/p4lang/p4-spec/pull/1231 For a summary of differences see here https://github.com/p4lang/p4-spec/pull/1180#issuecomment-1504165720   At the 2023-May-01 LDWG meeting, there was general agreement preferring the backwards-compatible syntax extension decribed in #1180 over #1231
- [x] **`p4c`**: Step 1 of implementation, adding only the new parts of the grammar https://github.com/p4lang/p4c/pull/3748
- [x] **`p4c`**: Step 2 of implementation, generating correct P4Info file generated by p4c: https://github.com/p4lang/p4c/issues/4016 (must wait until P4 API work group approves how this feature is represented in P4Runtime API)
## Process
- [x] **LDWG discussed**: October 2022: https://docs.google.com/document/d/1XSgdXeG1UuF1FM_XAqxDrHeN4dHZWBnJPKVS6SnGNwM/edit#heading=h.37p6wudxij34
- [x] **LDWG approved**: Discussed last at 2023-May-01 LDWG meeting, and PR #1180 version was preferred over the alternative #1231
- [x] **Merged into `p4-spec`**: This PR was merged on 2023-May-15: https://github.com/p4lang/p4-spec/pull/1180
- [x] **Merged into `p4c`**: 

=======================================

mihaibudiu:
https://github.com/p4lang/p4c/pull/3748 is the proof-of-concept, I will finalize it based on the accepted syntax.

jnfoster:
Closing as the P4C PR has also been merged.

PR #1180, opened 2022-11-08
Add definition of initial table entries, without const qualifier

jfingerh:
Summary of comments with approve/disapprove/change-requests as of 2022-Dec-27:

+ Sander Tolsma approved on 2022-Nov-16 after several of his review comments were addressed.
+ Mario Baldi approved in 2022-Nov
+ Steve Licking  approved in 2022-Nov
+ Mihai Budiu wrote several comments and created partial implementation in p4c.  Not clear whether he approves yet.
+ Andrew Pinski wrote some comments.  Overall approves of the idea, but would prefer a different syntax for specifying priorities of entries.

@thantry Please do provide your review comments, or approval of this PR as written.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6934,13 +6934,14 @@ Unfortunately there are two commonly used, but different, ways of
 interpreting numeric priority values.
 
 The P4Runtime API requires numeric priorities to be positive integers,
-and define that entries with larger priorities must win over
-entries with smaller priorities.  We will call this convention
-`largest_priority_wins`.
+i.e. 1 or larger, and defines that entries with larger priorities must
+win over entries with smaller priorities.  We will call this
+convention `largest_priority_wins`.
 
-TDI requires numeric priorities to be non-negative integers, and
-define that entries with smaller priorities must win over entries with
-larger priorities.
+TDI requires numeric priorities to be non-negative integers, i.e. 0 or
+larger, and defines that entries with smaller priorities must win over
+entries with larger priorities.  We will call this convention
+`smallest_priority_wins`.
 
 #### Size {#sec-size-table-property}
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -7092,7 +7092,7 @@ for developers who explicitly choose to specify entries in an order
 that do not match their relative priority order.
 
 The following example is the same as the first example in section
-[#sec-entries], except for the definition of table `t_entry_ternary`
+[#sec-entries], except for the definition of table `t_exact_ternary`
 shown below.
 
 ~ Begin P4Example
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6895,12 +6895,12 @@ table t_exact_ternary {
     priority_delta = 10;
     @noWarn("duplicate_priorities")
     entries = {
-        const priority=10 (0x01, 0x1111 &&& 0xF   ) : a_params(1);
-                          (0x02, 0x1181           ) : a_params(2); // priority=20
-                          (0x03, 0x1000 &&& 0xF000) : a_params(3); // priority=30
-        const             (0x04, 0x0210 &&& 0x02F0) : a_params(4); // priority=40
-              priority=40 (0x04, 0x0010 &&& 0x02F0) : a_params(5);
-                          (0x06, _                ) : a_params(6); // priority=50
+        const priority=10: (0x01, 0x1111 &&& 0xF   ) : a_params(1);
+                           (0x02, 0x1181           ) : a_params(2); // priority=20
+                           (0x03, 0x1000 &&& 0xF000) : a_params(3); // priority=30
+        const              (0x04, 0x0210 &&& 0x02F0) : a_params(4); // priority=40
+              priority=40: (0x04, 0x0010 &&& 0x02F0) : a_params(5);
+                           (0x06, _                ) : a_params(6); // priority=50
     }
 }
 ~ End P4Example
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6973,8 +6973,7 @@ table property is not present at all, then the default convention is
 We also wish to support developers that want the convenience of
 predictable entry priority values automatically selected by the
 compiler, without having to write them in the program, plus the
-control to choose to specify entry priorities explicitly, if they
-wish.
+ability to specify entry priorities explicitly, if they wish.
 
 In some cases, developers may wish the initial priority values to have
 "gaps" between their values, to leave room for possible later
@@ -7049,6 +7048,10 @@ for (int j = 1; j < n; j += 1) {
 ~ End P4Pseudo
 
 This is the end of the first step: determining entry priorities.
+
+The priorities determined in this way are the values used when the P4
+program is first loaded into a device.  Afterwards, the priorities may
+only change by means provided by the control plane API in use.
 
 In the second step, the compiler issues errors for out of range
 priority values, and/or warnings for certain combinations of entry
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6888,7 +6888,7 @@ control ingress(inout Header_t h, inout Meta_t m,
                 inout standard_metadata_t standard_meta) {
 
     action a() { standard_meta.egress_spec = 0; }
-    action a_with_control_params(bit<9> x) { standard_meta.egress_spec = x; }
+    action a_params(bit<9> x) { standard_meta.egress_spec = x; }
 
     table t_exact_ternary {
 
@@ -6899,18 +6899,18 @@ control ingress(inout Header_t h, inout Meta_t m,
 
 	actions = {
             a;
-            a_with_control_params;
+            a_params;
         }
 
 	default_action = a;
 
         const entries = {
-            (0x01, 0x1111 &&& 0xF   ) : a_with_control_params(1);
-            (0x02, 0x1181           ) : a_with_control_params(2);
-            (0x03, 0x1111 &&& 0xF000) : a_with_control_params(3);
-            (0x04, 0x1211 &&& 0x02F0) : a_with_control_params(4);
-            (0x04, 0x1311 &&& 0x02F0) : a_with_control_params(5);
-            (0x06, _                ) : a_with_control_params(6);
+            (0x01, 0x1111 &&& 0xF   ) : a_params(1);
+            (0x02, 0x1181           ) : a_params(2);
+            (0x03, 0x1111 &&& 0xF000) : a_params(3);
+            (0x04, 0x1211 &&& 0x02F0) : a_params(4);
+            (0x04, 0x1311 &&& 0x02F0) : a_params(5);
+            (0x06, _                ) : a_params(6);
             _                         : a;
         }
     }
@@ -6919,7 +6919,7 @@ control ingress(inout Header_t h, inout Meta_t m,
 ~ End P4Example
 
 In this example we define a set of 7 entries, all of which invoke
-action `a_with_control_params` except for the final entry which
+action `a_params` except for the final entry which
 invokes action `a`. Once the program is loaded, these entries are
 installed in the table in the order they are enumerated in the
 program.
@@ -7090,6 +7090,47 @@ on the `entries` table property, then no warnings of this type are
 ever issued by the compiler for this table.  This option is provided
 for developers who explicitly choose to specify entries in an order
 that do not match their relative priority order.
+
+The following example is the same as the first example in section
+[#sec-entries], except for the definition of table `t_entry_ternary`
+shown below.
+
+~ Begin P4Example
+table t_exact_ternary {
+    key = {
+        h.h.e : exact;
+        h.h.t : ternary;
+    }
+
+    actions = {
+        a;
+        a_params;
+    }
+
+    default_action = a;
+
+    largest_priority_wins = false;
+    priority_delta = 10;
+    @noWarn("duplicate_priorities")
+    entries = {
+        const priority=10 (0x01, 0x1111 &&& 0xF   ) : a_params(1);
+                          (0x02, 0x1181           ) : a_params(2); // priority=20
+                          (0x03, 0x1000 &&& 0xF000) : a_params(3); // priority=30
+        const             (0x04, 0x0210 &&& 0x02F0) : a_params(4); // priority=40
+              priority=40 (0x04, 0x0010 &&& 0x02F0) : a_params(5);
+                          (0x06, _                ) : a_params(6); // priority=50
+    }
+}
+~ End P4Example
+
+The entries that do not have an explicit priority specified will be
+assigned the priority values shown in the comments, because
+`priority_delta` is 10, and because of those entries that do have
+priority values specified.
+
+Normally this program would cause a warning about multiple entries
+with the same priority of 40, but those warnings will be suppressed
+because of the `@noWarn("duplicate_priorities")` annotation.
 
 #### Size {#sec-size-table-property}
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6361,7 +6361,12 @@ sets" [P4SizeProperty] for further discussion on the implementation of parser
 value set size.
 
 The value set is populated by the control plane by methods specified in the
-P4Runtime specification.
+P4Runtime specification[^P4RuntimeAPI].
+
+[^P4RuntimeAPI]: The P4Runtime API is defined as a Google Protocol
+    Buffer `.proto` file and an accompanying English specification document
+    here:
+    <https://github.com/p4lang/p4runtime>
 
 # Control blocks { #sec-control }
 
@@ -6536,6 +6541,7 @@ tableProperty
     : KEY '=' '{' keyElementList '}'
     | ACTIONS '=' '{' actionList '}'
     | optAnnotations CONST ENTRIES '=' '{' entriesList '}' /* immutable entries */
+    | optAnnotations ENTRIES '=' '{' entriesList '}' /* mix of mutable and/or immutable entries */
     | optAnnotations CONST nonTableKwName '=' initializer ';'
     | optAnnotations nonTableKwName '=' initializer ';'
     ;
@@ -6767,12 +6773,21 @@ and illegal specifications of default actions for the `table t`:
 
 While table entries are typically installed by the control plane,
 tables may also be initialized at compile time with a set of entries.
-This is useful in situations where tables are
+
+Declaring these entries with `const entries`
+is useful in situations where tables are
 used to implement fixed algorithms---defining table entries statically
 enables expressing these algorithms directly in P4,
 which allows the compiler to infer how the table is actually used and
 potentially make better allocation decisions for targets with limited
-resources. Entries declared in the P4 source are installed in the
+resources.
+
+Declaring entries with `entries` (without the `const` qualifier)
+enables one to specify a mix of some immutable entries that are always
+in the table, and some mutable entries that the control plane is
+allowed to later change or remove.
+
+Entries declared in the P4 source are installed in the
 table when the program is loaded onto the target. Entries cannot be
 specified for a table with no key (see Sec. [#sec-table-keys]).
 
@@ -6780,7 +6795,8 @@ Table entries are defined using the following syntax:
 
 ~ Begin P4Grammar
 tableProperty
- : const ENTRIES '=' '{' entriesLlist '}' /* immutable entries */
+ : optAnnotations const ENTRIES '=' '{' entriesLlist '}' /* immutable entries */
+ | optAnnotations ENTRIES '=' '{' entriesList '}' /* mix of mutable and/or immutable entries */
 
 entriesList
  : /* empty */
@@ -6789,6 +6805,7 @@ entriesList
 
 entry
  : keysetExpression ':' actionRef optAnnotations ';'
+ | const keysetExpression ':' actionRef optAnnotations ';'
  ;
 ~ End P4Grammar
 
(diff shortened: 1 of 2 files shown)


jfingerh:
Intended to be a change to the spec that resolves issue https://github.com/p4lang/p4-spec/issues/1159, but note that it is not complete yet.  You are free to add review comments for the changes that have been made so far, but the section defining the options for initial entry priorities has not been written yet.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6885,6 +6908,40 @@ invokes action `a`. Once the program is loaded, these entries are
...
+interpreting numeric priority values.
+
+The P4Runtime API requires numeric priorities to be positive integers,
+and define that entries with larger priorities must win over
+entries with smaller priorities.  We will call this convention
+`largest_priority_wins`.
+
+TDI requires numeric priorities to be non-negative integers, and
+define that entries with smaller priorities must win over entries with
+larger priorities.
```
stolsma, line 6943: How is this convention called?? (the P4Runtime strategy is called 'largest-priority_wins', so I'm curious what is this called)

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6885,6 +6908,40 @@ invokes action `a`. Once the program is loaded, these entries are
...
+
+Unfortunately there are two commonly used, but different, ways of
+interpreting numeric priority values.
+
+The P4Runtime API requires numeric priorities to be positive integers,
+and define that entries with larger priorities must win over
+entries with smaller priorities.  We will call this convention
+`largest_priority_wins`.
+
+TDI requires numeric priorities to be non-negative integers, and
```
stolsma, line 6941: > non-negative integers

Also same question: Defined as?? Starting at and including 0 (zero) ?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6885,6 +6908,40 @@ invokes action `a`. Once the program is loaded, these entries are
...
+This enables the data plane to determine which of several matching
+entries is the "winner", i.e. the one entry whose action is invoked.
+
+[^TDI]: TDI is the Table Driven Interface.  More information can be
+    found here: https://github.com/p4lang/tdi
+
+Unfortunately there are two commonly used, but different, ways of
+interpreting numeric priority values.
+
+The P4Runtime API requires numeric priorities to be positive integers,
```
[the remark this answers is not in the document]
jfingerh: In mathematics, positive usually means strictly greater than 0, so starting at 1.  Non-negative means not less than 0, so starting at 0, but I'm happy to clarify both of these in the text if it will help.

p4-16/spec/P4-16-spec.mdk
jfingerh: Well largest_priority_wins is just the name I'm using in this document, to give a name to it.  I would call TDI smallest_priority_wins.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6885,6 +6908,41 @@ invokes action `a`. Once the program is loaded, these entries are
...
+
+If a table has fields where their `match_kind`s are all `exact` or
+`lpm`, there is no reason to assign numeric priorities to its entries.
+If they are all `exact`, duplicate keys are not allowed, and thus
+every lookup key can match at most one entry, so there is no need for
+a tiebreaker.  If there is an `lpm` field, the priority of the entry
+corresponds to the length of the prefix, i.e. if a lookup key matches
+multiple prefixes, the longest prefix is always the winner.
+
+For tables with other `match_kind` values, e.g. at least one `ternary`
```
mbudiu-vmw, line 6921: How do you specify the initial priorities?

jfingerh: To be described in the part of the PR that I have not yet written.  Probably another week or two.

jfingerh: @mbudiu-vmw I added another commit to this PR on 2022-Nov-15 with a description of how priorities are determined.

jfingerh:
@antoninbas @smolkaj @jnfoster @thantry @mariobaldi @mbudiu-vmw This is ready for review.  Corrections and comments (or approval) before the next P4 LDWG meeting on 2022-Dec-05 would be much appreciated, if you have time.  Otherwise it probably won't get into the spec for another month or two after that (if it gets into the spec).

Note that the per-entry `const` qualifier and per-entry initial priority values might require extensions to the data model of P4Runtime API, e..g perhaps adding an optional `const` field for each entry, and the auto-generated `.entries.txt` would need a way to specify that a table had initial entries without a global `const` before `entries`.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6849,7 +6888,7 @@ control ingress(inout Header_t h, inout Meta_t m,
                 inout standard_metadata_t standard_meta) {
 
     action a() { standard_meta.egress_spec = 0; }
-    action a_with_control_params(bit<9> x) { standard_meta.egress_spec = x; }
```
jafingerhut, line 6852: The only reason to change the name of `a_with_control_params` in this example is to make it shorter, so I can reuse most of this example in a new one, and have more page width to work with for adding priority values and comments.

jafingerhut:
@vgurevich @sayanb @slicking This is ready for review. Corrections and comments (or approval) before the next P4 LDWG meeting on 2022-Dec-05 would be much appreciated, if you have time. Otherwise it probably won't get into the spec for another month or two after that (if it gets into the spec).

See my previous comment on this PR, regarding new fields likely need for P4Runtime `.entries.txt` files.  This feature addition seems likely to require similar enhancements in TDI/bfrt.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
stolsma: @jfingerh thanks for the clarifying text! Really nice!! 👍 A long time  ago I went to a meteorology study and there the rule was everything upwards -0.0 is called a positive temperature (I know, the '-' is crazy) so from that moment on I always ask for precise definition of positive etc...

stolsma (approved):

jafingerhut:
@thantry @antoninbas @smolkaj @mariobaldi @vgurevich @sayanb @slicking  This is ready for review. Corrections and comments (or approval) before the next P4 LDWG meeting on 2022-Dec-05 (one week from today) would be much appreciated, if you have time. Otherwise it probably won't get into the spec for another month or two after that (if it gets into the spec).

Note that the per-entry const qualifier and per-entry initial priority values might require extensions to the data model of P4Runtime API, e..g perhaps adding an optional const field for each entry, and the auto-generated .entries.txt would need a way to specify that a table had initial entries without a global const before entries.

This feature seems likely to require similar enhancements in TDI/bfrt.

mariobaldi:
This looks good to me. I am in favor of approving this.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6769,55 +6782,81 @@ and illegal specifications of default actions for the `table t`:
...
  ;
 
+optCONST
+ : /* empty */
+ | const
+ ;
+
+optEntryPriority
+ : /* empty */
+ | priority '=' INTEGER
```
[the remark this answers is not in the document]
mbudiu-vmw: This will forbid using expressions here, e.g., constant values.

mbudiu-vmw (changes requested):
See a working implementation here: https://github.com/p4lang/p4c/pull/3748

mbudiu-vmw:
@jafingerhut please note that the grammar requires more extensive changes to allow "priority" to be used as a regular identifier.
(We should review the spec grammar again, it may have fallen behind the implementation.)

slicking (approved):
This looks good to me, thanks @jfingerh for working on this.  I remember requests to add additional entries at runtime to a table with initial entries has come up before.  So this will be a nice addition.  Looks very complete and I'm also glad to see the ability to mix const and non-const entries here.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6882,11 +6921,219 @@ control ingress(inout Header_t h, inout Meta_t m,
...
+
+    actions = {
+        a;
+        a_params;
+    }
+
+    default_action = a;
+
+    largest_priority_wins = false;
+    priority_delta = 10;
```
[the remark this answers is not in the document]
jfingerh: We do have a `@priority` annotation already implemented in p4c, yes.  The reason for the suggested syntax NOT using that here is the same general "prefer NOT to use annotations for things that can change the semantics of the program if they were removed" guidance that we often hear in the LDWG meetings when annotations are proposed.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6769,55 +6782,81 @@ and illegal specifications of default actions for the `table t`:
...
+before them, but this is redundant when the entries are declared using
+`const entries`.
+
+Table entries defined using `entries` (without a `const` qualifier
+before it) may have `const` before them, or not, independently for
+each entry.  Entries with `const` before them may not be modified or
+removed by the control plane.  Entries without `const` may be modified
+or removed by the control plane.  It is permitted for the control
+plane to add entries to such a table (subject to table capacity
+limitations), unlike tables declared with `const entries`.
```
mbudiu-vmw, line 6844: I don't know if it's worth specifying that having an entry in a table that is unreachable is possible, so the fact something can't be removed doesn't mean it will ever apply.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6882,11 +6921,219 @@ control ingress(inout Header_t h, inout Meta_t m,
...
 In this example we define a set of 7 entries, all of which invoke
-action `a_with_control_params` except for the final entry which
+action `a_params` except for the final entry which
 invokes action `a`. Once the program is loaded, these entries are
 installed in the table in the order they are enumerated in the
 program.
 
+##### Entry priorities { #sec-entry-priorities }
+
+If a table has fields where their `match_kind`s are all `exact` or
```
mbudiu-vmw, line 6931: I guess we have to merge https://github.com/p4lang/p4-spec/pull/1145 first

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
jafingerhut: There is a `@priority` annotation implemented in p4c since a 2017 commit to p4c by Calin Cascaval, but it is described nowhere in the language spec, and I believe never has.

Do you want me to deprecate something not mentioned in the spec, in the spec?

p4-16/spec/P4-16-spec.mdk
jafingerhut: By "unreachable" do you mean that it is impossible that a table.apply() call could ever find it as the "winning entry", because other higher priority entries mask it?

If so, this concept is never mentioned in the spec, and as far as I can see, need not be mentioned in the spec.

If an implementation wants to optimize away such entries, that seems to me a fine optimization that an implementation could make, but the spec doesn't need to mandate or forbid it.  The spec defines the observable behavior.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
mbudiu-vmw: Maybe you can find out if the barefoot backend uses them. We care about implementations even if they are not conformant with the spec.

p4-16/spec/P4-16-spec.mdk
jafingerhut: We could merge #1145 first, but I do not see any way in which this PR depends upon changes made in #1145.  Do you?

mbudiu-vmw: What each of these kinds does is described there

jafingerhut: I guess I would prefer if merging this PR was not blocked waiting for #1145, if #1145 is judged as not ready to merge for some reason, but this one is.  I think the text of this PR makes perfect sense even in the absence of #1145 being merged in, even if it might be more fully understood by someone first encountering what exact, lpm, and ternary mean after reading #1145's changes.  But the spec has lived for 6 years without #1145, so it doesn't seem like that should be blocking other enhancements now.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6769,55 +6782,81 @@ and illegal specifications of default actions for the `table t`:
...
+Entries declared in the P4 source are installed in the
 table when the program is loaded onto the target. Entries cannot be
 specified for a table with no key (see Sec. [#sec-table-keys]).
 
 Table entries are defined using the following syntax:
 
 ~ Begin P4Grammar
 tableProperty
- : const ENTRIES '=' '{' entriesLlist '}' /* immutable entries */
+ : optAnnotations const ENTRIES '=' '{' entriesLlist '}' /* immutable entries */
```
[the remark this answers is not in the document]
jafingerhut: I will compare your grammar changes in this p4c PR against what I wrote in this PR, and get back with updates to this PR, and/or questions about the differences in future comments.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
jafingerhut: Good point on the forbidding of expressions.  I will think about what grammar would enable expressions, but whatever expressions we allow, they should evaluate at compile time to integer constants.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
jafingerhut: Checking with @slicking @sayanb @vgurevich to see if they know.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
mbudiu-vmw: The implementation supports expressions, but they must be parenthesized to be unambiguous.
Similar to the `bit<(expression)>` case.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6769,55 +6782,81 @@ and illegal specifications of default actions for the `table t`:
...
+ ;
+
+optEntryPriority
+ : /* empty */
+ | priority '=' INTEGER
+ ;
+
 entry
- : keysetExpression ':' actionRef optAnnotations ';'
+ : optCONST optEntryPriority keysetExpression ':' actionRef optAnnotations ';'
```
apinski-cavium, line 6826: Hmm, an entry would look like
`const priority = 1     2 ... 3   :   a(t)    @opn;`
I don't know if like 1 and 2 being next to each other there without some kind of separator. 
The P4 language has separators for most things, having a separator removes the ambiguous issue too and why the need () for the expression case too.
I am ok with not having a separator but many folks might have to do a double take of looking at the code to understand it.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
jafingerhut: I checked with @slicking and the only uses he found of `@priority` in our code were test programs copied from public p4c.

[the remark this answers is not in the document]
mbudiu-vmw: In fact, this is what the compiler outputs if you use `@priority`:
```
table-entries-priority-bmv2.p4(79): [--Wwarn=deprecated] warning: The @priority annotation on Entry is not part of the P4 specification, nor of the P4Runtime specification, and will be ignored
            0x1181 &&& 0xF00F : a_with_control_params(3) @priority(1);
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

So I guess they are already deprecated.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
jafingerhut: With commit 9 I fixed priority->PRIORITY and also added the grammar rule to allow parenthesized expressions as priority values.  Thanks.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
jafingerhut: With commit 10 on this PR I have updated the grammar in the spec to match what p4c has, at least for the non-terminal symbol tableProperty

p4-16/spec/P4-16-spec.mdk
jafingerhut: I agree that does not look great.

Note that the syntax also allows one to write it more clearly as:
```
const priority = 1 (2 .. 3) : a(t) @opn;
```
We could recommend that people use the parentheses there for clarity, perhaps?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6882,11 +6920,219 @@ control ingress(inout Header_t h, inout Meta_t m,
...
+              priority=40 (0x04, 0x0010 &&& 0x02F0) : a_params(5);
+                          (0x06, _                ) : a_params(6); // priority=50
+    }
+}
+~ End P4Example
+
+The entries that do not have an explicit priority specified will be
+assigned the priority values shown in the comments, because
+`priority_delta` is 10, and because of those entries that do have
+priority values specified.
```
mbudiu-vmw, line 7130: these priorities do not change when the table is modified.

jfingerh: I added a short paragraph to the text explaining this.  Rather than saying that the priorities do not change when the table is modified, I left it up to the behavior defined by the control plane API in use.  This leaves open the possibility for a control plane API to define mechanisms to modify the priority values if entries, if this is deemed useful, without contradicting the language spec.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6769,55 +6780,82 @@ and illegal specifications of default actions for the `table t`:
...
+
+optEntryPriority
+ : /* empty */
+ | PRIORITY '=' INTEGER
+ | PRIORITY '=' '(' expression ')'
+ ;
+
 entry
- : keysetExpression ':' actionRef optAnnotations ';'
+ : optCONST optEntryPriority keysetExpression ':' actionRef optAnnotations ';'
```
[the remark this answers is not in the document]
jafingerhut: I agree we could.  I do not have a strong preference either way.  By what decision process should this syntax question be decided?

jafingerhut:
TODO for Andy when all other open questions about this PR have been resolved: Add changes to grammar.mdk file, and consider whether `priority` should be added to list of keywords or not.

apinski-cavium:
Overall I am ok with this change. I am not happy with the syntax around priorities but I won't block the change if other people are ok with the syntax.

jfingerh:
@apinksi-cavium Any alternate syntax(es) for specifying priorities that you might suggest?

apinski-cavium:
> @apinksi-cavium Any alternate syntax(es) for specifying priorities that you might suggest?

Just have a colon (':') after the priority if it is present. But as I said I don't want to bike shed the syntax here (https://en.wikipedia.org/wiki/Law_of_triviality ).

jfingerh:
Understood the desire not to bikeshed, but if we don't have comments here about it, the only other place is the LDWG meetings, and they are short enough that I don't want to wait until those meetings to resolve issues, if possible.

jnfoster:
If we are brainstorming syntax, how about using curly braces for an entry, with the priority before it?

```p4
...
  entries = [
    10 : { 1, a1() },   // priority 10
    { 2,  a2(99) },   // no priority given; compiler allocates 
    _ : { 3, a3() },   // priority explicitly elided, compiler allocates
    100 : { 4, a4() }, // priority 100
    200 : { 5 & 1, a5() } // priority 200
] 
```
I like the idea that an entry (match + action) is a unit of data that gets special syntactic treatment. I prefer having the priority outside of the entry. But one could also consider moving it inside...

As a further extension, one could also allow using optional labels if being extra explicit:
```p4
  100 : { key = 4, action = a4() }, 
```

And we could support both 
```
entries = { ... } // no order implied
```
and 
```
entries = [ ... ] // order implied
```
which matters when the compiler is allocating priorities.

This syntax is obviously not backwards compatible, but we could deprecate the earlier syntax for `const entries` in several versions. I don't think it's in widespread use, so the disruption should be minimal...

thantry:
I like Nate’s way of specification. It starts to look more like a hashtable with composite types.Otherwise, looks quite good to me.ThanksHariSent from my iPhoneOn Dec 28, 2022, at 9:40 PM, Nate Foster ***@***.***> wrote:﻿
If we are brainstorming syntax, how about using curly braces for an entry, with the priority before it?
...
  entries = [
    10 : { 1, a1() },   // priority 10
    { 2,  a2(99) },   // no priority given; compiler allocates 
    _ : { 3, a3() },   // priority explicitly elided, compiler allocates
    100 : { 4, a4() }, // priority 100
    200 : { 5 & 1, a5() } // priority 200
] 
I like the idea that an entry (match + action) is a unit of data that gets special syntactic treatment. I prefer having the priority outside of the entry. But one could also consider moving it inside...
As a further extension, one could also allow using optional labels if being extra explicit:
  100 : { key = 4, action = a4() }, 
And we could support both
entries = { ... } // no order implied

and
entries = [ ... ] // order implied

which matters when the compiler is allocating priorities.
This syntax is obviously not backwards compatible, but we could deprecate the earlier syntax for const entries in several versions. I don't think it's in widespread use, so the disruption should be minimal...

jafingerhut:
Hari wrote: "I like Nate’s way of specification. It starts to look more like a hashtable with composite types."

I have no objections to the syntax.  I would like to caution one about explaining it as a hash table, though.  Both the P4Runtime API and TDI control plane APIs explicitly support and allow there to be multiple entries in a table with the same priority as each other.

Background details on what it means if multiple entries have the same priority, and why it might be useful: Which one is the "winner" if more than one entry with the same numeric priority is not specified by the P4Runtime API.  Some control plane software only uses this capability in situations where all entries with the same priority are mutually exclusive, i.e. at most one of them can match any given search key value.  It can provide efficiency advantage when adding entries to such tables, in that fewer entries need to be moved when new entries are added.

mbudiu-vmw (approved):

mbudiu-vmw:
@jafingerhut will this be ready for discussion at the next LDWG meeting?

p4-16/spec/P4-16-spec.mdk
jafingerhut: #1145 is merged now, so resolving this comment.

jafingerhut:
Sorry, still no updates to this PR from me, but I would like to try to nail down the desired syntax, if possible.

This comment probably uses too many words to explain why I think we don't need to introduce two separate delimiters like square brackets vs. curly brackets for a list of table entries, and why curly brackets should be sufficient.

Gory details:

Today's `const entries = { .... }` syntax DOES imply an order to the entries in the `...` list.  See this paragraph in the current language spec:

```
If the runtime API requires a priority for the entries of a table—e.g. when using the P4 Runtime API, tables with at least one ternary search key field—then the entries are matched in program order, stopping at the first matching entry. Architectures should define the significance of entry order (if any) for other kinds of tables.
```

I mention this because of the proposal in comments above for a syntax using square brackets around a list of entries, e.g. `entries = [ ... ]`, with order required, vs. `{ ... }` with order not required.  It seems to me like this would be a backwards-incompatible change, and while we can certainly consider such a backwards-incompatible change, we should do it fully aware that we are doing so, and that it is worth doing so.

It is not clear to me that having a way to specify "order not required" is useful here.  Sure, there are tables where all key fields are exact match_kind only, and there the order is not important as long as all key values are distinct (and I believe we should require them to be distinct, or give an error message at compile time, in for such exact match tables).  But having only a syntax that means "source code of entries is significant" is easy to implement in that case: the compiler can ignore that order if it wants to.

jafingerhut:
Alternate syntax for initial entries has been proposed in this alternate PR.  At most one of this PR and that one should be merged: https://github.com/p4lang/p4-spec/pull/1231

jafingerhut:
@jnfoster @thantry I have tried to implement a grammar similar to what you propose in your comments above, but everything I have tried that looks similar to your example syntaxes leads to a grammar that causes shift/reduce conflicts in bison.  My work in progress version of it can be found on an alternate PR here: #1231 

Unless someone can find a bison-friendly grammar for an alternate syntax they like, that alternative syntax will not make any progress.

jafingerhut:
Commits 13 & 14 made on 2023-Apr-10 should make no changes to this PR's content, other than to merge it with the latest main branch version of the spec, which had some conflicts because of the use of `[INCLUDE=grammar.mdk:nonTerminalName]` Madoko constructs in the latest main version, which was merged into main after this PR was first written.

jafingerhut:
Current summary of this issue and possible proposed syntaxes for specifying the initial entries:

See slides 35 and 36 near the end of this slide deck: https://docs.google.com/presentation/d/1GW-9FDtyLuwqV_mdGeDh_qqDiey0VkQ1/edit#slide=id.g22d0e434be1_0_5

It shows examples of the syntaxes of the two PRs that exist for this issue, with links to those two PRs.

I am pinging people who have given comments on preferred syntaxes here.  I am personally happy with any of the ones mentioned in the slides linked above.  If someone wants to come up with another syntax that is not one of those, then it almost certainly will not happen in the P4 spec release in April 2023, but some later month after we discuss it again in a later LDWG meeting.

@mbudiu-vmw @jnfoster @thantry @apinski-cavium @vgurevich

thantry:
Approve! Thanks for the great work in pushing this through!

jnfoster (approved):

jnfoster (approved):

PR #1184, opened 2022-11-12
Add expression to represent an invalid header

jfingerh:


jafingerhut:
One possible change to the spec that would resolve this issue: https://github.com/p4lang/p4-spec/issues/1021

mbudiu-vmw (approved):
Similarly to the list PR, I suggest we accept this with a required type specified for the header, and we relax this once we know that type inference works in the compiler.

You need a changelog.

jfingerh:
Regarding changelogs, I will beg, repeatedly, to let me modify the change log in separate PRs from the changes.  I really, really dislike rebasing spec PRs solely because the change log needs rebasing, sorry.

jnfoster:
Fair. But as the person who creates the releases of the spec, let me beg all authors to include a ChangeLog. It's very painful for me to recreate them from the Git history because people forget to add them.

apinski-cavium:
Note I didn't notice an issue with `#` using outside of the preprocessor until I was trying to implement this in my front-end.
Take:
```
#define line }
header h{}
h f()
{
  return (h){
# line ;
}
```
Yes this is odd looking source (and one which you hope never see in the real life) but it shows ambiguity in the way the preprocessor and parser interact together. And is one of the reasons why `#` is not used outside of the preprocessor for C/C++.

mbudiu-vmw:
Thought about this, but this is the only ASCII special character that seemed reasonable.

apinski-cavium:
This is the simplest testcase which shows the issue with `#` on the front of the line:
```
header h{}
h f()
{
  return (h){
# };
}
```
I definitely could see this happening somehow on accident and the developer not understanding why there was an error message either.

jafingerhut:
I have added commit 3 to this PR that adds a note that `#` must not be the first non-white space character of a line, following Andrew Pinski's suggestion in an earlier comment (if we keep the currently proposed syntax using `{#}`).

jnfoster:
If the token is `{#}` then it cannot be the first character of the line...

Generally, tokens are indivisible. For example, the following is not legal P4:
```p4
co nt rol C() { 
 t
 a
 b
 l
 e
 t { k   e  y s = {} 
      a c t i o n s = {} 
  }
  ap ply { 
     t.a pp l y();
  } 
```

apinski-cavium:
> If the token is `{#}` then it cannot be the first character of the line...

Right but the current implementation inside p4c is three different tokens `{` `#` `}`
as seen by https://github.com/p4lang/p4c/pull/3667/files#diff-85fbe45d422156081940f1eb7e38cd4ff1bd03574da101de1cf7021a166851c9R187 .

I am ok with `{#}` being a token but that definitely requires some changes to p4c now.

jnfoster:
If `#` only ever has meaning when it appears immediately within a set of curly braces, then I think `{#}` should be a token...

This is a trivial front-end change. Even I could do it :-)

mbudiu-vmw:
Currently it is 3 tokens - spaces allowed. Having it be 1 token is a bit ugly, but would probably solve the problem.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4266,11 +4266,27 @@ h = { 10, 12 };  // This also makes the header h valid
...
+~ Begin P4Example
+header H { bit<32> x; bit<32> y; }
+H h;
+h = (H) {#};  // This makes the header h invalid
+if (h == (H) {#}) {  // This is equivalent to the condition !h.isValid()
+    // ...
+}
+~ End P4Example
+
+Note that the `#` character in this expression must not be the first
```
mbudiu-vmw, line 4286: the grammar should be inserted here.

mbudiu-vmw: An alternative grammar is H.invalid

mbudiu-vmw: Another alternative grammar is header.invalid

mbudiu-vmw: `S s = { ... } => S s = { (H){#} };`

mbudiu-vmw:
Yesterday during the LDWG discussion the idea was floated to use H.invalid for an invalid header.
This doesn't work so well when the header type is generic:
```
header H<T> {
   T field;
}
```
The syntax is more unwidely: `H<bit<32>>.invalid`.

apinski-cavium:
I just noticed one thing which is missing (it is a minor issue though).
The invalid header literal is not listed as a compile time known constant in the "Compile-time known values" section. I had assumed one of the reasons to do this was explicitly for a compile time known value for an invalid header.

jafingerhut:
@apinski-cavium I added commit 5 to this PR that adds `(H) {#}` to the list of compile-time known values.

apinski-cavium:
I see you also added about lists there; I am ok with both changes.

jafingerhut:
TODO for Andy when all other open questions about this PR have been resolved: Add changes to grammar.mdk file, if necessary.

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: H.invalid does not work so well for generic headers.

vgurevich: Can we have implicit casting of invalid headers added, similar how we do implicit casting of `int` ? I think that in both examples, `(H)` can be be easily figured out from the context. Not a biggie, but would be nice and I think that will eliminate the need for explicit casts in most of the cases.

mbudiu-vmw: Yes, we plan to do this in two steps. first require casting, then implement type inference and make casting optional.
We first want to implement fully the `...` syntax, though.
As always, I am leery putting something in the spec which isn't implemented yet.
The main question to discuss still is whether {#} is a single lexeme or 3.

apinski-cavium:
This can be addressed afterwards but how would be the best way to specify an header union without any valid headers? I would have assumed we could do the following also:
```
header H1{}
header H2{}
header_union HU { H1 h1; H2 h2; }
HU func() { return (HU){#}; }
```
but that is rejected currently:
```
headerunion-1.p4(4): [--Werror=type-error] error: cast: invalid header expression has a non-header type `header_union HU`
HU func() { return (HU){#}; }
                   ^^^^^^^
headerunion-1.p4(3)
header_union HU { H1 h1; H2 h2; }
             ^^
```

mbudiu-vmw:
One way to do it is to use any of the headers in the union. Haven't tried to see if it works, though.

mbudiu-vmw:
That probably won't work.
We'll have to define {#} for unions too.

mbudiu-vmw:
We decided to make `{#}` a single token.

mbudiu-vmw:
@jafingerhut I think only a tiny change to the grammar is needed: https://github.com/p4lang/p4c/pull/3835

p4-16/spec/P4-16-spec.mdk
jafingerhut: grammar inserted

mbudiu-vmw (approved):

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4118,11 +4118,34 @@ h = { 10, 12 };  // This also makes the header h valid
 h = { y = 12, x = 10 };  // Same effect as above
 ~ End P4Example
 
-Two headers can be compared for equality (==) or inequality (!=) only
+Two headers can be compared for equality (`==`) or inequality (`!=`) only
 if they have the same type.  Two headers are equal if and only if they
 are both invalid, or they are both valid and all their corresponding
 fields are equal.
 
+The expression `(H) {#}` represents an invalid header of type `H`,
```
jafingerhut, line 4126: Proposed replacement for this paragraph:

The expression `{#}` represents an invalid header.  In some contexts, the compiler may require
it to be written `(H) {#}` so that it need not infer the type of header `H` itself.

jafingerhut: Nate prefers that we make the same technical decision here as we do for structure-valued expressions, as to whether we call `(H)` before it a cast, vs. "some other thing that isn't a cast".

jafingerhut:
2023-Mar LDWG meeting Nate pointed out that this set of issues are very similar to each other, both in how they are specified and likely how they are implemented in p4c.  It would be nice if they had a consistent definition in terms of whether the `(typeRef)` before them was considered a cast or part of the syntax of the expression, and if it is not a required part of the expression, what is the type of the expression when there is no cast.  There may be other related questions raised in the LDWG discussion that I am not capturing here.

+ This issue
+ https://github.com/p4lang/p4-spec/pull/1220
+ Perhaps also this one?  Not sure: https://github.com/p4lang/p4-spec/pull/1218

@jnfoster If you have any thoughts on what you would like to see in the spec for these, please let us know.

jafingerhut:
@jnfoster Do you have a recommendation for what spec language you would like to see for this PR and the ones linked above in this comment: https://github.com/p4lang/p4-spec/pull/1184#issuecomment-1468638149

such that the resulting PRs would be as consistent as you would like them to be with each other, and the rest of the spec?

jnfoster:
I don't really care about the concrete syntax. My wish was to separate constructs that denote polymorphic literals like invalid headers from casts, and to do this consistently throughout the language.

jafingerhut:
I may have misunderstood some of the discussion in the March 2023 LDWG meeting.  I was under the impression that some of your concerns expressed there would require at least one, or all of:

+ some changes to this PR
+ some change to the PR introducing an expression for invalid header union values
+ perhaps even changes to other parts of the spec to clarify or change the syntax that "looks like a cast" for structure-values expressions, to clarify what that actually is that is not a cast. (I am unclear whether it technically _is_ a cast, or not.)

If this PR and the following ones are ready to be approved and merged in as they are written now, then I got confused in the last meeting that for some reason, they are not ready.

+ https://github.com/p4lang/p4-spec/pull/1220
+ Perhaps also this one? Not sure: https://github.com/p4lang/p4-spec/pull/1218

jnfoster:
The problem with #1220 and #1218 from my perspective is that they require a cast to appear along with the invalid header.

jafingerhut:
If we change those PRs so that they do not require a cast, then is it important to you if there is a defined type for the expression `{#}` by itself?  If so, should it be some new type that is "an invalid header or header union value of some type, but it could be any header type or header union type"?

jnfoster:
That's right. `{#}` would have any header type, just like `[]` in OCaml has any list type. 

Formally, we could model its type as a type variable `X <: header` where the notation means that `X` is constrained to have the `header` trait. That constraint then impacts type inference.

jfingerh:
I have created a nearly identical PR here https://github.com/p4lang/p4-spec/pull/1240 that attempts to follow the suggestion of making `{#}` an expression by itself, regardless of whether it has a cast on it.  In that other PR, I do NOT specify exactly when the compiler will or will not require a cast on the expression, only that the compiler might require one.

jfingerh:
@jnfoster If you find #1240 preferable to this PR, let me know, and I can consider updating #1220 and #1218 in a similar style.

jafingerhut:
Closing this, because the alternate PR #1240 was merged instead of this one.

PR #1190, opened 2022-11-26
Add definition of P4Example Madoko 'macro' to PSA spec, and use it

jfingerh:
This brings it closer to the P4_16 and PNA specifications.


jafingerhut (approved):
This is not a change to the human-readable part of the PSA specification, so approving it myself and merging it in.  It is only a change to the Madoko source code.

PR #1198, opened 2022-12-06
Update list of Ubuntu versions supported by install script

jfingerh:


jfingerh:
There are no spec changes here, only updates to some slightly outdated list of supported Ubuntu Linux versions on which the bash install script is known to work.

PR #1200, opened 2022-12-08
Add scripts to compare spec grammar.mdk file against p4c Bison parser

jfingerh:


jfingerh:
With the recent changes to the grammar.mdk file in the p4-spec repo, running these scripts mechanically gets to versions of grammar.mdk and p4parser.ypp from the p4lang/p4c repo that are under 50 "hunks" of output in the diffs.

With a small enhancement that I might still make to the Python program in this PR, I can probably get it to under 30 (by putting every alternative of a grammar rule onto a single line, always).

jfingerh:
As of commit 4 on this PR, I am done modifying it now.  If you run it with latest p4c and p4-spec repos, there are only 27 "hunks" of difference between the two grammars when trimmed of comments and lines beginning with `%`.

p4-16/spec/scripts/trim-p4-grammar-file.py
```diff
@@ -0,0 +1,136 @@
+#! /usr/bin/env python3
```
mbudiu-vmw, line 1: the robust way is to use a real parser to get a representation of the grammar.
the bnf language grammar isn't that complicated.
I have used the parglare parser generator which is written and generates Python, it's relatively easy to use.
http://www.igordejanovic.net/parglare/0.16.0/

jfingerh: Anyone else is welcome to create a robust solution if they want.  This code is tested and effective with the current grammar.

jafingerhut: Do you think it is reasonable to merge this in as is without support for general Bison grammars?  These programs work with the current grammar, and any likely changes I ever expect to see in the P4 grammar files, but I can look into fixing them if changes are made to these grammars in the future that break these programs (that seems unlikely to me, though).

PR #1201, opened 2022-12-10
Use Madoko INCLUDE directive for grammar excerpts within the spec

jfingerh:


jfingerh:
Note: This PR intentionally only uses the Madoko INCLUDE directive for 4 or 5 of the grammar excerpts in the specification.  Many more could be similarly done like this, but this PR was created quickly merely as an example of what could be done throughout the specification, so that spec maintainers can see what the resulting Madoko file would look like.

jfingerh:
Is there a strong reason not to show the full production, in some cases?  Is it primarily for focusing on a few of the production rules being presented in that section of the spec?

mbudiu-vmw:
That's right

jfingerh:
After 3 commits on this PR, I have replaced all occurrences of grammar rule definitions in the file P4-16-spec.mdk with Madoko INCLUDE directives, but ONLY for the ones where the definition in the file grammar.mdk was functionally identical to the one that resulted from replacing it with the INCLUDE statement.  Usually this was also textually identical, but in a few cases the order of some of the alternatives was different, but I left them in the order they were in the grammar.mdk file, because with recent changes made there, those orders match the order of corresponding grammar rules in p4c, for easier text-based comparison between p4c and grammar.mdk.

There are only about 10 or so grammar rules left as they were in P4-16-spec.mdk.  Those are either the ones that were partial rules, or were in some other way not functionally identical with the ones in grammar.mdk.  Those we could consider whether they should become identical between them, but not in this PR.

p4-16/spec/grammar.mdk
```diff
@@ -1,10 +1,12 @@
-~ Begin P4Grammar
+// BEGIN:p4program
```
mbudiu-vmw, line 1: maybe you could have done this using a script, but now it's done

p4-16/spec/grammar.mdk
```diff
@@ -27,47 +31,60 @@ nonTypeName
...
     ;
+// END:annotation
 
+// BEGIN:annotationBody
 annotationBody
     : /* empty */
     | annotationBody "(" annotationBody ")"
     | annotationBody annotationToken
     ;
+// END:annotationBody
```
mbudiu-vmw, line 87: how come annotationToken below doesn't need a comment?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -1111,11 +1111,7 @@ excerpts from the grammar:
 - `UPPERCASE` symbols denote terminals in the grammar.
 - Excerpts from the grammar are given in BNF notation as follows:
 ~ Begin P4Grammar
-p4program
-    : /* empty */
-    | p4program declaration
-    | p4program ';'
-    ;
+[INCLUDE=grammar.mdk:p4program]
```
mbudiu-vmw, line 1114: this looks great.

mbudiu-vmw: I assume that madoko will complain if the link is broken

mbudiu-vmw (approved):

p4-16/spec/grammar.mdk
jafingerhut: Given the process that I used to make these changes, if a rule for a nonterminal does no have BEGIN/END comments around it, it is because it was never used in an INCLUDE statement added to the file P4-16-spec.mdk.  Why not?  Two possible reasons that would not be a mistake on my part: (a) the nonterminal is not defined in P4-16-spec.mdk anywhere, or (b) the nonterminal is defined in P4-16-spec.mdk, but in the places where it is, its right-hand side is not identical to what is in grammar.mdk, and I did not want to decide in this PR which of those should be considered correct.

p4-16/spec/P4-16-spec.mdk
jfingerh: I just ran a local test on my system (not part of this PR) where I changed the INCLUDE line above to have a name that was not defined in the grammar.mdk file: `[INCLUDE=grammar.mdk:p4progra]` (I deleted the final character of `p4program` to introduce a mistake.  Madoko produces a warning near the beginning of its output, as shown below on the Ubuntu 20.04 system where I tested this:
```
$ make
madoko --pdf -vv --png --odir=build P4-16-spec.mdk
process: P4-16-spec.mdk -> build/P4-16-spec.html
  warning: P4-16-spec.mdk:1114: unknown snippet name: p4progra

```

p4-16/spec/grammar.mdk
jfingerh: I just double-checked: there is no definition for `annotationToken` in the P4-16-spec.mdk file that I started making the changes in this PR.  There IS a mention of `annotationToken` in the text before `annotationBody`'s definition is shown, but it appears `annotationToken` either never made it into the grammar definition in the body of the spec there, or someone deleted it at some point in the past.

jfingerh:
I reviewed one more time the grammar rules in P4-16-spec.mdk that had NOT yet been replaced with Madoko INCLUDE directives, and found 3 more that were equivalent between the two, and thus also replaced them with INCLUDE directives.

There remain only 3 non-terminal definitions in P4-16-spec.mdk that are not identical to the corresponding definition in grammar.mdk, and also do not appear to be excerpts.  I will file a separate issue on the spec to record those, in case someone wants to think about how to address them.

In the mean time, since this PR makes no functional changes to the spec, only in how the Madoko source looks in producing it, I will go ahead and merge this in with Mihai's one approval.

PR #1210, opened 2022-12-22
Release version 1.2 of PSA specification

jfingerh:


jafingerhut (approved):
No changes except adding one item to revision history, and marking version as 1.2

PR #1218, opened 2023-01-10
Added header stack expressions

mihaibudiu:

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2929,7 +2929,7 @@ package types.
 
 ## Numeric types { #sec-numeric-values }
 
-Many P4 operations are restrained to expressions that evaluate to numeric values. 
+Many P4 operations are restrained to expressions that evaluate to numeric values.
 Such expressions must have one of the following numeric types:
 
   - `int` - an arbitrary-precision integer (section
@@ -4032,7 +4032,7 @@ compile-time (Section [#sec-minsizeinbits]).
 
 ## Operations on sets { #sec-set-exprs }
 
-Some P4 expressions denote sets of values (`set<T>`, for some type `T`; 
+Some P4 expressions denote sets of values (`set<T>`, for some type `T`;
 see Section [#sec-set-types]). These expressions can
 appear only in a few contexts---parsers and table
 entries. For example, the `select` expression (Section
@@ -4093,7 +4093,7 @@ select (hdr.ipv4.version) {
 ### Masks { #sec-cubes }
 
 The infix operator `&&&` takes two arguments of the same numeric type (Section [#sec-numeric-values]),
-and creates a value of the same type. 
+and creates a value of the same type.
 The right value is used as a
 "mask", where each bit set to `0` in the mask indicates a "don't care"
 bit. More formally, the set denoted by `a &&& b` is defined as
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
Implemented in https://github.com/p4lang/p4c/pull/3803

mbudiu-vmw:
Discussed during January 2023 LDWG

mbudiu-vmw:
Note: this requires no grammar changes.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4263,6 +4263,24 @@ only if they have the same element type and the same length.  Two
...
 equal.  Note that the `nextIndex` value is not used in the equality comparison.
 
+### Header stack expressions { #sec-hs-init }
+
+An expression of the following form evaluates to a header stack: '('
+`typeRef` ')' "{" expressionList "}", where typeRef is a header stack
+type name.  Each expression in the list must evaluate to a header of
+the appropriate type.  The number of headers in the list must equal
+the stack size.
+
```
vgurevich, line 4273: The example below greatly clarifies the statement above. Here are some editorial thoughts:
1. We state that `typeRef` is " header stack type *name*`, which might be misconstrued that we do not allow complex typeRefs in here. Perhaps it makes sense to remove the word "name"?
2. In the next sentence it is unclear what "the appropriate type" is. Do we have a term for the "base type" of the header stack (i.e. the type of the individual element"? 
  * Note also that in the example below we require explicit cast for the header stack expression and the invalid header, but not for the headers (of that "appropriate type"). Maybe it would make sense to add that we do support implicit casting in this context?
  * Hopefully, `...` is on the way and will eliminate the need for that explicit invalid header.
3. I would suggest replacing "must equal" with "must be equal to"

vgurevich (changes requested):

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4376,6 +4376,32 @@ only if they have the same element type and the same length.  Two
 stacks are equal if and only if all their corresponding elements are
 equal.  Note that the `nextIndex` value is not used in the equality comparison.
 
+### Header stack expressions { #sec-hs-init }
+
+An expression of the following form evaluates to a header stack: '('
```
mbudiu-vmw, line 4381: Should rephrase this in the same way as header initializers.
Add the cast to stacks as a legal cast.

mbudiu-vmw:
I have added in this PR some of the changes requested for #1250 
If we merge both PRs we should be fine

jnfoster:
Hard to review with conflicts still...

mbudiu-vmw:
Conflicts because I merged two other things. I will rebase again

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4394,6 +4396,39 @@ only if they have the same element type and the same length.  Two
 stacks are equal if and only if all their corresponding elements are
 equal.  Note that the `nextIndex` value is not used in the equality comparison.
 
+### Header stack expressions { #sec-hs-init }
+
+One can write expressions that evaluate to a header stack.  The
+syntax of these expressions is given by:
+
+~ Begin P4Grammar
+expression ...
```
jafingerhut, line 4405: Minor nit that we can try to make more consistent with this PR, or a later one:

There are some cases where we do partial lists of grammar production rules like this:
```
expression
    : ...
    | expression '<' realTypeArgumentList '>' '(' argumentList ')'
    | expression '(' argumentList ')'
```
and others like `expression ...` as shown here, and one other variant I saw in another place.  I am happy to create a separate PR and make these all consistent, since that is not the main point of this PR.

jafingerhut (approved):
LGTM

p4-16/spec/P4-16-spec.mdk
jnfoster: This is now addressed.

jnfoster (approved):

PR #1220, opened 2023-01-13
Introduce invalid header union expressions

### LINKED ISSUE #1215: Add expression to represent an invalid header_union value
This issue is identical in motivation as this earlier issue and PR: https://github.com/p4lang/p4-spec/pull/1184 except that this issue is to ensure that invalid headers can be written with a header_union type.

mbudiu-vmw:
Fixes #1215

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2416,6 +2416,8 @@ Mpls_h[10] mpls;
 introduce a header stack called `mpls` containing ten entries, each
 of type `Mpls_h`.
 
+Operations on header stacks are described in Section [#sec-expr-hs].
+
 ### Header unions {#sec-header-unions}
 
 A header union represents an alternative containing at most one of several different
@@ -2445,6 +2447,8 @@ header_union IP_h {
 ~ End P4Example
 
 A header union is not considered a type with fixed length.
+
+Operation on header unions are described in Section [#sec-expr-hu].
 
 ### Struct types { #sec-struct-types }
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
It is unclear to me that there is any semantic change to the specification intended by these changes to the text.  It appears that it might simply be different words to mean the same thing.

Are there any semantic changes intended by this PR?  If so, what are they?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4465,7 +4472,15 @@ parser Tcp_option_parser(packet_in b, out Tcp_option_stack vec) {
 }
 ~End P4Example
 
-Two header unions can be compared for equality (==) or inequality (!=)
+An expression of the form '(' `type` ')' '{#}', where `type` is a header union type,
+denotes an invalid header union with the specified type.
+
+~ Begin P4Example
+header_union HU { ... }
+HU h = (HU){#};  // invalid header union; same as an uninitialized header union.
```
jafingerhut, line 4480: This is a fairly nit-pick comment, but this assignment is only the "same as an uninitialized header union" if that header union is defined as an `out` direction parameter of the enclosing parser or control, but it is only the same in that case BECAUSE the spec defines all such `out` headers and header_unions to be initialized to invalid.

If you declared a local variable of type header or header_union, I believe the spec defines NOTHING about the initial value.  That is, for such a local varriable, even h.isValid() would return an unspecified true or false value, whatever an implementation decides to do.

jafingerhut: I would thus recommend simply removing "same as an uninitialized header union" from the comment, leaving it to be only "invalid header union" and nothing else.

jafingerhut: Hmmm.   On a re-read of the spec regarding header_union, I now notice this statement at the beginning of the section "Operations on header unions" that I do not recall noticing before, but it has been there since 2017:

```
A variable declared with a union type is initially invalid.
```
That appears to be _different_ than the initialization behavior for normal headers, but I have no suggestion of changing it now.

Please disregard my earlier comments that are at odds with the language spec earlier on this point.

mbudiu-vmw:
> It is unclear to me that there is any semantic change to the specification intended by these changes to the text. It appears that it might simply be different words to mean the same thing.
> 
> Are there any semantic changes intended by this PR? If so, what are they?
> 
> I should clarify: I do see the new definition for syntax that represents the value of an invalid header_union. That is new, I agree.
> 
> It is the earlier parts of the changes in this PR that I am asking about if there are any intended semantic changes.

The intent is to clarify the spec, not to change it. There is also an enumeration which seems incomplete.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -7413,6 +7430,8 @@ The following are compile-time known values:
 - The `_` identifier when used as a `select` expression label
 - The expression `{#}` representing an invalid header or header union
   value.
+- The expression `(HU){#}` representing an invalid header union of type HU.
+- The expression `(H){#}` representing an header of type H.
```
mbudiu-vmw, line 7434: Move casts from the list below into a separate bullet and delete the above 2.

jafingerhut (approved):
LGTM

mbudiu-vmw:
As discussed during the LDWG meeting, I will merge this.

PR #1222, opened 2023-02-02
Clarify constraints on parameters with default values

### LINKED ISSUE #1097: Default values for parameter resolution is not well specificied
Take:
```
extern void f(in bit a, in bit<3> b = 2, in bit<5> c);

void g()
{
  f(1, 1);
}
```
The current specifications does not mention the extern call f will not work when used with not naming the arguments in the call.

The following could be added to solve this issue:
```
When not using named arguments, only defaulted arguments at the end are takened into account.
```

mbudiu-vmw:
Fixes #1097

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1661,12 +1661,10 @@ directions:
   See Section [#sec-invoke-actions] for further details.
 - Default parameter values are only allowed for 'in' or direction-less parameters;
   these values must evaluate to compile-time constants.
-- If a callable object (action, control, parser, extern method, etc.)
-  has a signature containing default values for parameters, and if the
-  parameters containing default values do *not*  all appear at the end of
-  the list of parameters, then the object can be invoked only by
-  using named arguments, to make the call easy to disambiguate.
-  For example:
+- If parameters with default values do not appear at the
+  end of the list of parameters, invocations that use
+  the default values must use named arguments, as in the following
+  example:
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -1661,6 +1661,24 @@ directions:
   See Section [#sec-invoke-actions] for further details.
 - Default parameter values are only allowed for 'in' or direction-less parameters;
   these values must evaluate to compile-time constants.
+- If a callable object (action, control, parser, extern method, etc.)
+  has a signature containing default values for parameters, and if the
+  parameters containing default values do *not*  all appear at the end of
+  the list of parameters, then the object can be invoked only by
+  using named arguments, to make the call easy to disambiguate.
```
apinski-cavium, line 1668: This wording is confusing I think.
Especially when the example below shows an example of where not using named arguments is still allowed.
I don't have better wording on this but we definitely should combine the example below even with the example under @optional section below this too. Or rather move the example of default arguments only to this section from that section too.

mbudiu-vmw:
@apinski-cavium is this better?

mbudiu-vmw:
Let me make another section

apinski-cavium:
> @apinski-cavium is this better?

Yes that is much easier to understand. So +1 from me now.

rst0git (approved):
LGTM

jafingerhut (approved):
LGTM

PR #1230, opened 2023-03-03
Add section on operations on extern objects

### LINKED ISSUE #1225: There is no section on "operations on extern objects"
Do we need one in the spec?
The operations are instantiations and method calls.

mbudiu-vmw:
Fixes #1225

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4640,7 +4640,7 @@ following:
 
 Controls, parsers, packages, actions, and extern constructors can have parameters of type
 `extern` only if they are directionless parameters.  Extern object
-instances can thus used as arguments in the construction of such
+instances can thus be used as arguments in the construction of such
 objects.  For parsers and controls only constructor parameters can
 have `extern` types.
 
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -4638,11 +4638,10 @@ following:
 - Invoking a method of an extern object instance using a method call expression, as
   described in Section [#sec-invocations].
 
-Controls, parsers, packages, actions, and extern constructors can have parameters of type
-`extern` only if they are directionless parameters.  Extern object
-instances can thus be used as arguments in the construction of such
-objects.  For parsers and controls only constructor parameters can
-have `extern` types.
+Controls, parsers, packages, actions, and extern constructors can have
+parameters of type `extern` only if they are directionless parameters.
+Extern object instances can thus be used as arguments in the
+construction of such objects.
 
 No other operations are available on extern objects, e.g., equality
 comparison is not defined.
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -4566,6 +4566,25 @@ table tbl {
...
+following:
+
+- Instantiating an extern object using a constructor invocation, as described in
+  Section [#sec-constructor].
+- Invoking a method of an extern object instance using a method call expression, as
+  described in Section [#sec-invocations].
+
+Controls, parsers, packages, actions, and extern constructors can have parameters of type
+`extern` only if they are directionless parameters.  Extern object
+instances can thus used as arguments in the construction of such
```
jafingerhut, line 4581: typo?  "thus used" -> "thus be used" ?

jafingerhut (approved):
LGTM except for the one typo I mention in a comment.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4566,6 +4566,25 @@ table tbl {
...
+
+- Instantiating an extern object using a constructor invocation, as described in
+  Section [#sec-constructor].
+- Invoking a method of an extern object instance using a method call expression, as
+  described in Section [#sec-invocations].
+
+Controls, parsers, packages, actions, and extern constructors can have parameters of type
+`extern` only if they are directionless parameters.  Extern object
+instances can thus be used as arguments in the construction of such
+objects.  For parsers and controls only constructor parameters can
```
mbudiu-vmw, line 4582: This is not true for parsers and packet_in

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4657,6 +4657,23 @@ table tbl {
...
+
+The only operations that can be performed on `extern` objects are the
+following:
+
+- Instantiating an extern object using a constructor invocation, as described in
+  Section [#sec-constructor].
+- Invoking a method of an extern object instance using a method call expression, as
+  described in Section [#sec-invocations].
+
+Controls, parsers, packages, actions, and extern constructors can have
```
jfingerh, line 4670: There was some discussion during the 2023-Apr-10 LDWG meeting asking whether `actions` should be removed from this sentence.

After the meeting, I was reminded of the following sentence in Section 14.1 "Actions" that has been in the spec since before version 1.0 was published: "Action parameters may not have extern types. "

I traced it back as far as 2016-Dec-16 before I stopped going back in time looking further.

We can generalize this, of course, but it seems like it could use more thought on what it means, how it is defined, any restrictions on extern instances as action parameters, etc. before adding it.

vgurevich: I agree -- for now the best thing would be to remove the word "actions" from the line 4670. Allowing externs to be action parameters opens a whole can of worms and while I can see how this might be useful, a closer look reveals many problems. Let's continue with the current approach (externs cannot be action parameters) -- we can always allow them later.

jnfoster: Ahh, thanks.

PR #1231, opened 2023-03-06
Add definition of initial table entries, without const qualifier

jfingerh:
I am creating a separate PR from my original one here: https://github.com/p4lang/p4-spec/pull/1180

for two reasons:
(1) In case the original one ends up being preferred for some reason, I won't have to revert changes to #1180.
(2) This one is based upon the latest main version of the spec as of 2023-Mar-06, so should have no merge conflicts, at least for changes to the spec up to that date.

The first commit on this PR is textually and semantically identical to the version on commit number 12 of #1180, with the minor difference that the new and updated grammar rules are in the grammar.mdk file, and in the P4-16-spec.mdk file we use the new Madoko "INCLUDE" style.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -6539,9 +6539,19 @@ tableProperty
 
 [INCLUDE=grammar.mdk:optCONST]
 
+[INCLUDE=grammar.mdk:optPriority]
+
 [INCLUDE=grammar.mdk:optEntryPriority]
 
 [INCLUDE=grammar.mdk:entry]
+
+[INCLUDE=grammar.mdk:entryElementList]
+
+[INCLUDE=grammar.mdk:entryElement]
+
+[INCLUDE=grammar.mdk:optKey]
+
+[INCLUDE=grammar.mdk:optAction]
 ~ End P4Grammar
 
 Table entries defined using `const entries` are immutable---i.e., they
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
If you want to have this discussed in the LDWG you can add it to the agenda.

jafingerhut:
Note: This alternate syntax has not been implemented or checked carefully for ambiguity yet.

jafingerhut:
As far as I can tell, this grammar has at least shift/reduce conflicts, or other issues that Bison cannot handle.

One possible way to keep backwards compatibility _and_ introduce this new feature is to pick a new table property with a name other than `entries`.

jnfoster:
Will you attach the `.y` file so interested parties can fiddle?

jafingerhut:
Attached is a zip file with a recent version of the bison parser from from the p4lang/p4c repo, and my attempts to modify it in a way similar to the grammar proposed in this PR, but not exactly the same as this PR because I was already starting to try to modify the grammar proposed in commit #2 of this PR to avoid shift/reduce conflicts, but there are still some even with those changes.

[p4c-bison-parsers.zip](https://github.com/p4lang/p4-spec/files/10910368/p4c-bison-parsers.zip)

jafingerhut:
Attached is one modification to the current p4c bison parser that has no shift/reduce conflicts.  It has optional `const` at the beginning of each entry, and an optional `priority = INTEGER :` after that, followed by the current `keysetExpression ":" actionRef`.

Trying to make the `priority` keyword followed by `=` sign there seems to introduce a shift/reduce conflict that I don't know how to avoid.

[p4parser.ex5.ypp.txt](https://github.com/p4lang/p4-spec/files/10914077/p4parser.ex5.ypp.txt)

jafingerhut:
@thantry @jnfoster With commit 4 on this PR, I have a grammar that bison is happy with.  It achieves this by making NO changes to the definition of the existing `const entries` table property syntax at all.  Instead, it keeps that as defined now, and introduces a new table property named `initial_entries`, with a different syntax for specifying its entries, with optional priorities.  Please take a look at this PR to see if it meets with your approval.

Except for the syntax change in `initial_entries`, this should be semantically the same as this older PR that many people have reviewed and found satisfactory: https://github.com/p4lang/p4-spec/pull/1180

jnfoster:
Yes, LGTM too. We should have a LDWG discussion, but let's plan to get this in for the next release!

p4-16/spec/P4-16-spec.mdk
```diff
@@ -6607,11 +6629,269 @@ control ingress(inout Header_t h, inout Meta_t m,
...
+        a_params;
+    }
+
+    default_action = a;
+
+    largest_priority_wins = false;
+    priority_delta = 10;
+    @noWarn("duplicate_priorities")
+    initial_entries = {
+        const 10: { (0x01, 0x1111 &&& 0xF   ), a_params(1) };
```
mbudiu-vmw, line 6876: I find this syntax hard to read. I really liked the explicit "priority"

jafingerhut: The grammar allows an optional `priority=` before the number, if you want, but does not require it.

I'm at the point where as soon as we find a syntax that bison is happy with and most people reviewing it are happy with, I'm good to go :-)

mbudiu-vmw:
@jafingerhut is this ready for re-review?

jfingerh:
@mbudiu-vmw As of commit 4 of this PR (the latest one as I write this), it is ready for review, and has already been reviewed and approved by Nate Foster and Hari Thantry.  You have added one comment to it since that time, which I replied to in another comment, but I have made no changes since then.

jafingerhut:
In the 2023-May-01 LDWG meeting, this PR was discussed versus the other proposed syntax in https://github.com/p4lang/p4-spec/pull/1180.  The consensus was that #1180 should be preferred over this one.

jafingerhut:
Closing this since an alternate PR is preferred over this one.

Issue #1223, opened 2023-02-06
Loose description of implicit cast


pataei:
# Summary
Implicit cast insertion is partially specified in the spec but many of the details are left up to implementers. Type equality and type inference have received similar treatments. This means that two different compilers may insert implicit casts differently, which will affect the portability of programs between different P4 type checkers. The spec should avoid this.

P4 spec handles implicit cast haphazardly which takes the following shapes throughout the spec:
* Instead of spelling out where and how implicit cast occurs, it just hints at it. 
* It doesn’t use consistent language to describe implicit cast. For example, sometimes it uses (automatic) conversion instead of implicit cast. 
* The description of the implicit cast is very limited, it only discusses implicit cast when binary operations are used or when it comes to enums. However, these are absolutely not the only cases where implicit cast occurs. 

Addressing these issues will require addressing similar issues with type inference and [type equality](https://github.com/p4lang/p4-spec/issues/875), which are related to implicit casts. Inference may solve constraints between types and type variables by inserting casts. Type equality comparisons are often (but not always!) meant to check type equality modulo implicit casts.

## Examples 
We provide various examples for each of the problems.
### Not spelling out implicit cast when it occurs
* [Assignment](https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-assignment). P4 spec states: "An assignment, written with the = sign, first evaluates its left sub-expression to an l-value, then evaluates its right sub-expression to a value, and finally copies the value into the l-value. Derived types (e.g. structs) are copied recursively, and all components of headers are copied, including “validity” bits. Assignment is not defined for extern values."
This explanation doesn't say anything about inserting an implicit cast on the RHS to match the type of the LHS of an assignment. However, P4 spec has lots of examples in various subsections of needing to cast the RHS of an assignment. For example, in [Section 8.12](https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-list-exprs) P4 spec states that "A list may be used to initialize a structure if the list has the same number of elements as fields in the structure. The effect of such an initializer is to assign the nth element of the list to the nth field in the structure:" Here, for the assignment to type check, an implicit cast from list to struct has to be inserted. 
```
struct S {
    bit<32> a;
    bit<32> b;
}
const S x = { 10, 20 }; //a = 10, b = 20
```
[...]

PR #1232, opened 2023-03-06
Fix casting

pataei:
This PR addresses some of the issues mentioned in #1223. Specifically, it addresses the following:

- Uses consistent terminology to talk about the implicit cast
- Expands explicit cast enumeration
- Adds comments in some examples where implicit cast occurs
- Adds the recursiveness of casting

Some questions arose while formalizing some of P4 which are related to this PR:
- [x] Should we add record types (or anonymous header types) so that we can explicitly add them to the casting section, instead of dancing around it every time we use record types to initialize a struct or header? Discussed this in the LDWG meeting on March 6th. Not sure if there's a consensus. Suggestion: propose the change in a PR. 
- [x] Other than the potential casting of the subtypes of a list, can a list be casted to another type or be the result of a cast? No, it seems that the value of a list can only be written using an explicit cast expression, however, I don't think spec defines it as such. It seems that that's just part of the syntax.
- [x] Are there any restrictions for the nesting of generic types? No, this is checked when types are instantiated.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3689,9 +3689,11 @@ The following casts are legal in P4:
   loss, and then truncates the result to `W` bits. The compiler should
   emit a warning on overflow.
 - casts between two types that are introduced by `typedef` and are equivalent to one of the above combinations.
-- casts between a typedef and the original type.
-- casts between a type introduced by `type` and the original type.
-- casts between an `enum` with an explicit type and its underlying type
+- casts from a typedef to the original type and vice versa.
+- casts from a type introduced by `type` to the original type and vice versa.
+- casts from an `enum` with an explicit type to its underlying type and vice versa.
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -1797,7 +1797,7 @@ restricted places in P4 programs.
 
 ### The error type
 
-The error type contains opaque values that can be used to signal
+The error type contains opaque distinct values that can be used to signal
```
jnfoster, line 1800: This might be an improvement to the spec, but it seems unrelated to casting?

pataei: Yes, it's not related to casting. I just fixed little things that came up as I was working on the formalization.

pataei:
I'll break this down into multiple PRs and then submit them.

jnfoster:
Thanks. That'll be much easier to review.

Issue #1105, opened 2022-05-22
header stacks of an header union


apinski-cavium:
I noticed that p4c accepts header stacks of header unions (via the testcase in https://github.com/p4lang/p4c/pull/3342) even though the specification seems to imply they are not valid.

The current language in the specifications is:
```
where `typeName` is the name of a header type.
```
It would be nice if `or header union type` was added there.

apinski-cavium:
https://github.com/p4lang/p4c/pull/3121

apinski-cavium:
Dup of issue #876 in the end.

Issue #876, opened 2020-06-24
Legal to form header stacks of header unions?


hackedy:
I'm having trouble telling if header stacks `H[n]` require `H` to be a header only or if they also allow header unions. The p4c typechecker allows unions, and the original proposal #139 said unions were ok. Maybe this should be an additional column in the type nesting table?

jafingerhut:
I have thought that the intent was that header stacks where the elements are header unions were intended to be supported in the spec, but at least right now do not see anything explicitly saying so in the spec.  It does seem worth mentioning it explicitly.

apinski-cavium:
I actually found the same issue plus the p4c is getting many new testcases for header stacks with header unions so this gets more important to fix the specifications if we want to allow it.

apinski-cavium:
There is a testcase in the p4c already for header stacks of header unions. ./p4_16_samples_outputs/spec-ex09.p4 and p4_16_samples/spec-ex19.p4 which were added with https://github.com/p4lang/p4c/commit/bb47b71882 (well it was moved to the p4_16 directory later on).

jnfoster:
Closing as this seems to have been fixed.

PR #1233, opened 2023-03-08
Fix header stack

pataei:
This PR adds header unions as a subtype of header stacks and adds a new column in the type nesting section for header stacks, i.e., it addresses issues #876 and #1105


mbudiu-vmw (approved):
I don't think we need to discuss this, it's clearly fixing an omission.
If we get one more approval we should merge it.

jnfoster (approved):
LGTM

Issue #1228, opened 2023-02-14
Reason for limiting fixed-length types to have a constant width


pataei:
I'm wondering why P4 spec forces `bit<W>` or `int<W>` in some places, such as when they're used as [the underlying type of a serializable enum](https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-enum-types) or as [the element type of a header](https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-type-nesting). Why doesn't it just require a compile-time-known expression instead of a constant?

Also, does the use of `W` in spec always indicate a constant?

vgurevich:
@pataei  -- compile-time-known expressions are supported as long as you enclose them in parentheses. See Section 7.1.6.2 in the spec.

mbudiu-vmw:
it should be a compile-time known value in the spec.

pataei:
To double check my understanding, as an example, so there isn't anywhere in P4 that allows only `bit<4>` and not `bit<(2+2)>`?

mbudiu-vmw:
These two should be fully equivalent and can be used in the same contexts. See also #1213

pataei:
Closing this after the merge.

PR #1234, opened 2023-03-08
Clarify fixed length types

pataei:
This PR clarifies the use of `bit<W>`, `int<W>`, and `varbit<W>`. More specifically, it states that these represent the case where the width of the type is either a constant integer or a compile-time known expression that evaluates to an integer. In other words, it addresses issue #1228.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3731,10 +3731,9 @@ The following casts are legal in P4:
   loss, and then truncates the result to `W` bits. The compiler should
   emit a warning on overflow.
 - casts between two types that are introduced by `typedef` and are equivalent to one of the above combinations.
-- casts from a typedef to the original type and vice versa.
-- casts from a type introduced by `type` to the original type and vice versa.
-- casts from an `enum` with an explicit type to its underlying type and vice versa.
-- casts between two `enum` with the same underlying type.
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -4209,9 +4217,9 @@ expressions are legal:
   The `index` is an expression that must be one of the following types:
 
   - `int` - an arbitrary-precision integer (Section [#sec-arbitrary-precision-integers])
-  - `bit<W>` - a `W`-bit unsigned integer where W≥0 (Section [#sec-unsigned-integers])
-  - `int<W>` - a `W`-bit signed integer where W≥1 (Section [#sec-signed-integers])
-  - a serializable `enum` with an underlying type that is `bit<W>` or `int<W>` where W≥1 (Section [#sec-enum-types]).
```
pataei, line 4214: This condition is only correct for signed integers as the underlying type of an enum and not for unsigned integers. It might be better to discuss when these types are well-formed (the conditions of their width holds) and state that when a type is used anywhere it's assumed that it is well-formed so we don't have to keep repeating these conditions.

mbudiu-vmw (approved):
This change also looks like could be merged without an LDWG discussion.
But I will wait for at least one more review.

jnfoster (approved):
Agree. LGTM!

PR #1235, opened 2023-03-08
Add distinct fields

pataei:
This PR adds the distinctness of fields for types such as enum, error, struct, header, and header union.


mbudiu-vmw (approved):

mbudiu-vmw:
I think we can merge this right away, without an LDWG discussion, but it would be great to get more reviews.

jafingerhut (approved):
LGTM

PR #1236, opened 2023-03-08
Fix cast

pataei:
This PR addresses some of the issues mentioned in #1223. Specifically, it addresses the following:

- Uses consistent terminology to talk about implicit cast
- Adds grammar for casts
- Expands explicit cast enumeration
- Adds comments in some examples where implicit cast occurs
- Adds the recursiveness of casting

Some questions arose while formalizing some of P4 which are related to this PR:
- [x] Should we add record types (or anonymous header types) so that we can explicitly add them to the casting section, instead of dancing around it every time we use record types to initialize a struct or header? Discussed this in the LDWG meeting on March 6th. Not sure if there's a consensus. I marked specific places where having a specific type of structure-value expression is needed. Suggestion: propose the change in a PR. 
- [x] Other than the potential casting of the subtypes of a list, can a list be casted to another type or be the result of a cast? No. At first glance, it seems that the value of a list can only be written using an explicit cast expression, however, I don't think spec defines it as such. It seems that that's just part of the syntax. However, this is similar to the grammar defined for operations on structure-valued expressions, i.e., `'(' typeref ')' expression`, which is basically what a cast is but it's not recognized as a cast. 
- [x] Are there any restrictions for the nesting of generic types? No, this is checked when types are instantiated.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3689,9 +3689,11 @@ The following casts are legal in P4:
   loss, and then truncates the result to `W` bits. The compiler should
   emit a warning on overflow.
 - casts between two types that are introduced by `typedef` and are equivalent to one of the above combinations.
-- casts between a typedef and the original type.
-- casts between a type introduced by `type` and the original type.
-- casts between an `enum` with an explicit type and its underlying type
+- casts from a typedef to the original type and vice versa.
+- casts from a type introduced by `type` to the original type and vice versa.
+- casts from an `enum` with an explicit type to its underlying type and vice versa.
+- casts between two `enum` with the same underlying type.
+
 
 ### Implicit casts { sec-implicit-casts }
 
@@ -3753,6 +3755,7 @@ several ways that the expression could be manually rewritten into a
 legal expression. Note that for some expression there are several
 legal alternatives, which may produce different results! The compiler
 cannot guess the user intent, so P4 requires the user to disambiguate.
+Thus, it cannot insert implicit casts to make the expression legal.
 
 |-----|-----|-----|
 | Expression | Why it is illegal  | Alternatives |
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2175,9 +2175,9 @@ For example, the declaration
 
 ~ Begin P4Example
 enum bit<8> FailingExample {
-  first           = 1, // implicit cast converts 1 to 8w1
-  second          = 2, // implicit cast converts 1 to 8w2
-  third           = 3, // implicit cast converts 1 to 8w3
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -3655,9 +3661,24 @@ finding this information in a section dedicated to type `varbit`.
...
+    | '(' typeRef ')' expression
+    ;
+
+~ End P4Grammar
+
+The `typeRef` is a type or a name associated to a type.
+More specifically, it can be
+
+- a base type or a `tuple` type
+- a name that represents an `enum`, `struct`, `header`
```
pataei, line 3677: We may need to expand this list if we add anonymous header (struct) types.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3853,7 +3901,7 @@ struct S {
 
 extern void f<T>(in T data);
 
-f((S){ 10, 20 }); // automatically converted to f((S){a = 10, b = 20});
+f((S){ 10, 20 }); // {10, 20} is converted to {a = 10, b = 20}
```
pataei, line 3904: In reality, the tuple is casted to an anonymous struct type but since we don't have an anonymous struct type we cannot say that. So I just stated that the value is converted to a structure-valued expression.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4083,8 +4134,12 @@ struct S {
     bit<32> b;
 }
 const S x = { 10, 20 };             // tuple expression
+                                    // implicit cast of tuple to struct
+                                    // implicit cast of int to bit<32>
 const S x = { a = 10, b = 20 };     // structure-valued expression
+                                    // implicit cast of int to bit<32>
```
pataei, line 4140: Again this is casting an anonymous struct to a struct but we cannot say it since we don't have anonymous struct type yet.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4115,7 +4170,10 @@ automatically becomes valid:
 header H { bit<32> x; bit<32> y; }
 H h;
 h = { 10, 12 };  // This also makes the header h valid
+                 // implicit cast of int to bit<32>
+                 // implicit cast of tuple to header
 h = { y = 12, x = 10 };  // Same effect as above
+                         // implicit cast of int to bit<32>
```
pataei, line 4176: Same thing. Cast from anonymous struct to header. but cannot say it.

pataei:
This pull request isn't ready to be merged yet. I checked the spec up to section 8.19 (operations on header unions) for examples that some sort of cast is happening. I'll continue going through the rest of spec to spell out where cast is done.

mbudiu-vmw:
You should mark it as draft

jnfoster:

PR #1237, opened 2023-03-14
Compile time size determination edit

pataei:
Adds lists to compile time size determination.


apinski-cavium:
list has only one type for its field.
I think list should error out for minSizeInBits/maxSizeInBits because there are only list literals except for passing to externs. You cannot do anything else with them.

jnfoster:
Another perspective: because lists can be assigned to tuples, headers, and structs (Section 8.12) it could be argued that they should also support compile-time size calculations just like values of those types do.

mbudiu-vmw:
These are not tuple expressions, they are the new lists.
This does not work in the compiler at this point.
This would only work for list literals. We don't have any other lists currently, but this could change.

jnfoster:
Oops! Mea culpa. I was reading v1.2.3 (where what we now call lists are really tuples) and not the working draft. You're right.

jafingerhut (approved):
LGTM

mbudiu-vmw:
I will close this one, I think we agreed in the design group that this feature for the list<> type is premature.

PR #1240, opened 2023-04-06
Add expression to represent an invalid header

jfingerh:


jfingerh:
This PR is nearly identical to https://github.com/p4lang/p4-spec/pull/1184 which has been discussed a couple of times in the LDWG.  This version is intended to address some comments from Nate Foster, where he suggested that if we defined `{#}` to be an expression by itself, and that it can be cast to a particular header or header union type if needed by the compiler, and we did this consistently for `{#}` both as a header and header union type value, that would be preferable.

I have created a separate PR so that each of these PRs can be examined separately.  We should merge at most one of them.

jnfoster:
Thanks!

jfingerh:
@mbudiu-vmw What do you think of this PR, if you have time to look at it?  If it meets with your approval, I can volunteer over the weekend to create a modified version of #1220 with similar text as this one.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4180,11 +4180,38 @@ h = { 10, 12 };  // This also makes the header h valid
 h = { y = 12, x = 10 };  // Same effect as above
 ~ End P4Example
 
-Two headers can be compared for equality (==) or inequality (!=) only
+Two headers can be compared for equality (`==`) or inequality (`!=`) only
 if they have the same type.  Two headers are equal if and only if they
 are both invalid, or they are both valid and all their corresponding
 fields are equal.
 
+The expression `{#}` represents an invalid header of some type, but it
```
mbudiu-vmw, line 4188: represents an invalid header or an invalid header union of an unspecified type.
The expression `(H){#}` represents an invalid header or invalid header union with a specified type.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -4180,11 +4180,38 @@ h = { 10, 12 };  // This also makes the header h valid
...
+can be any header or header union type.  A P4 compiler may require an
+explicit cast on this expression in cases where it cannot determine
+the particular header or header union type from the context.
```
mbudiu-vmw, line 4191: If the type of '{#}' cannot be inferred from the context the program will be rejected.

jfingerh: @mbudiu-vmw  Thanks for taking a look.  Just so I am clear, you are recommending replacing my sentence "A P4 compiler may require an explicit cast on this expression in cases where it cannot determine the particular header or header union type from the context." with the text of your comment?  I am fine doing that, if so.

mbudiu-vmw: The main question is whether that is a cast or an entire new expression.
The implementation treats it as a cast and the converts it into a new expression.

p4-16/spec/P4-16-spec.mdk
jfingerh: So this is a minor detail, perhaps, but if I say the following things in the spec:

+ The expression `{#}` represents an invalid header or an invalid header union of an unspecified type.
+ If the type of `{#}` cannot be inferred from the context, the program will be rejected.
+ You may use a cast to convert the type of `{#}` to that of the header or header union type specified by the cast.

Then it seems to me that saying "The expression `(H) {#}` represents an invalid header or invalid header union with a specified type." is on a slippery slope to what I think is one of Nate's objections, which is that it sounds like you are specifying a NEW KIND of expression, which it really is not.  It is simply `{#}` with a cast on it.

mbudiu-vmw: Actually inside the compiler we have three kinds of expressions; these are defined in expressions.def:
`{#}` class Invalid
`(H)` class Cast
`(H){#}` class InvalidHeader or InvalidHeaderUnion
So the description is accurate.
It is just that the parser produces `(cast)(Invalid)` and the typechecker converts that to `InvalidHeader`.

jafingerhut: If you and Nate can come to agreement on what wording you think belongs in the spec for this, and hopefully the same for the expression that represents a value that is an invalid header union, then I suspect most other people would be willing to accept that.

I've got two PRs with my best guesses, but I don't know which of the comments on this PR represent suggested changes to this PR, and which of those Nate would also approve of.

jnfoster (approved):
LGTM.

mbudiu-vmw:
We should strive to merge this before the next release. Any blockers?

jnfoster:
I think this is ready. Shall I merge it?

mbudiu-vmw (approved):

mbudiu-vmw:
Let's go for it.

PR #1245, opened 2023-04-25
Release candidate for v1.2.4

jnfoster:


jnfoster:
@jafingerhut do you recall if there a backward in-compatible change? If there is, we would bump to v1.3.0 I think.

mbudiu-vmw:
I was hoping we could include a few more PRs, including the {#} and the non-const entries.
These are almost ready.

mbudiu-vmw:
Last year we released in June

jnfoster:
Sure, I'm happy to wait a bit longer.

jfingerh:
I do not recall any backwards-incompatible changes.

However, if there were, as a reminder those who follow semantic versioning faithfully would bump up the major version number, not the minor, e.g. we would go from 1.2.3 to 2.0.0.

jafingerhut:
I believe that with this PR merged in a few minutes ago, https://github.com/p4lang/p4-spec/pull/1254 that at least all of my changes to the language spec since version 1.2.3 was released now have a bullet item in the revision history, and I was probably the one most troublesome in ensuring that I added entries to the revision history since then.

Madoko says there are no unresolved cross-references.  I did not do a spelling check on this version, but the last time I did one and fixed things was only a couple of months ago or so, so not much in the way of misspelled words is likely to have crept in since that time.

It looks good to go to me.

jafingerhut (approved):
LGTM

PR #1246, opened 2023-04-26
Fix and prettify references to compile-time size determination

jnfoster:
- Replaced `[#sec-determin]` with `(Section [#sec-determin])` globally
- Also, `sec-determin` doesn't exist. I think it should have been `sec-minsizeinbits`?

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3322,7 +3322,7 @@ serialization data type and representation.
 
 Additionally, the size of a serializable enum can be determined at
 compile-time. However, the size of an enum without an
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
I just looked here recently https://p4.org/p4-spec/docs/p4-16-working-draft.html and noticed that there are a few more broken links, maybe you can take a look at all of them?

mbudiu-vmw:
They show up as "#" in the text.

jnfoster:
Yep, I was going to do that next. Just breaking things down into small pieces for easy review... though, maybe too small :-)

mbudiu-vmw (approved):
Looks good here, but should be tested by generating the markup.

PR #1248, opened 2023-05-01
Fix incorrect cross-reference label

jfingerh:


jfingerh:
Starting from the latest main as of 2023-Apr-30, there is only 1 warning about invalid cross-references that I see when creating HTML and PDF versions of the spec, and this PR fixes that.

It also updates what I believe is the only occurrence of "infinite-precision integer" that has not yet been replaced with "arbitrary-precision integer"

mbudiu-vmw:
In this document https://p4.org/p4-spec/docs/p4-16-working-draft.html you can search for `[#`.

jfingerh:
With commit 2 of this PR, there are no more occurrences of `[#` in the HTML output.

mbudiu-vmw (approved):

jnfoster:
Thanks @jfingerh!

PR #1250, opened 2023-05-01
Allow casts from a type to itself

### LINKED ISSUE #1241: Include no-op casts (t -> t) in list of explicit casts
Right now the list of explicit casts doesn't say that no-op casts (casts from a type to itself) are permitted, but my impression is that they are allowed.

mbudiu-vmw:
Fixes #1241

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3763,7 +3763,9 @@ The following casts are legal in P4:
 - casts between a typedef and the original type.
 - casts between a type introduced by `type` and the original type.
 - casts between an `enum` with an explicit type and its underlying type
-
+- casts where the destination type is the same as the source type
+  if the destination type can be the target of a cast (this excludes 
+  e.g., parsers or externs).
 
 ### Implicit casts { sec-implicit-casts }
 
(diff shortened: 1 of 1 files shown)


apinski-cavium:
Just want to double check this wording for:
```
struct s {}
void f(inout s a) { a = (s)a; }
```
or:
```
enum e {  a, b, c };
void f(inout e a) { a = (e)a; }
```
You explicitly list parsers and extern but not compounded (struct, headers, header_unions) types or non-serializable enums.

I don't mind either way. just wanting to make sure those get clarified because even though they can't be targets of a cast, the compounded type initializer looks like a cast so folks might try it ...

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3763,7 +3763,9 @@ The following casts are legal in P4:
 - casts between a typedef and the original type.
 - casts between a type introduced by `type` and the original type.
 - casts between an `enum` with an explicit type and its underlying type
-
+- casts where the destination type is the same as the source type
```
mbudiu-vmw, line 3766: make a list of all legal possibilities.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3763,7 +3763,9 @@ The following casts are legal in P4:
...
+  if the destination type can be the target of a cast (this excludes
```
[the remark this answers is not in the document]
mbudiu-vmw: I have added this stuff to #1218

p4-16/spec/P4-16-spec.mdk
mbudiu-vmw: I hope this is adequate after the last commit.

jnfoster (approved):

PR #1253, opened 2023-05-15
Add revision history item for initial entries

jafingerhut:


mihaibudiu (approved):

jafingerhut:
Apologies for the weirdly large number of commits in the PR with a bunch of merge branches, but the diff correctly shows that only one file is changed, and only by adding one new bullet item to the revision history.

PR #1254, opened 2023-05-15
Add revision history entry for PR #1132 defining empty key list for tables

jafingerhut:


jafingerhut:
With this PR, I have verified that every commit that I made to the spec since revision 1.2.3 was published has an entry in the revision history.

mihaibudiu (approved):

PR #1256, opened 2023-05-30
Changes to grammar in spec to make it closer to the p4c implementation

jafingerhut:


jafingerhut:
I believe that none of these changes to the grammar in the spec represent a change in the syntax or the semantics of the language at all.  They simply make the grammar.mdk file in the spec more similar to the grammar used in the open source p4c compiler.  See comments I have made on the grammar.mdk changes for some explanation.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3025,15 +3025,17 @@ The grammar production rule for general expressions is as follows:
...
 expression
     : INTEGER
+    | DOTS                           // DOTS is ...
+    | STRING_LITERAL
     | TRUE
     | FALSE
-    | STRING_LITERAL
-    | nonTypeName
-    | dotPrefix nonTypeName
+    | prefixedNonTypeName
```
jafingerhut, line 3032: The definition of `prefixedNonTypeName` is `(nonTypeName | dotPrefix nonTypeName)`, so this and the similar change in grammar.mdk are not a change in the syntax.  This changed version is what p4c uses.

p4-16/spec/grammar.mdk
```diff
@@ -383,6 +385,13 @@ tupleKeysetExpression
     ;
 // END:tupleKeysetExpression
 
+// BEGIN:optTrailingComma
+optTrailingComma
+    : /* empty */
+    | ","
+    ;
+// END:optTrailingComma
+
```
jafingerhut, line 394: No change in the definition of optTrailingComma -- simply moving its definition earlier in grammar.mdk to match the order that p4c defines these non-terminals relative to each other, so `diff` gives fewer differences.

jafingerhut, line 394: No change in the definition of optTrailingComma -- simply moving its definition earlier in grammar.mdk to match the order that p4c defines these non-terminals relative to each other, so `diff` gives fewer differences.

p4-16/spec/grammar.mdk
```diff
@@ -541,7 +550,7 @@ headerStackType
 
 // BEGIN:specializedType
 specializedType
-    : prefixedType "<" typeArgumentList ">"
+    : typeName "<" typeArgumentList ">"
```
jafingerhut, line 553: typeName's only production rule is to expand to prefixedType, so this is not a change in syntax.  The p4c grammar uses the version with typeName.

p4-16/spec/grammar.mdk
```diff
@@ -632,7 +634,7 @@ realTypeArgumentList
 // BEGIN:typeDeclaration
 typeDeclaration
     : derivedTypeDeclaration
-    | typedefDeclaration
+    | typedefDeclaration ";"
```
jafingerhut, line 637: This and the next change result in no syntax change, and matches p4c's grammar.

p4-16/spec/grammar.mdk
```diff
@@ -877,6 +879,12 @@ actionRef
     ;
 // END:actionRef
 
+// BEGIN:entry
+entry
+    : optCONST optEntryPriority keysetExpression ':' actionRef optAnnotations ';'
+    ;
+// END:entry
+
```
jafingerhut, line 887: No change in definition of `entry` -- simply moving it earlier to match order that p4c defines them in.

mihaibudiu (approved):

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3025,15 +3025,17 @@ The grammar production rule for general expressions is as follows:
...
     | expression '[' expression ']'
     | expression '[' expression ':' expression ']'
-    | '{' expressionList '}'
-    | '{' kvList '}'
+    | '{' expressionList optTrailingComma '}'
```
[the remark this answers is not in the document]
jafingerhut: They are one of the few grammar rules that is currently duplicated in both P4-16-spec.mdk and grammar.mdk, which might be part of the reason -- not sure about that.

jnfoster (approved):
Looks good. Thanks for doing this!

PR #1258, opened 2023-07-08
Update spec grammar to more closely match open source p4c implementation

jafingerhut:
In particular, to match the changes made to the p4c implementation with this commit:
https://github.com/p4lang/p4c/commit/64683c958b64b15a5a20c04c5c793a25b6fdb748

### DIFF
--- a/p4-16/spec/grammar.mdk
+++ b/p4-16/spec/grammar.mdk
@@ -886,12 +886,12 @@ entry
     ;
 // END:entry
 
-// BEGIN:optEntryPriority
+// BEGIN:entryPriority
 entryPriority
  : PRIORITY '=' INTEGER ":"
  | PRIORITY '=' '(' expression ')' ":"
  ;
-// END:optEntryPriority
+// END:entryPriority
 
 // BEGIN:entriesList
 entriesList
--- a/p4-16/spec/grammar.mdk
+++ b/p4-16/spec/grammar.mdk
@@ -30,6 +30,7 @@ nonTypeName
     | STATE
     | ENTRIES
     | TYPE
+    | PRIORITY
     ;
 // END:nonTypeName
 
@@ -37,7 +38,6 @@ nonTypeName
 name
     : nonTypeName
     | LIST
-    | PRIORITY
     | TYPE_IDENTIFIER
     ;
 // END:name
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

jafingerhut:
@jonathan-dilorenzo I do not know why, but when I click on the reviewers to try to add jonathan-dilorenzo to the list of reviewers, it does not let me add that user name, so pinging you in a comment instead for your review, if you have time.

jnfoster:
This looks good to me. Let me review it carefully though.

jafingerhut:
A couple of scripts in the p4-spec repo described in this README.md file can help do a diff between the p4c grammar and the spec grammar file: https://github.com/p4lang/p4-spec/blob/main/p4-16/spec/scripts/README.md

Those are the steps I use to find such differences and make the grammars closer to each other.  They are not identical, but they are fairly close to each other.

p4-16/spec/grammar.mdk
```diff
@@ -881,14 +881,14 @@ actionRef
 
 // BEGIN:entry
 entry
-    : optCONST optEntryPriority keysetExpression ':' actionRef optAnnotations ';'
+    : optCONST entryPriority keysetExpression ':' actionRef optAnnotations ';'
+    | optCONST keysetExpression ':' actionRef optAnnotations ';'
     ;
 // END:entry
 
 // BEGIN:optEntryPriority
```
jonathan-dilorenzo, line 889: nit: Here and below, please replace optEntryPriority with entryPriority.

jonathan-dilorenzo (approved):
Looks like it matches the p4c grammar to me. I don't see anywhere we'd actually want to allow PRIORITY as a nonTypeName, but maybe it exists?

p4-16/spec/grammar.mdk
jafingerhut: Another good catch.  Definitely good for consistency's sake to fix that, and should be done with latest commit.

jafingerhut:
> Looks like it matches the p4c grammar to me. I don't see anywhere we'd actually want to allow PRIORITY as a nonTypeName, but maybe it exists?

For the nonterminals like nonTypeName and what they can expand to, I honestly do not understand the issues involved.  @mhaibudiu probably does, as I am simply copying what he did from the p4c implementation when he made some changes.  My understanding is that somehow those most recent changes enabled `priority` to be used as a P4-developer-chosen identifier name in at least some contexts where the previous grammar and p4c implementation prevented that.

jonathan-dilorenzo:
Ah, that would make sense. And I suppose one might want that.

jonathan-dilorenzo (approved):

jnfoster (approved):
LGTM

p4-16/spec/grammar.mdk
jafingerhut: Fixed by commit 2 on this PR.

jonathan-dilorenzo:
Alright. Seeing consensus I will merge this.

PR #1267, opened 2023-09-18
Add initialize method to InternetChecksum

p-sawicki:
I would like to propose adding a new method to the `InternetChecksum` extern which would initialize the incremental checksum computation with the initial checksum value extracted from a header.

The motivation behind the method is that currently to achieve this, one needs to call the `subtract` method which may not be obvious to someone not familiar with the exact formula used for computation so a separate method with a more obvious name would be helpful. Another name considered for this method is `preload`.

Implementation will be simple because in compilers that already support PSA's interface, `ck.initialize(chksum);` can be transformed to `ck.clear(); ck.subtract(chksum);`


jafingerhut (approved):
LGTM

Issue #1273, opened 2024-03-12
L-Value Syntax


vgurevich:
The current spec provides the following definition (section 6.7):
```
prefixedNonTypeName
    : nonTypeName
    | dotPrefix nonTypeName
    ;

lvalue
    : prefixedNonTypeName
    | lvalue "." member
    | lvalue "[" expression "]"
    | lvalue "[" expression ":" expression "]"
    ;
```
This is fine overall, but as one can notice, it does **not** allow parentheses. In other words:
```
a = 1;
```
is OK, but
```
(a) = 1;
```
is not.

This is a minor issue, but it gets confusing for the people that are used to define their C preprocessor macros "defensively" by putting parentheses around the parameters just in case. As a result, macros, defined that way work as R-values, but not as L-values. Here is an example where it hurts:
```
#define DEVPORT_PIPE(port) ((port)[8:7])
```
The macro above should be usable both as the R-value and L-value, but it isn't. (Yes, I know, these macros are not officially allowed, but everyone uses them, since they are not disabled either :) )

I do not know what's the story there. Perhaps it was done in such a way to eliminate the ambiguity in the grammar. In that case, it's just a minor annoyance that we can accept. 

On the other hand, if there is no specific reason to not allow parentheses in L-values, perhaps we should.

jnfoster:
This seems to be fine. I've made a quick patch.

jnfoster:
This has been merged into the compiler. @vgurevich will you write a PR on the spec?

vgurevich:
@jnfoster -- done. And thanks again. Would you like to review it first?

PR #1275, opened 2024-03-17
Allow parentheses in lvalues

vgurevich:
This pull request addresses the issue #1273.

Added the definition allowing parentheses around lvalues to the grammar, similar to what was done in p4lang/p4c#4530.

In addition to that performed a minor cleanup, replacing the (copied) grammar code with the proper INCLUDE.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -1457,12 +1457,7 @@ following expressions are legal l-values:
 ~ Begin P4Grammar
 [INCLUDE=grammar.mdk:prefixedNonTypeName]
 
-lvalue
-    : prefixedNonTypeName
-    | lvalue "." member
-    | lvalue "[" expression "]"
-    | lvalue "[" expression ":" expression "]"
-    ;
+[INCLUDE=grammar.mdk:lvalue]
 ~ End P4Grammar
 
 - Identifiers of a base or derived type.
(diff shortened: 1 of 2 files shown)


jafingerhut:
I do not know if this is important or not, but note that the INCLUDE was not being done before, because the grammar rules were different.  The grammar.mdk file has `THIS` as one of the possible productions, but the spec never had that in it (yet).  Maybe it should, but when I was introducing the use of Madoko's INCLUDE feature in the spec, if you see a grammar rule that is not INCLUDE'd, it is probably for the reason that they are not identical.

vgurevich:
Thanks, @jafingerhut ! I did check, even before changing to INCLUDE to make sure that the rule looked exactly the same in both places, but if you think it might be safer to leave it the way it used to be before, I can revert the change.

jafingerhut:
I do not recall at the moment whether `THIS` or `this` are mentioned anywhere in the P4 language spec, or whether that is a p4c extension.  It is possible that removing `THIS` from the grammar.mdk file in the spec might be an even better approach.

jnfoster:
It is in the spec: 

https://staging.p4.org/p4-spec/docs/P4-16-v1.2.4.html#sec-instantiating-abstract-methods

jafingerhut (approved):
LGTM

jnfoster:
Huzzah! We merged this into the spec at the April 2024 meeting.

jnfoster:
(It doesn't seem worth adding a ChangeLog for this, but if someone disagrees, let me know.)

PR #1282, opened 2024-05-13
Madoko requires blank line before first element of a list

jafingerhut:
Otherwise it does not format it as a bulleted list.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -3340,6 +3340,7 @@ const bool same = exact == fuzzy;  // always 'false'
 ## Expressions on Booleans { #sec-bool-exprs }
 
 The following operations are provided on Boolean expressions:
+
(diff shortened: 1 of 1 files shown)


jafingerhut:
@rcgoodfellow Since you quickly approved a similar small markup fix recently, I wanted to call your attention to this one.

You can see here: https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-bool-exprs what the HTML output of the current Madoko markup looks like.  In particular, note that what was intended to be a bullet list looks like this instead, with all the bullet items run together into a single paragraph:

```
The following operations are provided on Boolean expressions: - “And”, denoted by && - “Or”, denoted by || - Negation, denoted by ! - Equality and inequality tests, denoted by == and != respectively.
```

with the addition of a single blank line into the markup, the bullet list looks as intended.

rcgoodfellow (approved):
LGTM. Thanks!

Issue #1293, opened 2024-07-29
Do we allow `string` type to be aliased by `typedef`?


jaehyun1ee:
The section **7.2.8. Type nesting rules** enumerates what type can be aliased or not by a `typedef` declaration.

The table below seems to omit the case for the `string` type. I expect that new type declaration would not allow `string` type, but I wonder if P4 allows it for typedef case. Checking with the p4c compiler, it currently allows `string` to be aliased, e.g., `typedef string string_t;` typechecks.

![image](https://github.com/user-attachments/assets/0ffe0336-c56f-446d-bcda-ee437c56dd68)

jafingerhut:
Would you be willing to create a PR with a new row for the string type, with entries that match what p4c currently supports, so we can discuss it in a future P4 language design work group meeting?  It seems reasonable to me if the spec follows what the implementation currently does, but discussion at such a meeting might lead to other conclusions.

PR #1294, opened 2024-07-30
Specify valid type alias rule for string type

jaehyun1ee:
Proposing clarification for under-specified case, whether it is valid when `string` is aliased by either `typedef` or `type`. (#1293)

I expect that new type declaration (`type`) does not allow string type, and such is allowed for `typedef`. Checking with the p4c compiler, it currently allows string to be aliased by typedef, e.g.,` typedef string string_t;` typechecks.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2570,6 +2570,7 @@ The table below lists all types that may appear as base types in a
 | `varbit<W>`      | allowed            | error           |
 | `int`            | allowed            | error           |
 | `void`           | error              | error           |
+| `string`         | allowed            | error           |
 | `error`          | allowed            | error           |
 | `match_kind`     | error              | error           |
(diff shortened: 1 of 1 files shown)


jonathan-dilorenzo:
Hi @jaehyun1ee! Thanks a lot for sending this PR.  It turns out that there is another PR (https://github.com/p4lang/p4-spec/pull/1296) that coincidentally fixes this issue as well that we were planning to merge anyway, so I think we'll just merge that one. 

@vlstill, this is a nice separation of concerns though, so you could just merge this and then have yours only deal with the other table? Up to you.

jaehyun1ee:
Thank you! It would be fine to merge @vlstill's PR without merging this one since we were dealing with the same problem :)

vlstill:
If we want to separate concerns, we should add the list case to this PR. I don't mind going either way.

jaehyun1ee:
Closing in favor of #1296

PR #1296, opened 2024-08-01
Clarify nesting rules

### LINKED ISSUE #1285: The type nesting and typedef rules are not clear
The tables in (v1.2.4 spec) [sec 7.2.8 type nesting rules](https://p4.org/p4-spec/docs/P4-16-v1.2.4.html#sec-type-nesting) obviously contain two kinds of "objects" in the "element type"/"base type B".

- actual types (possibly generic), like `int`, `bit<N>`, `bool`
- categories of types defined with specific construct: header, header stack, header union, struct, ...

It would be nice to re-word this so that this distinction is clearer as it can lead to some confuction.

For example, one could interpret the `match_kind` line in the second table as "it is imposible to do `typedef match_kind foo`", or one could interpret it as "one can't apply `typedef` to the match kind values (like `typedef lpm foo`)". I presume the second is what was intended and indeed in P4C, `typedef match_kind foo` works, which, while not useful, seems consistent to me (and `typedef lpm foo` does not work). But this interpretation contrasts with what that line means for e.g. `bit<N>`. Obviously there we are saying that `typedef bit<42> foo` is possible, not that `typedef 8w42 foo` is possible. Or possibly the spec is actually suggesting that `typdef match_kind foo` should not be accepted, in which case we have a bug in the compiler.

But in general I believe it would be clearer to have the first column contain things like: `bit<N>`, `int`, "header types"/"headers", "header stacks", "structures", etc. Maybe with some explanation that we mean e.g. all types defined with the `header` keyword.

### LINKED ISSUE #1293: Do we allow `string` type to be aliased by `typedef`?

vlstill:
This tries to resolve #1285. If merged in the current form, it would also fix #1293 and supercede PR #1294. I tried to make the "categories"/"kinds" of types (like headers and enums) distinct from actual types -- I used for each the name that is used in its section. Furthermore, in the second table (`typedef`/`type` rules) I've added the two missing lines for `string` and for list types. Apparently compiler handles them as would be expected / as I added.

I've also tried to add rationale for the `typedef` rules. I must say I am not happy with it and I wonder if someone can get a better one. Or we may just allow it (C allows typedefing even `void`).

Fixes #1285.

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -2585,12 +2585,9 @@ The table below lists all types that may appear as base types in a
 | a `type` name     | allowed            | allowed         |
 |-------------------|--------------------|-----------------|
 
-Rationale: `void` is not useful except for specifying a function or
-method that does not return anything, therefore there is no reason in
-aliasing it and it could be confusing. `match_kind` values are not
-useful to store in a variable, as they are only used to specify how
-to match fields in table search keys, which are all declared at
-compile time and the declaration does not contain type.
+Rationale: So far, no clear motivation for allowing `typedef` for `void`
+and `match_kind` was presented. Therefore, to be on the safe side this
+is disallowed.
 
 [^type_allowed]: `type B <name>` is allowed for a type name `B`
     defined via `typedef X B` if `type X <name>` is allowed.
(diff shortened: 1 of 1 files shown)


jonathan-dilorenzo:
Discussed in the P4 LDWG. @vlstill is planning to change the rationale to say roughly "We don't know of any reason to want this so we are conservatively disallowing it."

Then ready to merge.

jafingerhut (approved):
LGTM

jonathan-dilorenzo (approved):
LGTM!

PR #1300, opened 2024-08-09
Shorten long lines in code snippets v1

jafingerhut:
Discovered while comparing Madoko and AsciiDoc output.  In two out of three of the places changed by this PR, the text already wraps, or "escapes" its background-shaded box, in Madoko output.  The 3rd one is just quite long, and wraps in AsciiDoc but not Madoko, but seems worth shortening in any case.


rcgoodfellow (approved):
LGTM

PR #1315, opened 2024-10-12
For-loop additions

ChrisDodd:
Started with basic additions to the spec for for loops

### DIFF
--- a/p4-16/spec/P4-16-spec.mdk
+++ b/p4-16/spec/P4-16-spec.mdk
@@ -5284,6 +5284,7 @@ statement
     | exitStatement
     | returnStatement
     | switchStatement
+    | loopStatement
     ;
 
 [INCLUDE=grammar.mdk:assignmentOrMethodCallStatement]
@@ -5522,6 +5523,42 @@ See "Implementing generalized P4_16 switch statements"
 use to implement generalized switch statements.
 
 [GeneralizedSwitchStatements]: <https://github.com/p4lang/p4-spec/blob/master/p4-16/spec/docs/implementing-generalized-switch-statements.md>
+
+## Loop statement { #sec-loop-stmt }
+
+A `loop` statement executes a statement zero or more times, based on a condition
+or a range or collection.
+
+~ Begin P4Grammar
+
+[INCLUDE=grammar.mdk:loopStatement]
+
+[INCLUDE=grammar.mdk:forInitStatement]
+
+[INCLUDE=grammar.mdk:forUpdateStatement]
+
+~ End P4Grammar
+
+The basic 3-clause `for` statement is similar to a C for statement.  The init statements
+will be executed prior to executing the loop.  The condition will be evaluated before
+each iteration, to determine if the loop should exit.  If the condition is false, the loop
+will exit without executing any statements from the loop body or update.
+The update statements will be executed after the loop body and before evaluating the
+condition again for the next iteration.
+
+The `for`-`in` statement executes the body once for each value in a range or each
+element in a list expression or header stack.  The list or range expression itself
+will only be evaluated once, before the first iteration of the loop.  All side effects
+in the list or range expression will occur before the first iteration of the loop body.
+
+A `break;` statement may be executed in a loop body to immediately exit the loop,
+without executing the rest of the body or the update statements.
+
+A `continue;` statement may be executed in a loop body to skip the rest of the current
+loop body, executing the update statements and then evaluating the condition to
+either execute the loop body again, or terminate the loop.
+
+The scope of any declaration in a for statement is limited to the for statement and its body.
 
 # Packet parsing { #sec-packet-parsing }
 
--- a/p4-16/spec/grammar.mdk
+++ b/p4-16/spec/grammar.mdk
@@ -98,7 +98,9 @@ annotationToken
     | APPLY
     | BOOL
     | BIT
+    | BREAK
     | CONST
+    | CONTINUE
     | CONTROL
     | DEFAULT
     | ELSE
@@ -108,6 +110,7 @@ annotationToken
     | EXIT
     | EXTERN
     | FALSE
+    | FOR
     | HEADER
     | HEADER_UNION
     | IF
@@ -737,6 +740,18 @@ assignmentOrMethodCallStatement
     ;
 // END:assignmentOrMethodCallStatement
 
+// BEGIN:breakStatement
+breakStatement
+    : BREAK ";"
+    ;
+// END:breakStatement
+
+// BEGIN:continueStatement
+continueStatement
+    : CONTINUE ";"
+    ;
+// END:continueStatement
+
 // BEGIN:emptyStatement
 emptyStatement
     : ";"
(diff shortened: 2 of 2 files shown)


p4-16/spec/P4-16-spec.mdk
```diff
@@ -5523,6 +5524,37 @@ use to implement generalized switch statements.
...
+~ End P4Grammar
+
+The basic 3-clause `for` statement is similar to a C for statement.  The init statements
+will be executed prior to executing the loop.  The condition will be evaluated before
+the first iteration and after each iteration, to determine if the loop should exit.
+The update statements will be executed after the loop body and before evaluating the
+condition again for the next iteration.
+
+The `for`-`in` statement executes the body once for each value in a range or each
+element in a list expression or header stack.
```
jafingerhut, line 5549: Maybe point out that the range endpoints, and list elements, are only evaluated once before the first execution of the loop body?

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5523,6 +5524,37 @@ use to implement generalized switch statements.
...
+
+A `break;` statement may be executed in a loop body to immediately exit the loop,
+without executing the rest of the body or the update statements.
+
+A `continue;` statement may be executed in a loop body to skip the rest of the current
+loop body, executing the update statements and then evaluating the condition to
+either execute the loop body again, or terminate the loop.
+
```
jafingerhut, line 5557: Do you think it is worth explicitly saying that any loop variable preceded by a type has a scope that is limited to the loop statement?

p4-16/spec/P4-16-spec.mdk
ChrisDodd: This brings up a potentially annoying corner case -- for something like `for T x in (list<T>){ a, b(arg), c(arg) } body` where there are side effects within the list expression, but assoicated with specific elements, must all side effects in those function calls occur before the first iteration?  A simple unrolling might want to evaluate each list element initializer just before that element's iteration.

Probably be best to require evaluating all the side effects first.  After unrolling, copy propagation should be able to move down side effects if that is legal and useful.

p4-16/spec/P4-16-spec.mdk
ChrisDodd: Perhaps just say that the scope of anything declared within a loop statement is limited to the loop statement?  That is consistent with (recent) C++ and logical if we ever add C++ style if statements with initializers.

p4-16/spec/P4-16-spec.mdk
jafingerhut: The only two alternatives I can think of that might make sense are:

(a) specify that all evaluation (including side effects, if any) of evaluating list elements is done before the first loop iteration
(b) specify that the evaluation of a list element is done just before the iteration that uses that list element, which might have changed since the first loop iteration.

I'm not dead-set against (b).  I realize that (a) could require a lot of temporary storage of list elements in the worst case, but we can explain to developers this fact, and they can avoid such programs that require that if they prefer.

jafingerhut:
At 2024-Nov-04 LDWG meeting, it was briefly discussed and agreed upon that for a loop of the form `for (typeRef var in <P4-list-of-expressions>) { ... }`, it seems best to explicitly specify that an implementation should behave the same as if every expression in the list is evaluated before the first iteration of the loop body.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5523,6 +5524,40 @@ use to implement generalized switch statements.
...
+[INCLUDE=grammar.mdk:forInitStatement]
+
+[INCLUDE=grammar.mdk:forUpdateStatement]
+
+~ End P4Grammar
+
+The basic 3-clause `for` statement is similar to a C for statement.  The init statements
+will be executed prior to executing the loop.  The condition will be evaluated before
+the first iteration and after each iteration, to determine if the loop should exit.
```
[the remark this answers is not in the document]
jafingerhut: Perhaps it is because Chris's text is more exhaustive?  The condition is also evaluated after the last iteration.

p4-16/spec/grammar.mdk
```diff
@@ -822,6 +840,59 @@ switchLabel
     ;
 // END:switchLabel
 
+// BEGIN:loopStatement
+loopStatement
+    : optAnnotations FOR "(" optForInitStatements ";" expression ";"
+      optForUpdateStatements ")" statement
+    | optAnnotations FOR "(" typeRef name IN forCollectionExpr ")" statement
```
jafingerhut, line 847: The nonterminal forCollectionExpr is not defined.

jafingerhut: fixed

jafingerhut (approved):
Looks good to me.

p4-16/spec/P4-16-spec.mdk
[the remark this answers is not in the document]
ChrisDodd: There's a fine line between being concise (good) and terse (bad).  For a spec, it is probably better to err on the side of being sure to cover every corner case explicitly.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -5523,6 +5524,42 @@ use to implement generalized switch statements.
...
+
+[INCLUDE=grammar.mdk:loopStatement]
+
+[INCLUDE=grammar.mdk:forInitStatement]
+
+[INCLUDE=grammar.mdk:forUpdateStatement]
+
+~ End P4Grammar
+
+The basic 3-clause `for` statement is similar to a C for statement.  The init statements
```
vlstill, line 5542:
```suggestion
The basic 3-clause `for` statement is similar to a C `for` statement.  The init statements

File: p4-16/spec/P4-16-spec.mdk [hunk 6]
```diff
@@ -5523,6 +5524,42 @@ use to implement generalized switch statements.
...
+in the list or range expression will occur before the first iteration of the loop body.
+
+A `break;` statement may be executed in a loop body to immediately exit the loop,
+without executing the rest of the body or the update statements.
+
+A `continue;` statement may be executed in a loop body to skip the rest of the current
+loop body, executing the update statements and then evaluating the condition to
+either execute the loop body again, or terminate the loop.
+
+The scope of any declaration in a for statement is limited to the for statement and its body.
```
vlstill, line 5561:
```suggestion
The scope of any declaration in a `for` statement is limited to the for statement and its body.

vlstill (commented):
✅  I agree with this, I don't see any problems in the wording, except for some formatting, but that is a minor issue. Also, it seems that the draft version has a remainder of some conflict resolution at the end of this section (https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-loop-stmt)

p4-16/spec/P4-16-spec.mdk
jafingerhut: Thanks for the detailed comments.  Given that this PR has already been merged, it would be great if someone could create a new PR with the proposed changes.  vlstill I do not know if you are interested in doing so?

vlstill: I'll open a PR.

p4-16/spec/P4-16-spec.mdk
jafingerhut: Modified in a separate PR.

PR #1316, opened 2024-10-12
add +: slices

ChrisDodd:


p4-16/spec/P4-16-spec.mdk
```diff
@@ -3489,6 +3489,10 @@ Bit-strings also support the following operations:
   The effect of this statement is to set bits `H` through `L` (inclusive of
   both) of `e` to the bit-pattern represented by `x`, and leaves all other bits
   of `e` unchanged.  A slice of an unsigned integer is an unsigned integer.
+  A slice may also be specified as `[L+:W]` where L is the lowest bit of the slice
+  and W is the width of the slice.  In this case, only W must be a non-negative compile-time
```
jafingerhut, line 3493: Should we restrict `W` to a local compile-time known value, like the spec currently does for both `H` and `L` for the current slices, if I recall correctly?

ChrisDodd: yes

ChrisDodd:
I'm unclear whether these changes need to be also in the asciidoc version, or if that happens automatically.  Or perhaps we should wait until after the switchover to asciidoc and do it only there.

jafingerhut:
I think it is worth discussing at the next LDWG meeting, exactly when Madoko -> AsciiDoc conversion happens.

Fortunately, for text paragraphs with little or no markup, it usually makes no difference.  That said, the changes are so extensive for Madoko->AsciiDoc that I expect if that change is made before this PR, some small changes may be required in this PR.  I think small changes to a PR like this to update it will not be wasted effort.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3489,6 +3489,11 @@ Bit-strings also support the following operations:
   The effect of this statement is to set bits `H` through `L` (inclusive of
   both) of `e` to the bit-pattern represented by `x`, and leaves all other bits
   of `e` unchanged.  A slice of an unsigned integer is an unsigned integer.
+  A slice may also be specified as `[L+:W]` where L is the lowest bit of the slice
+  and W is the width of the slice.  In this case, only W must be a non-negative
+  local compile-time known numeric value.  L must also be a numeric value and
```
jafingerhut, line 3494: Please add markup to this occurrence of L, i.e. `L` with backticks

jafingerhut: Also for the earlier occurrence of W in the paragraph.

p4-16/spec/P4-16-spec.mdk
```diff
@@ -3489,6 +3489,11 @@ Bit-strings also support the following operations:
...
+  some architectures may require it to be compile-time known.  Some architectures
+  may allow variable indexing.
```
vlstill, line 3496: This wording currently allows `L` to be negative (e.g. either negative `int` literal, or `int<N>` expression which can have a negative value). A way out would be to say that its type must be either `int`, in which cast it must be non-negative, or `bit<N>`-expression. I think it would make sense to restrict this, as slices on negative bit index don't make sense.

vlstill (commented):
Looks good to me, except we should add wording that prohibits negative bit indices.

p4-16/spec/P4-16-spec.mdk
ChrisDodd: Yes, this is tricky.  I added some verbiage to require non-negative in-range values, but only require an error message for constants, as it may be tricky to do that for variable values, and we don't have any way of dealing with it at runtime.

p4-16/spec/P4-16-spec.mdk
ChrisDodd: The `int<W>` description was just a copy of the whole paragraph from here, so I shortened that to just refer to the bitstring description.

jafingerhut (approved):
LGTM

p4-16/spec/P4-16-spec.adoc
```diff
@@ -3657,21 +3663,8 @@ The `int<W>` datatype also support the following operations:
...
-  The conditions `+0 <= L <= H < W+` are checked statically (where `W` is
-  the length of the source bit-string). Note that both endpoints of
-  the extraction are inclusive. The bounds are required to be values
-  that are known at compile time so that the width of the result can be
-  computed at compile time. Slices are also l-values, which means that P4
-  supports assigning to a slice: `e[H:L] = x`.
-  The effect of this statement is to set bits `H` through `L` of `e` to the
-  bit-pattern represented by `x`, and leaves all other bits of `e` unchanged.
+  denoted by `[H:L]` or `[L+:W]`, with the same semantics as slices of
+  bit strings.
```
vlstill, line 3667: I think we need to explicitly state that the result is unsigned. One could wonder what happens if the slice starts at the MSB.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -3581,6 +3581,12 @@ Bit-strings also support the following operations:
   The effect of this statement is to set bits `H` through `L` (inclusive of
   both) of `e` to the bit-pattern represented by `x`, and leaves all other bits
   of `e` unchanged.  A slice of an unsigned integer is an unsigned integer.
+  A slice may also be specified as `[L+:W]` where `L` is the lowest bit of the slice
+  and `W` is the width of the slice.  In this case, only `W` must be a non-negative
+  local compile-time known numeric value.  `L` must also be a numeric value and
+  some architectures may require it to be compile-time known.  It must also be
+  non-negative and in-range for the type.  Some architectures may allow variable indexing,
+  which might prevent diagnosing an out of range value.
```
vlstill, line 3589: We should define semantics of out-of-bounds slices. Similar to how we have it defined for header stack indexing, we should not leave it to the compiler.

- we might consider the "out-of-range" bits to be undefined
- or we might want then to be 0s
- we probably want write to out-of-bounds bits to have no effect.

I would personally prefer the read values to be zeroed, e.g.

```p4
bit<42> foo = ~0;
bit<4> bar = foo[a+:4];
```

`bar` would be

- `0b111` for `a = 0`, `a = 10`, etc.
- `0b0011` for `a = 40`
- `0b0000` for `a > 42`

I.e. this would be equivalent to `(bit<4>)(foo >> a)`.

p4-16/spec/P4-16-spec.adoc
ChrisDodd: Yes, the result of a slice should always be unsigned (`bit<n>`, never `int<n>`)

p4-16/spec/P4-16-spec.adoc
ChrisDodd: This should have the same semantics as shift.

jafingerhut:
@vlstill Please take another look to see if you have any further comments or questions, and if not, please approve.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -3577,10 +3577,18 @@ Bit-strings also support the following operations:
...
   The effect of this statement is to set bits `H` through `L` (inclusive of
   both) of `e` to the bit-pattern represented by `x`, and leaves all other bits
-  of `e` unchanged.  A slice of an unsigned integer is an unsigned integer.
+  of `e` unchanged.  A slice is always an unsigned integer.
+  A slice may also be specified as `[L+:W]` where `L` is the lowest bit of the slice
+  and `W` is the width of the slice.  In this case, only `W` must be a non-negative
+  local compile-time known numeric value.  `L` must also be a numeric value and
+  some architectures may require it to be compile-time known.  It must also be
+  non-negative and in-range for the type.  Some architectures may allow variable indexing,
+  in which case an out of range value will be equivalent to a shift of that amount.
```
vlstill, line 3591: I suggest wording this as for shifts, allowing only `int` or `bit<W>`, not `int<W>` (https://p4.org/wp-content/uploads/2024/10/P4-16-spec-v1.2.5.html#sec-a-note-about-shifts). This also matches what was discussed on the last LDWG. We cannot defer to the shift for the `int<W>` index as shift does not allow `int<W>` to avoid probelems with negative values.

ChrisDodd: Maybe add a precise definition for "non-negative" somewhere -- either a `bit<W>` value or a compile time known value that is >= 0.   I don't think we need to (or should) specifically disallow `int<W>` compile-time-known values.

vlstill: I think the formulation that is used for RHS of shifts permits that:

> [...] the right operand of shifts must be either an expression of type bit<S> or a compile-time known value that is a non-negative integer.

I suggest we use the same.

I think a restriction to operation types is something important enough to have it written in the operation description and not hidden in a definition of non-negativity.

jafingerhut: @ChrisDodd Any thoughts on updates to this PR to address this comment?

ChrisDodd: I added a line here to clarify that the index must be a `bit<W>` type if it is not compile-time-known, in order to be compatible with shift.

jafingerhut: I see.  @vlstill Any additional comments, or does it look good to you?

jonathan-dilorenzo:
@vlstill, please take another look or let us know if this is ready to merge so I can merge it. Otherwise, @jonathan-dilorenzo and @vlstill can discuss this in the September LDWG.

jonathan-dilorenzo (approved):
LGTM. Approving and merging. @vlstill, if you have remaining comments, let's fix them in a follow-up CL.

Issue #1295, opened 2024-07-31
Misleading code example for 8.24. Operations on types that are type variables


jaehyun1ee:
Below is the code example from spec section 8.24 explaining the operations on types that are type variables.
A local variable `x` of type `T` is defined inside the apply block of a control declaration.

```p4
control C<T>() {
    apply {
        T x;  // the type of x is T, a type variable
    }
}
```

However, section 14 on control block mentions that,
> Unlike control type declarations, control declarations may not be generic

So the above example is an invalid P4 program where `C` is a generic control declaration (not a control type declaration).
I believe below example would be more appropriate for the purpose. 
```p4
void f<T>() {
   T x; // the type of x is T, a type variable
}
```

vlstill:
I agree the example is invalid and the sentence above it should not mention controls and parsers. Can you please open PR with the change?

PR #1321, opened 2024-10-30
Change code example for sec-type-variable-operations

jaehyun1ee:
Following the discussion #1295, this changes the code example of sec-type-variable-operations such that declaration of a local of a variable type happens inside a generic function, not a generic control.


jafingerhut:
A comment on #1295 suggests also removing the mention of controls and parsers from the paragraph just before this code example.  That seems best to do in this same PR, if it is a correct change.

jaehyun1ee:
Ah, thank you for noticing it :) Also fixed the paragraph in correspondence!

Issue #1287, opened 2024-06-23
Investigate whether PDF/HTML generated for all versions are in git repo history


jafingerhut:
I am not sure, but I think that the way that PDF/HTML is generated for the language spec, the p4lang/p4-spec repository grows in size every time a commit is made to the repo.  I believe some other p4lang spec repos, e.g. p4runtime, do auto-generate PDF/HTML on every commit, but in a way that only the most recent PDF/HTML are ever in the git repo, not the history of all revisions.  I believe that binary PDF does not compress well with git across small changes in the source text.

jafingerhut:
Here is a change made in 2022-Nov to the repo https://github.com/p4lang/pna recommended to me by Antonin Bas that helps keep only the latest version of auto-generated PDF and HTML docs for the PNA specification in the `gh-pages` branch of that repo: https://github.com/p4lang/pna/commit/a791747b3aa3ea033771fd082b6887f17ed92ddb

The `p4runtime` repo also uses the `--amend` option for this purpose.

As far as I can tell, the `p4-spec` repo does not use this `--amend` option when committing new generated HTML and PDF files to the `gh-pages` branch, thus it has full history of all generated versions stored in its commit history.

jafingerhut:
I do not know how to calculate the precise quantitative effect of this, but as a rough kind of evidence that storage is creeping up on the p4-spec repository:

+ The p4c repository totals 105 MBytes of size in its .git directory, containing compressed deltas/versions of all files of the repo's entire history covering 3457 commits, and has 81 MBytes of files in its most recent version (outside of the .git directory).  This averages 105MB/3457commits = 30KBytes/commit

+ The p4-spec repository totals 177 MBytes of size in its .git directory, with 1905 commits, and has 56 MBytes of files in its most recent version (outside of the .git directory).  This averages 177MB/1905commits = 93KBytes/commit

PR #1334, opened 2024-11-14
Add documentation on published specifications and how P4.org finds them

jafingerhut:


jafingerhut:
@rcgoodfellow Some proposed docs on how the "Working draft" links on the P4.org specification pages work today.

jafingerhut:
@rcgoodfellow Do your Github actions for building the HTML and PDF from the AsciiDoc source, put the resulting HTML and PDF files in a place with a known URL that does not change from one run of the Github actions to the next?

If so, we can update the little HTML files on the P4.org server mentioned in this PR, to load those new URLs, instead of the URLs that they load now.

rcgoodfellow:
> Do your Github actions for building the HTML and PDF from the AsciiDoc source, put the resulting HTML and PDF files in a place with a known URL that does not change from one run of the Github actions to the next?

The artifact URLs change from run to run. They are of the form:

```
https://github.com/p4lang/p4-spec/actions/runs/$run_id/artifacts/$artifact_id
```

The [deploy job](https://github.com/p4lang/p4-spec/blob/main/.github/workflows/asciidoc-build.yml#L60-L119) is pushing artifacts to the `gh-pages` branch in an attempt to match the madoku behavior.

I've noticed that images are not rendering from the page

- https://p4.org/p4-spec/docs/p4-16-working-draft.html

I'll look into that as soon as I can.

tools/README.md
```diff
@@ -0,0 +1,135 @@
...
+e.g. [here](https://github.com/p4lang/p4-spec/tree/main/p4-16/spec#building).
+
+
+### Working draft versions of specifications, generated from AsciiDoc source
+
+The details of this are still to be decided.  Ideally they should be
+documnted here after those decisions have been made.
+
+
+### Working draft versions of specifications, generated from Madoko source (historical interest only as of 2024)
```
rcgoodfellow, line 38: The links are still relevant today, as the Asciidoc CI mechanisms are still publishing files to these URLs.

jafingerhut:
It is a space optimization, not a must-have, but at least the older way of generating Madoko PDF and HTML grew the Git history of the p4-spec repo on every commit, because I believe it was keeping the full commit history of all PDF and HTML generated versions.

The P4Runtime spec does not do that -- it uses a different way of publishing the generated PDF and HTML that always overwrites the previous PDF and HTML versions with the latest one, not keeping around the history.  See https://github.com/p4lang/p4-spec/issues/1287 for a few more details if you are curious.  It might be as small of a difference as a single command line option to one git command somewhere.

While you are investigating this, it would be nice if we could curb the future growth of the p4-spec repo in this same way.

PR #1335, opened 2024-11-14
Fix CI issues

rcgoodfellow:
- install rogue from upstream gem instead of building from source
- disable any-branch uploads job for now (possibly redundant with
  asciidoc-build job)


jafingerhut (approved):
LGTM

PR #1337, opened 2024-11-15
embed images in html docs

rcgoodfellow:
This PR fixes the issue of [images not showing up](https://p4.org/p4-spec/docs/p4-16-working-draft.html) in specs built from Asciidocs in CI. The solution is to use the `:data-uri:` directive to embed images as data in the HTML.


jafingerhut (approved):
LGTM

PR #1339, opened 2024-11-23
Add better text allignement for the table captions

Dscano:
Due to these new lines, the table captions will now be located at the bottom of the tables and centered.
@jfingerh @rcgoodfellow


rcgoodfellow:
Thanks @Dscano. Looks like this will only impact the PDF rendering. Should we have captions rendered like this in the HTML as well?

Dscano:
In the HTML, I centered the captions for tables and images and changed the font from bold to lighter. This makes the HTML more closely resemble the PDF

PR #1340, opened 2024-11-24
Add install of libreoffice in asciidoc install script

jafingerhut:


rcgoodfellow:
Having to install libreoffice to build the spec seems a bit heavy. What do you think about storing the images in a format like SVG so they'd still be in an editable form but not need an office suite install to render?

jafingerhut:
Checking with Davide Scano on alternate options besides this PR.   It is currently the case that only the P4Runtime API spec has its image original files in LibreOffice file formats, so a change either to install documentation steps in that repo alone, or as you suggest, saving pre-generated image files in the formats required for AsciiDoc generation are options.  I will close or update this PR after discussing it with Davide.

jafingerhut:
Closing this, as I believe the plan is to add instructions to p4lang/p4runtime repo that LibreOffice must be installed, in addition to running install-asciidoctor-linux.sh script in this repo, in order to build P4Runtime PDF/HTML spec.

Issue #1312, opened 2024-10-07
Clarifications on implicit cast on function/method calls


jaehyun1ee:
There seems to be unspecified bits regarding implicit cast on function/method/(and action) calls.

Spec section **8.20. Method invocations and function calls** mentions:

> The calling convention is copy-in/copy-out (Section [6.8](https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-calling-convention)). For generic functions the type arguments can be explicitly specified in the function call. The compiler only inserts implicit casts for direction in arguments to methods or functions as described in Section [8.11](https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-casts). The types for all other arguments must match the parameter types exactly.

**(1) How do we define implicit cast for action calls?**

Strictly speaking, actions are neither a function nor a method, while they are quite similar constructs.
But, the spec does _not_ mention how casts should be inserted on action calls.
Would it be safe to consider that implicit cast on action arguments behave the same as function/method arguments?

**(2) Do we insert implicit cast for directionless arguments?**

While the spec mentions that we only insert cast for direction `in` arguments, there exist some examples in the p4c compiler test suite where it expects to insert implicit casts for directionless arguments also.

(a) [`extern2.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/extern2.p4)

```p4
extern ext<H> {
    ext(H v);
    void method<T>(H h, T t);
}

control c() {
    ext<bit<16>>(16w0) x;
    apply {
        x.method(0, 8w0);
    }
}
```

Here, it is expected that the first directionless argument, `0` of `x.method` is implicitly casted to `bit<16>`.

(b) [`action-bind.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/action-bind.p4)

```p4
control c(inout bit<32> x) {
    action a(inout bit<32> b, bit<32> d) {
        b = d;
    }
    table t {
        actions = { a(x); }
        default_action = a(x, 0);
    }
    apply {
        t.apply();
    }
}

control proto(inout bit<32> x);
package top(proto p);

top(c()) main;
```

Also here, the second directionless argument `0` for the default action `a` is expected to be implicitly cast to `bit<32>` type.

I wonder whether the p4c test suite is in a disagreement with the current spec, or the current spec is ambiguous.

jaehyun1ee:
For (1), I noticed that the spec mentions the following in section 6.8.1.

[...]

PR #1341, opened 2024-11-25
Allow implicit cast for directionless args

jaehyun1ee:
I would like to propose allowing implicit cast for directionless method / function / constructor arguments, following the discussion made in #1312.

Regarding it, the current spec section 8.20 mentions that:

> The calling convention is copy-in/copy-out (Section [6.8](https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-calling-convention)). For generic functions the type arguments can be explicitly specified in the function call. **The compiler only inserts implicit casts for direction in arguments to methods or functions** as described in Section [8.11](https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-casts). The types for all other arguments must match the parameter types exactly.

So as per the spec, we *cannot* apply implicit cast for directionless **method / function / constructor** arguments.

As a side note, directionless **action** arguments *can* be implicitly cast as per the spec since it mentions the following in section 6.8.1.

> Actions can also be explicitly invoked using function call syntax, either from a control block or from another action. In this case, values for all action parameters must be supplied explicitly, including values for the directionless parameters. **The directionless parameters in this case behave like in parameters.**

So, like the case for action, I am proposing to allow implicit cast for directionless method / function / constructor arguments.

I believe it better reflects the use cases in the p4c test suite. Below is a drop-down list of tests that expect such a feature.
<details>
<summary>List of tests</summary>

* extern-funcs-bmv2.p4
* extern2.p4
* gauntlet_extern_arguments_2.p4
* gauntlet_hdr_in_value-bmv2.p4
* issue1001-1-bmv2.p4
* issue1001-bmv2.p4
* issue1043-bmv2.p4
* issue1334.p4
* issue1642-bmv2.p4
* issue1653-bmv2.p4
* issue1653-complex-bmv2.p4
* issue1660-bmv2.p4
* issue1765-1-bmv2.p4
* issue2648.p4
* issue3246-1.p4
* issue383-bmv2.p4
* issue562-bmv2.p4
* issue933-1.p4
* named-arg1.p4
* v1model-special-ops-bmv2.p4
* bfd_offload.p4
* calc-ebpf.p4
* constructor_cast.p4
* issue1006.p4
* issue1097-2-bmv2.p4
* issue1097-bmv2.p4
* issue1814-1-bmv2.p4
* issue1814-bmv2.p4
* issue1958.p4
* issue2844-enum.p4
* issue298-bmv2.p4
* issue4288.p4
* issue754.p4
* list7.p4
* pr1363.p4
* psa-action-profile1.p4
* psa-action-profile3.p4
* psa-action-profile4.p4
* psa-basic-counter-bmv2.p4
* psa-counter1.p4
* psa-counter2.p4
* psa-counter3.p4
* psa-custom-type-counter-index.p4
* psa-end-of-ingress-test-bmv2.p4
* psa-example-dpdk-byte-alignment_1.p4
* psa-example-dpdk-byte-alignment_2.p4
* psa-example-dpdk-byte-alignment_3.p4
* psa-example-dpdk-byte-alignment_5.p4
* psa-example-dpdk-byte-alignment_6.p4
* psa-example-dpdk-byte-alignment_7.p4
* psa-example-dpdk-byte-alignment_8.p4
* psa-example-dpdk-byte-alignment_9.p4
* pse-example-dpdk-counter.p4
* psa-example-dpdk-externs.p4
* psa-example-dpdk-meter-execute-err.p4
* psa-example-dpdk-meter.p4
* psa-example-dpdk-meter1.p4
* psa-example-dpdk-varbit-bmv2.p4
* psa-meter1.p4
* psa-meter3.p4
* psa-meter7-bmv2.p4
* psa-random.p4
* psa-register-complex-bmv2.p4
* psa-register-read-write-2-bmv2.p4
* psa-register-read-write-bmv2.p4
* psa-register1.p4
* psa-register2.p4
* psa-register3.p4
* rcp.p4
* rcp1.p4
* register-serenum-bmv2.p4
* simple-firewall_ubpf.p4
* unused-counter-bmv2.p4
* value-sets.p4
</details>

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -4880,8 +4880,8 @@ f(y = ya, x = xa);  // match arguments by name in any order
 
 The calling convention is copy-in/copy-out (<<sec-calling-convention>>).  For generic functions the type arguments
 can be explicitly specified in the function call. The compiler only
-inserts implicit casts for direction `in` or directionless arguments to methods or
-functions as described in <<sec-casts>>.  The types for all
+inserts implicit casts for direction `in` or directionless arguments to methods,
+functions, or constructors as described in <<sec-casts>>. The types for all
 other arguments must match the parameter types exactly.
 
 The result returned by a function call is discarded when the function
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -4880,7 +4880,7 @@ f(y = ya, x = xa);  // match arguments by name in any order
 
 The calling convention is copy-in/copy-out (<<sec-calling-convention>>).  For generic functions the type arguments
 can be explicitly specified in the function call. The compiler only
-inserts implicit casts for direction `in` arguments to methods or
+inserts implicit casts for direction `in` or directionless arguments to methods or
 functions as described in <<sec-casts>>.  The types for all
 other arguments must match the parameter types exactly.
 


rcgoodfellow:
From LDWG: This is clarifying and making something more explicit that was implicit in the spec before. We have tests currently that have this assumption baked in, and `p4c` implements the behavior described in the change.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -4880,7 +4880,7 @@ f(y = ya, x = xa);  // match arguments by name in any order
 
 The calling convention is copy-in/copy-out (<<sec-calling-convention>>).  For generic functions the type arguments
 can be explicitly specified in the function call. The compiler only
-inserts implicit casts for direction `in` arguments to methods or
+inserts implicit casts for direction `in` or directionless arguments to methods or
 functions as described in <<sec-casts>>.  The types for all
```
vlstill, line 4884: One more thing, the sentence says "to methods or functions", but there are two more cases that could be relevant:
- directionless arguments of actions with argument explicitly supplied in P4 (I think you have mentioned that for this case the casts are already allowed somewhere);
- arguments of extern and control/parser instances are are directionless arguments, and implicit casts could be applied here too (i.e. `Register<bit<32>, bit<3>>(8, 42)` -- the intializer type is presumably `bit<32>`, but one expects `int` argument will work).

So essentially I would suggest this should be allowed in any cases where a directionless argument can happen (except for directionless argument value coming from control plane, but there I can't see a way how one could want to cast it).

vlstill: If anything, invocations such as `Register<bit<32>, bit<3>>(8, 42)` are invocations of constructors, which apparently spec does consider to be methods.

7.2.10.2. Extern objects:
> Extern object declarations can also optionally declare **constructor methods**;


But for action, for some reason it is its own thing, different from a method or function (arguably it is a bit different, because action is implicitly a closure which has access to all elements of the control it is defined in, while function is completely freestanding and method has only access to the extern object itself (if anything)). So I think at least actions should be mentioned here (i.e. "methods, actions or functions").

jafingerhut:
@jafingerhut Review and either approve or add questions/comments for changes.

p4-16/spec/P4-16-spec.adoc
vlstill: OK, considering the constructors are methods, and `actions` have their own way of enabling it the current wording could be OK. But I still suggest adding actions here and then we can eliminate the sentence "The directionless parameters in this case behave like in parameters." from the wording of action calls as it would all be handled in one place.

jafingerhut (approved):
LGTM

jaehyun1ee:
@jaehyun1ee Just realized that I should mention also the constructor application

jonathan-dilorenzo:
From LDWG: @jaehyun1ee will fix the constructor part and ping me to approve and merge.

jaehyun1ee:
@jonathan-dilorenzo Just added a commit mentioning constructor invocation, I believe it is ready to be merged :)

Issue #1324, opened 2024-11-02
Type nesting rule for extern, parser, control, and package types


jaehyun1ee:
Section 7.2.8 defines the type nesting rules of P4. But it is not explicitly defined for extern, parser, control, and package types.

How about adding the following to the spec? I marked them as (new).

| Element type \ Container type       | header    | header_union | struct or tuple | list | header stack |
| ----------------  | --------  | -------------- | ------------- | --- | ------------ |
| `bit<W>`           | allowed   | error   |  allowed | allowed | error |
| `int<W>`           | allowed   | error   |  allowed | allowed | error |
| `varbit<W>`        | allowed   | error   |  allowed | allowed | error |
| `int`              | error     | error   |  error   | allowed | error |
| `void`             | error     | error   |  error   | error   | error |
| `string`           | error     | error   |  error   | allowed | error |
| `error`            | error     | error   |  allowed | allowed | error |
| `match_kind`       | error     | error   |  error   | allowed | error |
| `bool`             | allowed   | error   |  allowed | allowed | error |
| enumeration types  | allowed     | error   |  allowed | allowed | error |
| header types       | error     | allowed |  allowed | allowed | allowed |
| header stacks      | error     | error   |  allowed | allowed | error |
| header unions      | error     | error   |  allowed | allowed | allowed |
| struct types       | allowed   | error   |  allowed | allowed | error |
| tuple types        | error     | error   |  allowed | allowed | error |
| list types         | error     | error   |  error   | allowed | error |
| (new) extern types | error | error | error | error | error |
| (new) parser types | error | error | error | error | error |
| (new) control types | error | error | error | error | error |
| (new) package types | error | error | error | error | error |

And for `typedef` and `type`,

| Base type B       | `typedef B <name>` | `type B <name>` |
| --------------  | -------------------- | ------------------ |
| `bit<W>`          | allowed            | allowed         |
| `int<W>`          | allowed            | allowed         |
| `varbit<W>`       | allowed            | error           |
| `int`             | allowed            | error           |
| `void`            | error              | error           |
| `string`          | allowed            | error           |
| `error`           | allowed            | error           |
| `match_kind`      | error              | error           |
| `bool`            | allowed            | allowed         |
| enumeration types | allowed            | error           |
| header types      | allowed            | error           |
| header stacks     | allowed            | error           |
| header unions     | allowed            | error           |
| struct types      | allowed            | error           |
| tuple types       | allowed            | error           |
[...]

PR #1343, opened 2024-11-25
Add type nesting rules for extern, parser, control, and package types

jaehyun1ee:
I would like to propose a change to the spec following the discussion made in #1324.
This adds the cases for extern / parser / control / package types for the type nesting rule defined in section 7.2.8.

In summary,
* It *disallows* extern / parser / control / package types from being nested inside a header / header union / struct / tuple / list / header stack.
* It *allows* extern / parser / control / package to be aliased by `typedef`, but *disallows* for `type`.

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -2560,6 +2560,13 @@ arbitrary-precision integer, without a width specified.
 
 | `list types`         | error     | error   |  error   | allowed | error
 
+| `extern types`       | error     | error   |  error   | error   | error
+
+| `parser types`       | error     | error   |  error   | error   | error
+
+| `control types`      | error     | error   |  error   | error   | error
+
+| `package types`      | error     | error   |  error   | error   | error
 |===
 
 [1] An `enum` type used as a field in a `header` must specify a
@@ -2607,19 +2614,27 @@ The table below lists all types that may appear as base types in a
 
 | `bool`              | allowed            | allowed
 
-| enumeration types   | allowed            | error
+| `enumeration types`   | allowed            | error
 
-| header types        | allowed            | error
+| `header types`        | allowed            | error
 
-| header stacks       | allowed            | error
+| `header stacks`       | allowed            | error
 
-| header unions       | allowed            | error
+| `header unions`       | allowed            | error
 
-| struct types        | allowed            | error
+| `struct types`        | allowed            | error
 
-| tuple types         | allowed            | error
+| `tuple types`         | allowed            | error
 
-| list types          | allowed            | error
+| `list types`          | allowed            | error
+
+| `extern types`        | allowed            | error
+
+| `parser types`        | allowed            | error
+
+| `control types`       | allowed            | error
+
+| `package types`       | allowed            | error
 
 | a `typedef` name    | allowed            | allowed [3]
 


jafingerhut:
vlstill agreed to review.  Some discussion in 2025-Jan-13 LDWG meeting.  If vlstill approves, the other members of the LDWG meeting could think of no objections to merging with vlstill's approval.

jaehyun1ee:
Just double-checked with the behavior of p4c,

```p4
extern E {}
// struct S { E e; }
// type E e_new;
parser P() { state start {} }
// struct S { P p; }
// type P p_new;
control C() { apply {} }
// struct S { C c; }
// type C c_new;
package Pkg();
// struct S { Pkg pkg; }
// type Pkg pkg_new;
```

all commented lines result in the program being rejected by the frontend.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -2607,19 +2614,27 @@ The table below lists all types that may appear as base types in a
...
-| enumeration types   | allowed            | error
+| `enumeration types`   | allowed            | error
+
+| `header types`        | allowed            | error
+
+| `header stacks`       | allowed            | error
+
+| `header unions`       | allowed            | error
+
+| `struct types`        | allowed            | error
```
vlstill, line 2625: I think these should not be in backticks. The rule should be something like: if it is an actual piece of code/if it is type (possibly generic) than it is in backticks, but if it is a type "category" (like "header unions") than it should be plain text. For this second table, that was preserved in the asciidoc, but for the first one it probably got mistakenly reformated in the conversion -- compare https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-type-nesting and https://p4.org/wp-content/uploads/2024/10/P4-16-spec-v1.2.5.html#sec-type-nesting. I think the formatting in 1.2.5 is correct, while the draft is wrong in the first table.

I suggest we keep consistent in each of the tables separately in this PR and then file a follow-on PR to fix format in the first table (I can do that once this is merged).

vlstill (commented):
I agree with the changes, except for some formatting issues.

---

With `type` I think for these types of entities it is pretty clear the newly defined type would be (almost) unusable, and we already disallow a lot of types in `type`. So I am fine with disallowing those. I can see much more use for `type X = string` or `type X = int` than for these, but that is for a different discussion :-).

Side note: With these changes, it makes even less sense to prohibit `typedef` on `match_kind` and `void`. Maybe we should just say that any type can be typedef'ed.

p4-16/spec/P4-16-spec.adoc
ChrisDodd: My thought is we should have backticks around (only) things that literally might appear in a P4 program.  So "`enumueration types`" is definitely wrong, but maybe should have "`struct` types"?

jaehyun1ee: I agree to both: (1) having a separate PR to fix the backticks consistently and (2) putting [ `header` types, header stacks, header unions, and `struct` types ].

jafingerhut (approved):
Agreed it would be nice to remove the unwanted backticks mentioned in one of the review comments, but except for that these changes look good, so approving.

jonathan-dilorenzo (approved):
So then I believe the idea is to submit this as is and then @jaehyun1ee (ideally, if you don't mind) will send a follow-up PR fixing the backticks? Please confirm and I will merge.

PR #1345, opened 2024-11-25
Change code example for sec-type-variable-operations

jaehyun1ee:
Following the discussion https://github.com/p4lang/p4-spec/issues/1295, this changes the code example of sec-type-variable-operations such that declaration of a local of a variable type happens inside a generic function, not a generic control.


jafingerhut (approved):
LGTM

Issue #1347, opened 2024-11-25
Is value_set's size compile-time known?


jaehyun1ee:
The spec mentions about the `size` argument to a `value_set` constructor:

> Parser Value Sets support a size argument to provide hints to the compiler to reserve hardware resources to implement the value set.

Although the phrase "hints to the compiler" already implies that the `size` argument should be a ~local~** compile-time known value, can we make it clearly written in the spec? (** edited)

jafingerhut:
Does p4c currently support using a constructor parameter of a control as the size of a value_set?  Or the size of a P4 table?

I hope that the answer is "yes".   At least, I can imagine scenarios where that would be a nice-to-have choice available, as a P4 developer.

If yes, then I would recommend against specifying that these should be local compile-time known values.

jaehyun1ee:
Ah, yes, thank you for the comment :) The P4 spec mentions about the size property of a table,

> When present, its value must always be a compile-time known value that is an integer.

So I would instead propose: `size` argument to a `value_set` constructor should be a compile-time known value.

jafingerhut:
Adding an explicit restriction that the size parameter of a `value_set` must be compile-time known seems reasonable to me.  If it is not already explicitly stated that the value of a table's `size` property must be compile-time known, that also seems like a reasonable addition.

jaehyun1ee:
Thank you for the comment! I made a related PR :)

PR #1352, opened 2024-12-02
Specify that value_set's size is compile-time known

jaehyun1ee:
Following the discussion #1347, this specifies that the `size` argument of a `value_set` must be a compile-time known value.

Although the spec already implies it by mentioning:

> Parser Value Sets support a size argument to provide hints to the compiler to reserve hardware resources to implement the value set.

This revision makes it more explicit.

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -6565,8 +6565,8 @@ include::grammar.adoc[tag=valueSetDeclaration]
 ----
 
 Parser Value Sets support a `size` argument to provide hints to the compiler to
-reserve hardware resources to implement the value set. For example, this parser
-value set:
+reserve hardware resources to implement the value set. Thus, the `size` argument
+must be a compile-time known value. For example, this parser value set:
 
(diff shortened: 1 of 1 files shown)


vlstill (approved):
Looks good to me. I can't see a way in which this could be runtime defined.

jafingerhut (approved):
LGTM

jafingerhut:
@jonathan-dilorenzo @rcgoodfellow We have two positive reviews on this change, and no verbal objections from LDWG meeting of 2025-Jan-13 from several others.  Looks good to us for merging, but leaving it to one of your for final approval/merging.

jonathan-dilorenzo:
LGTM! Merging, but should this also be mentioned in the compile-time known values section?

jafingerhut:
> LGTM! Merging, but should this also be mentioned in the compile-time known values section?

My belief is that the section on compile-time known values primarily lists what kinds of expression _are_ compile-time known values, but does not attempt to enumerate where such values are required.  Where they are required is sprinkled thorughout the document.

Issue #1346, opened 2024-11-25
Clarify instantiation of objects with abstract methods


jaehyun1ee:
The spec defines instantiation of objects with abstract methods in section 11.3.1, and I found two points where the semantics is unclear.

**(1) Restriction on the parameter names of instantiated abstract methods**

The spec mentions only that:

> When instantiating an extern type that has abstract methods users have to supply implementations for all such methods.

However, it does *not* restrict that the parameter names in the implementation must match the parameter names in the abstract method signatures. For instance, [`virtual.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/virtual.p4) in the p4c test suite exhibits such a behavior.

```p4
extern Virtual {
    Virtual();
    abstract bit<16> f();
    abstract void g(inout data ix);
}
...
Virtual() cntr = {
    bit<16> f() {
        return 1;
    }
    void g(inout data x) {}
};
```

Notice that while the abstract method `g` was declared as having an `inout` parameter of `data` type and name `ix`, the provided implementation of `g` has parameter name `x` instead. Also, [`issue2175-1.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2175-1.p4), [`issue2175-3.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2175-3.p4), and [`issue2175-4.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2175-4.p4) behave similarly.

While the code above seems okay, it can get problematic considering another feature of P4, which is method overloading. P4 allows overloading of extern methods through parameter arity or names. This allows a developer to define extern objects like:

```p4
extern Virtual {
    Virtual();
    abstract void g(inout data x);
    abstract void g(inout data y);
}
```

Then, when instantiating the above object, we should be clear in which abstract method we are supplying an implementation for.

```p4
Virtual() cntr = {
    void g(inout data x) {}
    void g(inout data y) {}
};
```

**Thus, I believe that the spec should make clear that the method implementations must have the same parameter names as in the abstract method signature.** Without such a restriction, it would be unclear whether `void g(inout data x) {}` implements `abstract void g(inout data x);` or `abstract void g(inout data y);`.

**(2) Scope when implementing an abstract method**

When implementing an abstract method, the scope is limited to:

[...]

PR #1355, opened 2024-12-04
Clarify instantiation of objects with abstract methods

jaehyun1ee:
Following #1346, this PR adds more rules to the semantics of abstract method initialization.

**First**, it *restricts* that the abstract method being implemented, must have the same parameter names as declared in the object declaration.

This is to (i) disambiguate what abstract method is being implemented, and (ii) provide a clear interface for users.

For example, one may define an extern object with overloaded abstract methods, as follows.

```p4
extern Virtual {
    Virtual();
    abstract bit<32> g(inout bit<32> x); // (1)
    abstract bit<32> g(inout bit<32> y); // (2)
}
```

And in the instantiation site,

```p4
Virtual() virt = {
    bit<32> g(inout bit<32> z) { return x - 1; } // (a)
    bit<32> g(inout bit<32> w) { return x + 1; } // (b)
};
```

*Without* a restriction on matching parameter names, it is unclear whether (a) implements (1) or (2), and (b) implements (1) or (2).

This can further cause confusion when invoking the abstract methods. P4 defines overload resolution via parameter names. If we were to allow such a program, then a user would have to invoke `virt.g(z = 32w42);` and `vir.g(w = 32w42);`, while the extern object declaration suggests that it should be callable by `virt.g(x = 32w42);` or `virt.g(y = 32w42);` instead. i.e., there is a gap between the interface defined in the extern object declaration and what is actually implemented.

The consequences of adding this restriction to the spec would be: [virtual.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/virtual.p4), [issue2175-1.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2175-1.p4), [issue2175-3.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2175-3.p4), and [issue2175-4.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2175-4.p4) in the p4c test suite would have to be rejected by the frontend.

**Second**, it adds the initializer block to the scope of abstract method.

When implementing an abstract method, the scope is limited to the supplied arguments or values that are in the top-level scope.

But this overlooks the case where an instantiation declaration is used inside an object initializer block. Instantiation declaration may be used inside an object initializer block, which is used by program [virtual2.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/virtual2.p4).

```plaintext
objDeclaration
    : functionDeclaration
    | instantiation
    ;
```

```p4
extern Virtual {
    Virtual();
    void run(in bit<16> ix);
    @synchronous(run) abstract bit<16> f(in bit<16> ix);
}
extern State {
    State(int<16> size);
    bit<16> get(in bit<16> index);
}
...
Virtual() cntr = {
    State(16s1024) state;
    bit<16> f(in bit<16> ix) {
        return state.get(ix);
    }
};
```

Notice that `f` uses `state` in its implementation.

Strictly applying what is written in the spec, the program should not be accepted, since `f` refers to `state` that is neither in the top-level scope or the parameter list. i.e., the current spec *syntactically allows* instantiation inside object initializer but *semantically disallows* any way to refer to the instantiated instance.

So this PR adds the object initializer block to the scope as well.

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -5415,10 +5415,14 @@ include::grammar.adoc[tag=objDeclarations]
 include::grammar.adoc[tag=objDeclaration]
 ----
 
+Abstract methods must use the same parameter names as those declared
+in the extern object declaration. This avoids ambiguity about
+which abstract method is being implemented.
+
 The abstract methods can only use the supplied arguments or refer to
-values that are in the top-level scope.  When calling another method
-of the same instance the `this` keyword is used to indicate the
-current object instance:
+values that are in the initializer block or the top-level scope. 
+When calling another method of the same instance the `this` keyword is
+used to indicate the current object instance:
 
 [source,p4]
 ----
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -5415,13 +5415,13 @@ include::grammar.adoc[tag=objDeclarations]
 include::grammar.adoc[tag=objDeclaration]
 ----
 
-Abstract methods must use the same number of parameters, and the same
-parameter directions as those declared in the extern object declaration.
+Abstract method implementations must use the same number of parameters, and
+the same parameter directions as those declared in the extern object declaration.
 Note that overloading of abstract methods is not allowed, thus an
 implementation could be matched to a declaration only with the method name.
 
 The abstract methods can only use the supplied arguments or refer to
-values that are in the initializer block or the top-level scope. 
+values that are in the same initializer block or the top-level scope. 
 When calling another method of the same instance the `this` keyword is
 used to indicate the current object instance:
 
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -2765,7 +2765,7 @@ either by the number of arguments or by the names of the arguments,
 when calls are specifying argument names.  Argument type information
 is not used in disambiguating calls.
 
-Notice that overloading of parsers, controls, or packages is not allowed:
+Notice that overloading of abstract methods, parsers, controls, or packages is not allowed:
 
 [source,p4]
 ----
@@ -5415,9 +5415,10 @@ include::grammar.adoc[tag=objDeclarations]
 include::grammar.adoc[tag=objDeclaration]
 ----
 
-Abstract methods must use the same parameter names as those declared
-in the extern object declaration. This avoids ambiguity about
-which abstract method is being implemented.
+Abstract methods must use the same number of parameters, and the same
+parameter directions as those declared in the extern object declaration.
+Note that overloading of abstract methods is not allowed, thus an
+implementation could be matched to a declaration only with the method name.
 
 The abstract methods can only use the supplied arguments or refer to
 values that are in the initializer block or the top-level scope. 


ChrisDodd:
I think it should follow roughly the same rules as overload resolution -- so if the functionDeclaration unambiguously matches one in the extern (ignoring names), that should be fine.  If the names match exactly to resolve any ambiguity that should be fine too, but if there are multiple abstract functions in the extern and the names don't match, that should be an error.

jaehyun1ee:
I think it is a good idea. Just to clarify, below is my mental model of how it should work.

Overload resolution in P4 is based on argument arity and, if arguments are named, then we use the names to disambiguate. So the strategy would be:

(i) First lookup an abstract method with method name and arity.
(ii) If (i) yields a single method, then matching is done.
(iii) If (i) yields multiple methods, lookup again with method name, parameter names, and arity.
(iv) If (iii) yields a single method, then matching is done.
(v) If (iii) still yields multiple methods, we emit an error due to ambiguity.

```p4
extern Virtual() {
   Virtual();
   abstract bit<32> f(bit<32> a, bit<32> b); // (1)
   abstract bit<32> f(bit<32> c, bit<32> d); // (2)
   abstract bit<32> f(bit<32> a, bit<32> b, bit<32> c); // (3)
}
...
Virtual() cntr = {
   bit<32> f(bit<32> c, bit<32> d) { // (a)
        return 32w42;
    }
    bit<32> f(bit<32> d, bit<32> e, bit<32> f) { // (b)
        return 32w42;
    }
   bit<32> f(bit<32> a, bit<32> b) { // (c)
        return 32w42;
    }
};
```

This would match (1) to (c), (2) to (a), and (3) to (b).

But at the same time, I also think that restricting the abstract method implementations to use the same parameter names as they were declared, helps maintain a consistent call interface.

Referring to the example above, after instantiating the object, we would have to call `cntr.f(d = 32w1, e = 32w2, f = 32w3);` (if using named arguments), which is different from what was proposed in the object declaration, `cntr.f(a = 32w1, b = 32w2, c = 32w3);`.

ChrisDodd:
Hmm.  I think you're probably right that having different argument names should at least be a warning, as it makes calls with named arguments confusing as, depending on context, you might need to use either set of names -- calling `cntr` directly would use the names as you've done, while passing `cntr` as a directionless argument to a control with a `Virtual v` argument would require a call like `v.f(a = 1, b = 2, c = 3);` in that control.

rcgoodfellow:
LDWG discussion:

For accessing abstract methods, the changes proposed to codify what is already in the P4C compiler and should go in.

For the other bit, abstract methods should not be overloaded in a way where the only distinguishing factor is parameter names. Programmers should be free to choose their own names for parameters in implementations of overloaded methods. This is at odds with parameter names being the only distinguishing feature between abstract overloads.

jafingerhut (approved):
LGTM

p4-16/spec/P4-16-spec.adoc
```diff
@@ -2765,7 +2765,7 @@ either by the number of arguments or by the names of the arguments,
 when calls are specifying argument names.  Argument type information
 is not used in disambiguating calls.
 
-Notice that overloading of parsers, controls, or packages is not allowed:
+Notice that overloading of abstract methods, parsers, controls, or packages is not allowed:
```
vlstill, line 2768: I would just point out this forbids overloading of abstract methods completely, i.e. also by the number of arguments. I am OK with that, but I wanted to highlight that.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -5415,10 +5415,15 @@ include::grammar.adoc[tag=objDeclarations]
 include::grammar.adoc[tag=objDeclaration]
 ----
 
+Abstract methods must use the same number of parameters, and the same
```
vlstill, line 5418:
```suggestion
Abstract method implementations must use the same number of parameters, and the same

File: p4-16/spec/P4-16-spec.adoc [hunk 3]
```diff
@@ -5415,10 +5415,15 @@ include::grammar.adoc[tag=objDeclarations]
...
+parameter directions as those declared in the extern object declaration.
+Note that overloading of abstract methods is not allowed, thus an
+implementation could be matched to a declaration only with the method name.
+
 The abstract methods can only use the supplied arguments or refer to
-values that are in the top-level scope.  When calling another method
-of the same instance the `this` keyword is used to indicate the
-current object instance:
+values that are in the initializer block or the top-level scope.
```
vlstill, line 5424:
```suggestion
values that are in the same initializer block or the top-level scope.

vlstill (commented):
Overall looks good to me, although I have two small suggestions.

p4-16/spec/P4-16-spec.adoc
jaehyun1ee: Yes, it was intended because allowing overload via only the arity, and only for abstract methods, would be too complex. But in some future, if someone finds such a feature useful, we might consider lift this restriction then.

vlstill (approved):

jafingerhut:
@jonathan-dilorenzo @rcgoodfellow This has been discussed, updated by Jaeyhun, and approved by two LDWG members.  We think this is ready to merge, but would prefer if a co-chair could do so for final check.

jonathan-dilorenzo:
@jonathan-dilorenzo Jaehyun believes this is ready to be merged. TAL and ideally merge.

jafingerhut:
@jonathan-dilorenzo Reminder to please take a look and merge if you approve.

jonathan-dilorenzo (approved):
Looks great! Approved and will merge!

Issue #1349, opened 2024-11-25
Clarify restrictions on runtime and compile-time call sites


jaehyun1ee:
Spec Appendix F lists restrictions on compile time and run time calls. I have three questions regarding it.

**(1) Calling function from a parser or control top-level**

The appendix explicitly *disallows* invoking a function from a parser or control top-level at runtime. However, some p4c tests ([`issue1538.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue1538.p4), [`issue2957.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2957.p4), [`issue2488-bmv2.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2488-bmv2.p4), [`issue2543-1.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2543-1.p4)) exhibit such a behavior, for example,

```p4
bit<16> incr(in bit<16> x) {
    return x + 1;
}
parser ParserImpl(packet_in packet,
                  out headers hdr,
                  inout metadata meta,
                  inout standard_metadata_t standard_metadata) {
    bit<16> tmp_port = incr((bit<16>) standard_metadata.ingress_port);
    ...
}
```

May I ask the rationale for disallowing such?

**(2) Calling an extern function from a function body**

Understanding `extern` as meaning both an extern method and an extern function in the below table,

![image](https://github.com/user-attachments/assets/cc581315-65d4-40ef-82d7-69e25b1bc7b6)

a function cannot invoke an extern function.

However, [`issue3274.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue3274.p4), [`issue3274-2.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue3274-2.p4), [`issue3292.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue3292.p4), and [`issue4500.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue4500.p4) exhibit such a behavior.

```p4
extern bit<32> f(in bit<32> x, out bit<16> y);
bit<32> x() {
    return f(x = 1, y = _);
}
```

Also, what may be the rationale for disallowing such a behavior?

**(3) Classification of a package initializer**

The restriction of constructor call-sites is as follows.

[...]

PR #1356, opened 2024-12-04
Relax runtime and compile-time callsite restrictions

jaehyun1ee:
Following the discussion in #1349, this proposes some relaxation to **Appendix F: Restrictions on compile time and run time calls**.

**First, for runtime calls**,

(Current)

image

(Proposed) (EDITED)

image

(1) It adds a new column *top level* and refines the column *parser or control top level* (it is written *parser or top level* in the image, but it is likely a typo introduced while migrating to asciidoc) to *top level within a parser or control*.

(2) The explainer for the table has been modified such that the row *extern* means: both an extern method and a function, while it previously meant only the former. Also, the row *function* means: functions but not extern functions. (EDITED) 

(3) It *allows* function calls at the *top level within a parser or control*, but *disallows* function calls at the *top level*. (EDITED)

Allowing function calls at the top level within a parser or control reflects the use case demonstrated in p4c tests: [issue1538.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue1538.p4), [issue2957.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2957.p4), [issue2488-bmv2.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2488-bmv2.p4), [issue2543-1.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2543-1.p4). (EDITED)

For instance,

```p4
bit<16> incr(in bit<16> x) {
    return x + 1;
}
parser ParserImpl(packet_in packet,
                  out headers hdr,
                  inout metadata meta,
                  inout standard_metadata_t standard_metadata) {
    bit<16> tmp_port = incr((bit<16>) standard_metadata.ingress_port);
    ...
}
```

(4) It *allows* extern calls at a function. This reflects the use case in [issue3274.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue3274.p4), [issue3274-2.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue3274-2.p4), [issue3292.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue3292.p4), and [issue4500.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue4500.p4).

```p4
extern bit<32> f(in bit<32> x, out bit<16> y);
bit<32> x() {
    return f(x = 1, y = _);
}
```

**Second, for compile-time constructor calls,**

(Current)

image

(Proposed)

image

The column for *package* has been modified to allow instantiation of package, parser, control, and extern. Here, "*this type* can be instantiated in a package" means, that, *this type* can be instantiated as a constructor parameter for a package.

Note that a package instantiation can only happen at the top-level, while it involves nameless instantiations of package, parser, control, and extern objects as constructor arguments. With the current restriction, there is no clear way of classifying such nameless instantiation according to the table. But with the revised restriction, we may consider them as valid.

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -9553,9 +9553,9 @@ One way that an extern can be called from the top level of a parser or
 control is in an initializer expression for a declared variable,
 e.g. `bit<32> x = rand.get();`.
 
-[.center,cols="3,^2,^2,^2,^2,^2,^2",width=90%]
+[.center,cols="3,^2,^2,^2,^2,^2,^2,^2",width=90%]
 |===
-| This type 6+^| can be called at run time from this place in a P4 program
+| This type 7+^| can be called at run time from this place in a P4 program
 
 |             .>| top level .>| parser state .>| control apply block .>| top level within a parser or control .>| action .>| extern .>| function
 
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -9490,13 +9490,13 @@ program, but constants may.
 
 |             | top level | package | parser  | control | extern | function
 
-| package     | yes       |  no     | no      | no      |  no    |  no
+| package     | yes       |  yes    | no      | no      |  no    |  no
 
-| parser      |  no       |  no     | yes     | no      |  no    |  no
+| parser      |  no       |  yes    | yes     | no      |  no    |  no
 
-| control     |  no       |  no     | no      | yes     |  no    |  no
+| control     |  no       |  yes    | no      | yes     |  no    |  no
 
-| extern      | yes       |  no     | yes     | yes     |  no    |  no
+| extern      | yes       |  yes    | yes     | yes     |  no    |  no
 
 | function    | yes       |  no     | no      | no      |  no    |  no
 
@@ -9525,7 +9525,7 @@ extern functions), actions, and functions.
 
 | control     |  no       |  no     | no      | no      |  no
 
-| extern      | yes       | yes     | yes     | no      |  no
+| extern      |  yes      | yes     | yes     | no      |  no
 
 | table       |  no       |  no     | no      | no      |  no
 
@@ -9547,7 +9547,7 @@ The next table lists restrictions on what kinds of calls can be made
 from which places in a P4 program.  Calling a parser, control, or
 table means invoking its `apply()` method. Calling a value-set means using it
 in a select expression. The row for `extern` describes where extern method
-calls can be made from.
+or function calls can be made from.
 
 One way that an extern can be called from the top level of a parser or
 control is in an initializer expression for a declared variable,
@@ -9557,25 +9557,25 @@ e.g. `bit<32> x = rand.get();`.
 |===
 | This type 6+^| can be called at run time from this place in a P4 program
 
-|             .>| parser state .>| control apply block .>| parser or top level .>| action .>| extern .>| function
+|             .>| top level .>| parser state .>| control apply block .>| top level within a parser or control.>| action .>| extern .>| function
 
-| package     | N/A       | N/A     | N/A       | N/A     | N/A    | N/A
+| package     | N/A 	      | N/A       | N/A     | N/A       | N/A     | N/A    | N/A
 
-| parser      | yes       |  no     | no        | no      | no     | no
+| parser      | no            | yes       |  no     | no        | no      | no     | no
 
-| control     |  no       | yes     | no        | no      | no     | no
+| control     | no            |  no       | yes     | no        | no      | no     | no
 
-| extern      | yes       | yes     | yes       | yes     | no     | no
+| extern      | yes           | yes       | yes     | yes       | yes     | no     | yes
 
-| table       |  no       | yes     | no        | no      | no     | no
+| table       | no            |  no       | yes     | no        | no      | no     | no
 
-| value-set   | yes       | no      | no        | no      | no     | no
+| value-set   | no            | yes       | no      | no        | no      | no     | no
 
-| action      |  no       | yes     | no        | yes     | no     | no
+| action      | no            |  no       | yes     | no        | yes     | no     | no
 
-| function    | yes       | yes     | no        | yes     | no     | yes
+| function    | yes           | yes       | yes     | yes        | yes     | no     | yes
 
-| value types | N/A       | N/A     | N/A       | N/A     | N/A    | N/A
+| value types | N/A           | N/A       | N/A     | N/A       | N/A     | N/A    | N/A
 
 |===
 
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -9557,7 +9557,7 @@ e.g. `bit<32> x = rand.get();`.
 |===
 | This type 6+^| can be called at run time from this place in a P4 program
 
-|             .>| top level .>| parser state .>| control apply block .>| top level within a parser or control.>| action .>| extern .>| function
+|             .>| top level .>| parser state .>| control apply block .>| top level within a parser or control .>| action .>| extern .>| function
 
 | package     | N/A 	      | N/A       | N/A     | N/A       | N/A     | N/A    | N/A
 
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -9547,11 +9547,15 @@ The next table lists restrictions on what kinds of calls can be made
 from which places in a P4 program.  Calling a parser, control, or
 table means invoking its `apply()` method. Calling a value-set means using it
 in a select expression. The row for `extern` describes where extern method
-or function calls can be made from.
+or extern function calls can be made from. The row for `function` describes
+where calls to functions, but not extern functions, can be made from.
 
 One way that an extern can be called from the top level of a parser or
 control is in an initializer expression for a declared variable,
-e.g. `bit<32> x = rand.get();`.
+e.g. `bit<32> x = rand.get();`. Here, the top level of a parser or control
+means either within a parser declaration, but outside of any of its parser
+state declarations, or within a control declaration, but outside its
+`apply { ... }` block.
 
 [.center,cols="3,^2,^2,^2,^2,^2,^2,^2",width=90%]
 |===
@@ -9573,11 +9577,13 @@ e.g. `bit<32> x = rand.get();`.
 
 | action      | no            |  no       | yes     | no        | yes     | no     | no
 
-| function    | yes           | yes       | yes     | yes        | yes     | no     | yes
+| function    | no [1]        | yes       | yes     | yes        | yes     | no     | yes
 
 | value types | N/A           | N/A       | N/A     | N/A       | N/A     | N/A    | N/A
 
 |===
+
+[1] Calls to function `static_assert` are exceptionally allowed at the top level.
 
 There may not be any recursion in calls, neither by a thing calling
 itself directly, nor mutual recursion.
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -9569,7 +9569,7 @@ state declarations, or within a control declaration, but outside its
 
 | control     | no            |  no       | yes     | no        | no      | no     | no
 
-| extern      | yes           | yes       | yes     | yes       | yes     | no     | yes
+| extern      | no            | yes       | yes     | yes       | yes     | no     | yes
 
 | table       | no            |  no       | yes     | no        | no      | no     | no
 


rcgoodfellow:
LDWG notes:

General consensus is to accept this PR. However, there should be an addition to specify that some extern functions may not be accepted by some backends at the top level.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -9547,35 +9547,35 @@ The next table lists restrictions on what kinds of calls can be made
...
 e.g. `bit<32> x = rand.get();`.
 
-[.center,cols="3,^2,^2,^2,^2,^2,^2",width=90%]
+[.center,cols="3,^2,^2,^2,^2,^2,^2,^2",width=90%]
 |===
-| This type 6+^| can be called at run time from this place in a P4 program
+| This type 7+^| can be called at run time from this place in a P4 program
 
-|             .>| parser state .>| control apply block .>| parser or top level .>| action .>| extern .>| function
+|             .>| top level .>| parser state .>| control apply block .>| top level within a parser or control .>| action .>| extern .>| function
```
jafingerhut, line 9560: @rcgoodfellow Regarding your question during the meeting about the meaning of "top level" and "top level within a parser or control", if the text above the table said the following, would that make it clear for you?

"Top level" refers to the part of a P4 program that one is in at the beginning of the file, where all declarations have global scope, i.e. not within any control, parser, or function definition.

"Top level within or control" means either within a parser definition, but outside of any of its parser state definitions, or within a control definition, but outside of its `apply { ... }` block.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -9547,35 +9547,35 @@ The next table lists restrictions on what kinds of calls can be made
...
+| package     | N/A 	      | N/A       | N/A     | N/A       | N/A     | N/A    | N/A
 
-| parser      | yes       |  no     | no        | no      | no     | no
+| parser      | no            | yes       |  no     | no        | no      | no     | no
 
-| control     |  no       | yes     | no        | no      | no     | no
+| control     | no            |  no       | yes     | no        | no      | no     | no
 
-| extern      | yes       | yes     | yes       | yes     | no     | no
+| extern      | yes           | yes       | yes     | yes       | yes     | no     | yes
```
jafingerhut, line 9568: Suggest adding the following text before this table to elaborate on the "yes" in the "top level" column for the row "extern":

While the language syntactically allows extern function or method calls at the top level of a program, an implementation may restrict which extern functions or methods it allows in these places.  For example, an implementation might only support extern functions or method calls with the annotation `@pure`.  Any such calls are executed before the first packet is processed by the device.

(I am less sure about whether the last sentence is worth including.  It could open up a can of worms that I have not yet thought of.)

jaehyun1ee: This is a related question on top-level function/method calls.

I believe the only place where top-level function/method calls can happen is as constant initializers, e.g.,

```p4
const bit<32> c = f(...);
```

because the syntax only allows:

```plaintext
declaration
    : constantDeclaration
    | externDeclaration
    | actionDeclaration
    | parserDeclaration
    | typeDeclaration
    | controlDeclaration
    | instantiation
    | errorDeclaration
    | matchKindDeclaration
    | functionDeclaration
    ;
```

Then, the function/methods *must* produce local compile-time known values, since constant initializers should be local compile-time known.

... but is it? According to section 18.1, no sorts of function/method call is categorized as local compile-time known.

jafingerhut: Good catch.

It seems _possible_ that we could define a subset of function/method calls as local compile-time known values, but this seems like it might be a fairly nit-picky thing to do well, e.g. only pure functions whose parameters are all local compile-time known.  It seems very odd to me to call an expression with a function call, where the function call includes side effects on extern state, a local compile-time known expression, but perhaps it could be reasonable.  I just don't know if we want to go that far into the rabbit hole on this one.

Perhaps we just put "no" for function/method calls in the "top level" column of the table for now, and generalize it later if someone really wants to?

Are any function/method calls supported at the top level by p4c today?

jaehyun1ee: Consulting to my frontend implementation, it seems like `static_assert` is the only case where any kind of call (except constructor calls) is made in the top level.

```p4
struct Headers {
}

struct Meta {
}

const bool test = static_assert(V1MODEL_VERSION >= 20160101, "V1 model version is not >= 20160101");
```

jafingerhut: So it seems like we might want to define that `static_assert` returns a compile-time known value (I don't see a reason why we need to prepend "local" to that).  The spec already says that the parameters to `static_assert` must be compile-time known values.

Perhaps in the "top level" column, we could say "no" for extern function/method calls, but with a footnote explaining the exception that `static_assert` calls are allowed at the top level?

jaehyun1ee: I agree on saying "no" for function/method calls from the top-level, but with a footnote excusing `static_assert`.

Just to clarify, this PR also makes the following change,

```diff
in a select expression. The row for `extern` describes where extern method
- calls can be made from.
+ or function calls can be made from.
```

so extern functions belong to row "extern", and not "function".

And about "local" compile-time known, I suppose it should be "local" because section **11.1 Constants** says:

> The initializer expression must be a local compile-time known value.

, which also suggests that the parameters to `static_assert` should be local compile-time known rather than just compile-time known.

jafingerhut: If we have an "extern" row and a "function" row, then I believe that (perhaps confusingly, and worth saying explicitly somewhere), that "extern" means "extern function or extern method", and "function" means "function, but NOT and extern function".

It seems reasonable to me to also say "no" for column "top level", row "function", too, for the same reasons that it could get pretty nit-picky very quickly to define which subset of functions are supported.  For example, functions can call extern functions or methods in their body.

I guess that at the top level, everything that is compile-time known is _also_ local compile-time known, because there can be no constructor parameters in scope at the top level.  I don't have a strong preference for whether "lcoal" is there or not.

p4-16/spec/P4-16-spec.adoc
rcgoodfellow: SGTM, thanks!

p4-16/spec/P4-16-spec.adoc
jaehyun1ee: I also think that being compile-time known in the top level would *imply* being a local compile-time known value. But this being the case, I think making it *explicitly* written in the spec (e.g., "`static_assert` returns a local-compile time known value") is more accurate, and also in line with section 18.1 that makes a clear definition of the two concepts.

I guess this is not directly related to this PR, as the revisions to **Appendix F** need not discuss local compile-time known or compile-time known - ness. Maybe there should be a separate PR, revising the section for static_assert :)

jaehyun1ee:
I've made edits, reflecting the discussions made above.

- Adds explainer on `function` row, and `top level within a parser or control` column.
- Disallow function calls from the top-level, with a footnote excusing `static_assert` calls.
- Regarding the `extern` row, there was an explainer already:

   > Note that while the extern row shows that extern methods can be called from many places, particular externs may have additional restrictions not listed in this table. Any such restrictions should be documented in the description for each extern, as part of the documentation for the architecture that defines the extern.

jonathan-dilorenzo:
@jonathan-dilorenzo Jaehyun believes this is ready to be merged. TAL and ideally merge.

jafingerhut:
@jaingerhut Review this pr and the discussion, and either approve or add a comment with explicit suggestions for changes.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -9547,38 +9547,44 @@ The next table lists restrictions on what kinds of calls can be made
...
+| table       | no            |  no       | yes     | no        | no      | no     | no
 
-| value-set   | yes       | no      | no        | no      | no     | no
+| value-set   | no            | yes       | no      | no        | no      | no     | no
 
-| action      |  no       | yes     | no        | yes     | no     | no
+| action      | no            |  no       | yes     | no        | yes     | no     | no
 
-| function    | yes       | yes     | no        | yes     | no     | yes
+| function    | no [1]        | yes       | yes     | yes        | yes     | no     | yes
```
vlstill, line 9580: Why are functions disallowed at top-level? The table in the description allows them. There is of course a limitation to when it will be possible to evaluate them, which is related to the whole constness problem. But the same holds for initializers of constants that are not in top-level, so I would guess this problem would be handled in other parts of the spec.

vlstill (commented):
> The column for package has been modified to allow instantiation of package, parser, control, and extern. Here, "this type can be instantiated in a package" means, that, this type can be instantiated as a constructor parameter for a package.
>
> Note that a package instantiation can only happen at the top-level, while it involves nameless instantiations of package, parser, control, and extern objects as constructor arguments. With the current restriction, there is no clear way of classifying such nameless instantiation according to the table. But with the revised restriction, we may consider them as valid.

@jaehyun1ee  Could you please provide example for this one?

p4-16/spec/P4-16-spec.adoc
vlstill: I mean the table in the PR description, which is different from the table in the PR content.

I would be fine with making this conservative, but if we are changing `yes` to `no` we should also consider what the compiler currentely support. Also I would expect *extern functions* to be under *externs* not *functions*.

jafingerhut: vlstill asked: "Why are functions disallowed at top-level?"

I looked through all of the example test programs that Jaehyun linked to in his description, and did not notice any of them attempting to call a function at the top level.

The current spec, before this PR was created, does not support function calls at the top level, as far as I can tell (pointers to specific places in the spec saying otherwise are welcome).

@vlstill Do you have an example P4 program that calls a (regular) function at the top level, that you hope becomes supported?  It isn't obvious to me what such an example program would look like, regardless of whether today's p4c supports it.  That said, if there is such an example program you have in mind, and p4c accepts it today, I'd love to see one.

jafingerhut: My imagination improved somewhat on thinking about it, and I attach an example program that tries two different ways to call a regualar P4 function at the top level of a P4 program.  Even for a regular function that is a pure function, no side effects, no calls to extern functions or methods, the latest p4c as of 2025-Apr-01 gives errors when attempting to compile such code.  I copied the error messages I saw when attempting to compile the program into the source code, immediately before the two commented-out lines that caused compile-time errors when I attempted to uncomment them and compile.

jafingerhut:
@jaehyun1ee Sorry for the very long delay before asking this question.

In your new proposed table, row "extern", new column "top level", you have "yes".

Do you know the reason for this?  Is it because in the current version, row "extern" and column "parser or top level" has "yes", and the "yes" was copied from there to the new "top level" column?

I am pretty sure we have p4c test programs that call extern object constructors from the top level, to instantiate extern objects at the top level.

Do you know whether we have any test program that call non-constructor extern methods at the top level?

Or that call extern functions at the top level?

The reason I ask is that in general, defining the meaning of such calls seems potentially tricky, because of the nearly arbitrary side effects on the system state that such calls can have.  Supporting them at the top level seems to raise questions like:

(a) Are there any desirable use cases for such a thing?  If no one has a use case for such a thing, given the tricky questions they seem to raise (below), why explicitly say we support them?

(b) when do these calls occur?  After the P4 program is loaded, but before the first packet is processed?

(c) in what order are they called?  Without specifying this, different call orders could result in different initial states.

(d) What about extern methods/functions that effectively access or modify the state of the current packet being processed?  Such a call at the top level, if it should be done before the first packet is processed, has no packet to operate upon.

jaehyun1ee:
@vlstill 

Sorry for the long delay, and thank you for your review :)

> > The column for package has been modified to allow instantiation of package, parser, control, and extern. Here, "this type can be instantiated in a package" means, that, this type can be instantiated as a constructor parameter for a package.
> > Note that a package instantiation can only happen at the top-level, while it involves nameless instantiations of package, parser, control, and extern objects as constructor arguments. With the current restriction, there is no clear way of classifying such nameless instantiation according to the table. But with the revised restriction, we may consider them as valid.
> 
> @jaehyun1ee Could you please provide example for this one?

This is quite common, where package instantiation involves nameless instantiation of parsers and controls.

```p4
V1Switch(p = ParserImpl(),
         ig = ingress(),
         vr = verifyChecksum(),
         eg = egress(),
         ck = computeChecksum(),
         dep = DeparserImpl()) main;
```

The above code snippet, for example, is instantiating a `V1Switch` package at the top level. And it involves nameless instantiations of parsers and packages, `p`, `ig`, `vr`, `eg`, `ck`, and `dep`.

What I was trying to point out was, the current spec does not have a clear way of allowing such nameless instantiations. According to the current table in the spec explaining **can be instantiated in this place (Appendix F)**, the instantiation of `p`, `ig`, `vr`, `eg`, `ck`, and `dep` don't seem to fall into any column for instantiation sites: **top level**, **package**, **parser**, **control**,  **extern**, and **function**.

p4-16/spec/P4-16-spec.adoc
jaehyun1ee: @vlstill 

> Also I would expect extern functions to be under externs not functions.

Yes, in the PR content, it explains:

> The row for `extern` describes where extern method or extern function calls can be made from. The row for `function` describes where calls to functions, but not extern functions, can be made from. ([link](https://github.com/p4lang/p4-spec/blob/03ed212453b12dd379a587f1ad1916306cb762d9/p4-16/spec/P4-16-spec.adoc?plain=1#L9549-L9551))

jaehyun1ee: > I looked through all of the example test programs that Jaehyun linked to in his description, and did not notice any of them attempting to call a function at the top level.

@jafingerhut Thank you for pointing it out :) The example test programs illustrate function calls made at the top level within a parser or control, to justify saying "yes" to "allow function calls at the top-level within a parser or control".

@vlstill The original intention when I made the PR was to first distinguish the top-level and the top-level within a parser or control. Then, it led to a follow-up question, "do we allow function calls at the top-level"? At the point of making the PR, I thought we should *allow* function calls at the top-level, as explained in the PR description.

But in discussion with Andy, we figured out that the only cases where a function call may happen at the top-level are as constant initializers or object instantiation arguments, which both require the value to be compile-time known. Yet, the spec does not classify function call expressions as neither local compile-time known or compile-time known. i.e., as per the spec, there is no clear way to use a function call at the top level. Thus I've changed the table to say "no" to "allow function calls at the top-level".

To clarify this I've also made edits to the PR description.

jaehyun1ee:
@jafingerhut 

> In your new proposed table, row "extern", new column "top level", you have "yes".
> Do you know the reason for this? Is it because in the current version, row "extern" and column "parser or top level" has "yes", and the "yes" was copied from there to the new "top level" column?

Yes, I've copied it from the *parser or control top level* column from the table prior to this PR. But on second thought, like how we shifted to disallowing function calls at the top level at the discussion above, I think it would be conservative to disallow extern methods and function calls at the top level as well. Of course we may list exceptional cases where we may allow such as a footnote, if necessary. 

We can apply the same argument for disallowing function calls at the top-level:

```
declaration
    : constantDeclaration // here
    | externDeclaration
    | actionDeclaration
    | parserDeclaration
    | typeDeclaration
    | controlDeclaration
    | instantiation // here
    | errorDeclaration
    | matchKindDeclaration
    | functionDeclaration
    ;
```

The only places where extern methods or function calls may happen at the top-level are as constant initializers or instantiation arguments, but they require compile-time known values, which is not the property of extern methods or function call expressions.

Also, in p4cherry I have not encountered a p4c test program that tries to call a function at the top-level ([link to code](https://github.com/kaist-plrg/p4cherry/blob/e1f0d276d0fd6e986da57257810b84579ed30a6c/p4/lib/typing/typecheck.ml#L1948-L1994)):

```ocaml
and check_call_site (cursor : Ctx.cursor) (ctx : Ctx.t) (ft : FuncType.t) : unit
    =
  match cursor with
  | Global ->
      F.asprintf "(check_call_site) %a cannot be called from top level"
        (FuncType.pp ~level:0) ft
      |> error_no_info
  ...
```

Here, `Global` corresponds to the top-level, and no p4c test program has emitted an error here, supporting that we do not have a use case of a function/extern method/extern function call made at the top-level for now in the test suite.

jafingerhut:
> @jafingerhut
> 
> > In your new proposed table, row "extern", new column "top level", you have "yes".
> > Do you know the reason for this? Is it because in the current version, row "extern" and column "parser or top level" has "yes", and the "yes" was copied from there to the new "top level" column?
> 
> Yes, I've copied it from the _parser or control top level_ column from the table prior to this PR. But on second thought, like how we shifted to disallowing function calls at the top level at the discussion above, I think it would be conservative to disallow extern methods and function calls at the top level as well. Of course we may list exceptional cases where we may allow such as a footnote, if necessary.
> 
> We can apply the same argument for disallowing function calls at the top-level:
> 
> ```
> declaration
>     : constantDeclaration // here
>     | externDeclaration
>     | actionDeclaration
>     | parserDeclaration
>     | typeDeclaration
>     | controlDeclaration
>     | instantiation // here
>     | errorDeclaration
>     | matchKindDeclaration
>     | functionDeclaration
>     ;
> ```
> 
> The only places where extern methods or function calls may happen at the top-level are as constant initializers or instantiation arguments, but they require compile-time known values, which is not the property of extern methods or function call expressions.
> 
> Also, in p4cherry I have not encountered a p4c test program that tries to call a function at the top-level ([link to code](https://github.com/kaist-plrg/p4cherry/blob/e1f0d276d0fd6e986da57257810b84579ed30a6c/p4/lib/typing/typecheck.ml#L1948-L1994)):
> 
> ```ocaml
> and check_call_site (cursor : Ctx.cursor) (ctx : Ctx.t) (ft : FuncType.t) : unit
>     =
>   match cursor with
>   | Global ->
>       F.asprintf "(check_call_site) %a cannot be called from top level"
>         (FuncType.pp ~level:0) ft
>       |> error_no_info
>   ...
> ```
> 
> Here, `Global` corresponds to the top-level, and no p4c test program has emitted an error here, supporting that we do not have a use case of a function/extern method/extern function call made at the top-level for now in the test suite.

jafingerhut:
> > @jafingerhut
> > > In your new proposed table, row "extern", new column "top level", you have "yes".
> > > Do you know the reason for this? Is it because in the current version, row "extern" and column "parser or top level" has "yes", and the "yes" was copied from there to the new "top level" column?
> > 
> > 
> > Yes, I've copied it from the _parser or control top level_ column from the table prior to this PR. But on second thought, like how we shifted to disallowing function calls at the top level at the discussion above, I think it would be conservative to disallow extern methods and function calls at the top level as well. Of course we may list exceptional cases where we may allow such as a footnote, if necessary.
> 
> The only exception I can think of that we definitely do wish to support and allow is: constructors are allowed to be called at the top level, to instantiate extern objects.

I would recommend that we change this PR so that for row "extern", new column "top level", we have "no" instead of "yes".  Also that it have a footnote that says "extern object construct calls are allowed at the top level, to instantiate extern objects at the top level".

jafingerhut:
> > > @jafingerhut
> > > > In your new proposed table, row "extern", new column "top level", you have "yes".
> > > > Do you know the reason for this? Is it because in the current version, row "extern" and column "parser or top level" has "yes", and the "yes" was copied from there to the new "top level" column?
> > > 
> > > 
> > > Yes, I've copied it from the _parser or control top level_ column from the table prior to this PR. But on second thought, like how we shifted to disallowing function calls at the top level at the discussion above, I think it would be conservative to disallow extern methods and function calls at the top level as well. Of course we may list exceptional cases where we may allow such as a footnote, if necessary.
> > 
> > 
> > The only exception I can think of that we definitely do wish to support and allow is: constructors are allowed to be called at the top level, to instantiate extern objects.
> 
> I would recommend that we change this PR so that for row "extern", new column "top level", we have "no" instead of "yes". Also that it have a footnote that says "extern object constructor calls are allowed at the top level, to instantiate extern objects at the top level".

@jaehyun1ee Do you agree with my "yes" to "no" change suggested above?  If so, could you please make that change?  I believe I'm ready to give this PR approval with such a change.

jaehyun1ee:
Sorry again for the late reply, 

> I would recommend that we change this PR so that for row "extern", new column "top level", we have "no" instead of "yes". Also that it have a footnote that says "extern object constructor calls are allowed at the top level, to instantiate extern objects at the top level".

I think this change would not be necessary, as the table under discussion is: **can be called at run time from this place in a P4 program**, and the constructor call site restrictions are defined in another table in the same section, **can be instantiated in this place**. Also in that table, extern constructors are allowed to be called at top level.

jaehyun1ee:
This is how the tables look like:

image

> The row for extern describes where extern method or extern function calls can be made from. The row for function describes where calls to functions, but not extern functions, can be made from.

image

jafingerhut (approved):
As of commit 5 of this PR, reviewed 2025-May-13, I approve of this PR.

jafingerhut:
@vlstill As of commit 5 of this PR on 2025-May-13, I approve of this PR.  If you are interested in taking a look at it to see if it meets your expectations, that would be great, since you had some coments/questions earlier.  I have not attempted to review your earlier comments to see whether they have been addressed, only mine.

jafingerhut:
@rcgoodfellow @jonathan-dilorenzo I have approved this, so I think it is good to go, but always good to have another set of eyes on such things before merging.   I do not know whether vlstill has time to review this again.

rcgoodfellow (approved):

PR #1358, opened 2025-01-21
[#1292] Explicitly state that table entries declared with 'const entries' cannot be assigned numeric priorities

### LINKED ISSUE #1292: Document restriction that numeric priorities may not be specified within `const entries`
p4c currently has this restriction, and while it could be generalized, there is no known practical reason to allow specifying priorities explicitly per-entry within `const entries`.  See https://github.com/p4lang/p4c/issues/4792 for discussion.

kfcripps:
Closes #1292 and closes https://github.com/p4lang/p4c/issues/4792.

As mentioned in https://github.com/p4lang/p4c/issues/4792, `p4c` does not support the `priority` keyword for entries of a table that has `const entries`.

Feel free to suggest better wording or a better place for this addition.

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -7087,6 +7087,12 @@ order, stopping at the first matching entry. Architectures should
 define the significance of entry order (if any) for other kinds of
 tables.
 
+Because control-plane APIs cannot insert or remove entries of a table
+that is declared with `const entries`, the relative priorities of such a
+table's entries are determined solely by the program order of the entries.
+Therefore assigning numeric priorities to entries of a table that has
+`const entries` is not allowed.
+
 Depending on the `match_kind` of the keys, key set expressions may define
 one or multiple entries. The compiler will synthesize the correct number of
 entries to be installed in the table. Target constraints may further restrict


kfcripps:
I cannot request reviews in this repository, so instead I am tagging @jafingerhut @vlstill @ChrisDodd

jafingerhut (approved):
Looks good to me.

jafingerhut:
@kfcripps Can you edit your original comment on this PR to mention that p4c's implementation does not support explicitly specifying priority values within `const entries`?  (assuming I am recalling that correctly).  I am not certain, but I believe that p4c _never_ supported explicit priorities in this case, because when explicit priority support was added, it was done only for `entries` without `const`.

That is the kind of supporting detail that can help when/if this comes up in discussion in the language design work group.

kfcripps:
> @kfcripps Can you edit your original comment on this PR to mention that p4c's implementation does not support explicitly specifying priority values within `const entries`? (assuming I am recalling that correctly). I am not certain, but I believe that p4c _never_ supported explicit priorities in this case, because when explicit priority support was added, it was done only for `entries` without `const`.
> 
> That is the kind of supporting detail that can help when/if this comes up in discussion in the language design work group.

Done. I don't know if `p4c` _never_ supported explicit priorities for entries of tables with `const entries`, but it did not support this as of when I opened https://github.com/p4lang/p4c/issues/4792.

I cannot merge this so feel free to merge if you are all happy with the changes.

jafingerhut:
Added ldwg-discussion label to this issue, to remind us to talk about it at the next LDWG meeting in February, if it has not been merged before then.

vlstill (approved):

PR #1359, opened 2025-01-30
Add op= assignments

ChrisDodd:
For #1144

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -5498,6 +5498,15 @@ types (e.g. ``struct``s) are copied recursively, and all components
 of ``header``s are copied, including "validity" bits. Assignment is
 not defined for `extern` values.
 
+An assignment may also be written with a binary arithmetic or bit manipulation
+operator immediately before the `=` sign.  This performs the binary operator on
+the old value of the left sub-expression and the right sub-expression and assigns
+the result to the l-value.  Thus an assignment like `A += B` is equivalent
+to `A = A + B`, except that `A` is only evaluated once.  This means that
+any side-effects within this operand (eg, a function call inside an array index
+or slice expression) only occur once.  This is not valid
+for comparison operators, logical operators, concat, range, or mask operators.
+
 [#sec-empty-stmt]
 === Empty statement
 
@@ -9045,6 +9054,8 @@ The P4 compiler should provide:
 == Revision History
 
 ===  Summary of changes made in unreleased version
+
+* Added compound assignment statements (<<sec-assignment>>)
 
 === Summary of changes made in version 1.2.5, released October 11, 2024
 
--- a/p4-16/spec/grammar.adoc
+++ b/p4-16/spec/grammar.adoc
@@ -736,7 +736,19 @@ typedefDeclaration
 assignmentOrMethodCallStatement
     : lvalue "(" argumentList ")" ";"
     | lvalue "<" typeArgumentList ">" "(" argumentList ")" ";"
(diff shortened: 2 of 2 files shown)


jafingerhut (approved):
LGTM, but probably best to let one of the co-chairs do final approval and merging, rather than merging with only my approval.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -5498,6 +5498,12 @@ types (e.g. ``struct``s) are copied recursively, and all components
 of ``header``s are copied, including "validity" bits. Assignment is
 not defined for `extern` values.
 
+An assignment may also be written with a binary operator immediately
+before the `=` sign.  This performs the binary operator on the old
+value of the left sub-expression and the right sub-expression and assigns
+the result to the l-value.  Thus an assignment like `A += B` is equivalent
+to `A = A + B`, except that `A` is only evaluated once.
```
vlstill, line 5505: I would explicitly mention that any side effects that may be in A are performed just once (since P4 does not have any concept of references, it may not be entirely clear what an lvalue is evaluated to).

p4-16/spec/P4-16-spec.adoc
```diff
@@ -5498,6 +5498,12 @@ types (e.g. ``struct``s) are copied recursively, and all components
 of ``header``s are copied, including "validity" bits. Assignment is
 not defined for `extern` values.
 
+An assignment may also be written with a binary operator immediately
+before the `=` sign.  This performs the binary operator on the old
```
vlstill, line 5502: This only works for selected binary operations and that should be clearly stated. While you have included all arithmetic operators as far as I can see, there are more binary operators that are not included:

- the boolean operators `&&` and `||` (note that this relates also to https://github.com/p4lang/p4c/pull/5108#discussion_r1936243149)
- the relational operators, which are also binary operators, but they return boolean, so it does not make sense (neither from type point not from usability point) to use them in compound operators.
- the `++` string operator (which needs not be included as there are no string variables, and even if there were they would still be `const`)
- the mask (`&&&`) and range (`..`) operators, again, these don't make sense here.

vlstill (dismissed):
The biggest question to me is what to do with booleans. And we should be careful to specify which operators apply here, not all binary operators are usable.

jafingerhut:
> The biggest question to me is what to do with booleans. And we should be careful to specify which operators apply here, not all binary operators are usable.

@vlstill Are there any remaining concerns you have that have not been addressed?  If so, please explain what they are, as the comments that I see appear to be addressed.

vlstill (commented):
I would probably just add that this is allowed for integral types (`bit<N>`, `int<N>`, `int`).

vlstill (approved):

jafingerhut:
@jonathan-dilorenzo Discussed at 2025-Mar-10 LDWG meeting, and vlstill approved during the meeting.  If you wish to do a final approval pass before merging, please do so, but we believe this is ready to merge into the spec.  FYI, the implementation of this was merged into p4c recently.

jonathan-dilorenzo:
@ChrisDodd, can I ask you to add a bullet point summarizing the change to [this section](https://p4.org/p4-spec/docs/p4-16-working-draft.html#_summary_of_changes_made_in_unreleased_version) before I merge?

jonathan-dilorenzo (approved):

ChrisDodd:
> @ChrisDodd, can I ask you to add a bullet point summarizing the change to [this section](https://p4.org/p4-spec/docs/p4-16-working-draft.html#_summary_of_changes_made_in_unreleased_version) before I merge?

done

PR #1360, opened 2025-01-30
Add arrays of non-header types

ChrisDodd:


jafingerhut:
In this section:
```
[#sec-uninitialized-values-and-writing-invalid-headers]
=== Reading uninitialized values and writing fields of invalid headers
```
there is an explicit statement about accessing elements of header stacks that are outside of range:
```
The result of reading a value in any of the situations below is that
some unspecified value will be used for that field.
[ ... ]
* reading a field of a header that is an element of a header stack,
  where the index is out of range for the header stack.
```
Should that be generalized to reading an element of an array where the index is out of range for the array?

Similarly later in that section for attempted writes to a header stack element it says:
```
If any of these kinds of writes are performed:
[ ... ]
* a write to a field in an element of a header stack, where the index
  is out of range
[ ... ]
then that write must not change any state that is currently defined in
the system, neither in header fields nor anywhere else.
```
Should that also be generalized to attempting to write to elements of an array where the index is out of bounds?

p4-16/spec/P4-16-spec.adoc
```diff
@@ -2569,6 +2581,14 @@ underlying type and representation for `enum` elements.
 used as a field in a `header` must contain only `bit<W>`, `int<W>`, a
 serializable `enum`, or a `bool`.
 
+[3] An array whose element type is a header or header union type is a
+header stack.
+
+[4] This is for array types that are not header stacks (the array element
+type is not a header or header_union)
+
+[5] Architectures may disallow arrays of arrays
```
vlstill, line 2590: nitpick:

```suggestion
[4] This is for array types that are not header stacks (the array element
type is not a header or header_union).

[5] Architectures may disallow arrays of arrays.

**Review by vlstill — COMMENTED** [review]
Looks good to me, I agree we should have exact semantics of out-of-bounds access as suggested in https://github.com/p4lang/p4-spec/pull/1360#issuecomment-2625601634.

**Comment by kfcripps** [comment]
This opens up the possibility of having nested arrays, which could potentially complicate array / header stack processing in some passes. We might consider explicitly prohibiting nested arrays until there is real motivation to support them. Related discussion: https://github.com/p4lang/p4c/pull/5115#discussion_r1936292442

**Comment by ChrisDodd** [comment]
Most of the support for this is already in p4c -- last remaining piece is https://github.com/p4lang/p4c/pull/5338

**Comment by jonathan-dilorenzo** [comment]
Ready for review and merging. @jonathan-dilorenzo and @jafingerhut to review.

**Comment by jonathan-dilorenzo** [comment]
Let's resolve the "error?" to error and make a note that this isn't done for any deep reason, but simply due to the lack of compelling use cases and the ability to change our minds in a non-breaking way.

File: p4-16/spec/P4-16-spec.adoc [hunk 2]
```diff
@@ -2398,22 +2400,30 @@ parsing steps.
...
+where `typeRef` refers to the element type and the expression is a compile-time
+known value that is the size of the array.  For an array type such as `arr[n]`,
+`n-1` is the maximum defined index.
+
+[#sec-header-stacks]
+==== Header stacks
+
+A header stack represents an array of headers or header unions. A header stack type is
+defined as an array declaration
+where `typeRef` is the name of a header or header union type. For a
```
jaehyun1ee, line 2422: Considering `typedef`s which may alias header or header union types, e.g.,

```p4
header Hdr { bit<32> x }
typedef Hdr hhh;
... and somewhere else we use hhh[5], a header stack
```

"where `typeRef` is the name of a header or header union type" is a bit imprecise.

So maybe we can put:

```
A header stack type is defined as an array declaration where `typeRef` represents 
```

p4-16/spec/P4-16-spec.adoc
```diff
@@ -2549,39 +2559,43 @@ arbitrary-precision integer, without a width specified.
...
+| `struct types`       | allowed [2] | error   |  allowed | allowed | allowed | error
 
-| `struct types`       | allowed [2] | error   |  allowed | allowed | error
+| `tuple types`        | error     | error   |  allowed | allowed | error [7] | error
 
-| `tuple types`        | error     | error   |  allowed | allowed | error
+| `list types`         | error     | error   |  error   | allowed | error [7] | error
 
-| `list types`         | error     | error   |  error   | allowed | error
+| `extern instances`   | error     | error   |  error   | erroe  | allowed [6] | error
```
jonathan-dilorenzo, line 2598: nit: error*

p4-16/spec/P4-16-spec.adoc
jonathan-dilorenzo: +1. I think that phrasing is quite nice.

jonathan-dilorenzo (approved):
Minor comments, but approving and will merge immediately upon fix.

Issue #974, opened 2021-10-22
Name duplication and name shadowing


MollyDream:
The p4 spec does not specify the behavior of name duplication (reusing names in the same scope) and name shadowing (reusing names in the inner scope) in P4. 
Litmus Test 1
```
void f1() {
  int x;
  int x;
}
```
Litmus Test 2
```
void f2() {
  int x;
  { int x; }
}
```
Litmus Test 3
```
void f3(inout bit<8> x) {
  bit<8> x;
}
```
Litmus Test 4
```
void f4(inout bit<8> x) {
  { bit<8> x; }
}
```
Since there are no definitions in P4, we have to resort to a similar language for reference, namely C. For the examples above, test 1 and 3 are considered name duplication, and 2 and 4 considered name shadowing in C. Also in C, both the test 1 and 3 are invalid, and both 2 and 4 are valid.
An old issue report of p4c (https://github.com/p4lang/p4c/issues/1932) states that name duplication is generally not allowed, and the only exceptions are overloading, which, based on the [current spec](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec_extern), is now restrained to only functions and methods. (That post is outdated in that overloading is not allowed on any other types, so controls, actions, parsers and packages should not share the same names and the P4c compiler should reject those test cases.) 
However, there are p4c test examples similar to test 3 that are allowed, like [shadow-after-use.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/shadow-after-use.p4) and [shadow3.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/shadow3.p4). The naming of these examples already implies that P4 is right now regarding test 3 as shadowing, which differs from C. Personally I think the two `x`'s in test 3 are in the same scope, so I think it is more likely name duplication. I think it would be valuable to reevaluate this design decision and to clarify the definitions of name duplication in the spec.
Also, it is unclear if there are any constraints on name shadowing in P4, though it seems that there should be none. For example, name shadowing of two different types should also be valid.

MollyDream:
@mbudiu-bfn Do you have any comments on this? Based on the post, I think the spec should
- Clarify the definition of name duplication and name shadowing
- State clearly the situations (overloading) where name duplication is allowed
- State clearly the situations (everywhere?) where name shadowing is allowed

[...]

PR #1376, opened 2025-11-28
Disallow duplicate declarations within the same scope

jaehyun1ee:
This PR revises the specification to disallow duplicate declarations within the *same* scope. This has been discussed in P4 LDWG.

* It disallows duplicate declarations, e.g.,

```p4
control p() {
    const bit<8> x = 8;
    const bit<4> x = 4;   // Error: duplicate declaration of x
    apply {}
}
```

* But it still allows declaration shadowing (same identifier in an *inner* scope), e.g.,

```p4
const bit<4> x = 1;
control p() {
    const bit<8> x = 8;   // x declaration shadows global x
    apply {}
}
```

* And it still allows functions and methods with the same names (up to overload resolution). Recall that P4 allows overloading of functions and methods, when they can be differentiated by either the number of parameters or the names of parameters:

```p4
extern void f(in bit<8> x);
extern void f(in bit<8> y); // allowed: different parameter names
extern void f(in bit<16> x, in bit<8> y); // allowed: different number of parameters
extern void f(in bit<16> x, in bit<8> x); // Error: duplicate declaration of previous function
```
Some related past issues in p4-spec and p4c are:

* p4lang/p4-spec#974
  * This is perhaps the first issue asking for clarification regarding name duplication. Was closed after it got stale.
* p4lang/p4c#5096
  * The recent p4c issue, which led to a discussion in 2025-November P4 LDWG meeting. The consensus was to disallow duplicate declarations within the same scope.

In particular, this PR modifies sections **6.6.1. Scopes**, **6.9. Name Resolution**, and **7.2.11.2. Extern Objects**.

* **6.6.1** add that duplicate declarations are disallowed.
* **6.9** adds a detailed explanation of duplicate declaration, and also explains that name shadowing (up to compiler warning) and function/method overloading are still allowed.
* The part explaining function/method overloading in **7.2.11.2** was moved to **6.9**.

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -1361,6 +1361,9 @@ declaration. (This is a departure from P4~14~, which
 allows declarations in any order. This requirement significantly simplifies the implementation of
 compilers for P4, allowing compilers to use additional information
 about declared identifiers to resolve ambiguities.)
+Except for functions and methods, which may be overloaded, duplicate
+declarations of the same name within the same scope are not allowed.
+<<sec-name-resolution>> describes how names are resolved in P4.
 
 [#sec-stateful-elems]
 ==== Stateful elements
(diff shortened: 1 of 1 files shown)


ajwalton:
Thanks @rcgoodfellow 
It would seem that this text still doesn't resolve the meaning of "X a = a", where "a" is also defined in an out scope.  (@ChrisDodd suggested that the declaration is for the whole scope, so this is really setting "a" to itself, which is a use before definition, not the value from the outer scope.)
Does this also make clear whether parameters are in the same scope as the enclosing block?

jafingerhut:
@jafingerhut Review this.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -1769,6 +1772,52 @@ control p() {
...
+which method or function is being called, either by the number of arguments or
+by the names of the arguments, when calls are specifying argument names.
+Argument type information is _not_ used in disambiguating calls.
+
+[source,p4]
+----
+extern void f(in bit<8> x);
+extern void f(in bit<8> y); // allowed: different parameter names
+extern void f(in bit<16> x, in bit<8> y); // allowed: different number of parameters
+extern void f(in bit<16> x, in bit<8> x); // Error: duplicate declaration of previous function
```
jafingerhut: I am confused by this example for a couple of reasons.

(1) There are two parameters named `x`.  That seems like the primary reason that it would be an error to me, regardless of overloading considerations.

(2) If we changed the parameters names to 'x' and 'z' instead of 'x' and 'x', would it be legal?  The _sequence of parameter names_ is different, so it is not clear to me whether that is sufficient to be similar to the example a couple of lines above where it has the comment "allowed: different parameter names".  Does each individual parameter name need to be different?  Or does the entire _sequence_ of names need to differ in at least one of the names?

(3) I had not realized (or perhaps seen it before, but forgotten) that the example with the comment "allowed: different parameter names" was allowed.  Is it expected that calls to such an overloaded function must use the parameter name in calls, in order to distinguish which instance of `f` is meant?  e.g. `f(x = 5)` or `f(y=a+b)` are legal, but `f(5)` is not?

jaehyun1ee: Thank you for pointing it out, it was actually a typo, where I fixed the example to:

```p4
extern void f(in bit<8> x);
extern void f(in bit<8> y); // allowed: different parameter names
extern void f(in bit<16> x, in bit<8> y); // allowed: different number of parameters
extern void f(in bit<16> x, in bit<8> y); // Error: duplicate declaration of previous function
```

So I believe this resolves (1) and (2), sorry for making the confusion. And yes, declaration such as `extern void f(in bit<16> x, in bit<8> x);` is already invalid because it has duplicate parameter names.

For (3), in P4, overload resolution occurs via parameter names and the number of parameters. So, `extern void f(in bit<8> x);` and `extern void f(in bit<8> y);` are overloaded, and the caller MUST explicitly disambiguate by specifying either `x=...` or `y=...` in the argument. It is an error if they cannot be distinguished, as per the spec:

> When overloading is used, the compiler must be able to disambiguate at compile-time which method or function is being called, either by the number of arguments or by the names of the arguments, when calls are specifying argument names. (7.2.11.2)

jafingerhut (approved):
LGTM

Issue #1377, opened 2026-01-21
Table in section "Type nesting rules" is missing 1 column in each of last 4 rows


jafingerhut:
This is the first table in the section with title "Type nesting rules".

In particular, each of these last 4 lines in the table has only 6 vertical bars, whereas all earlier rows of that table have 7.
```
| `extern types`       | error     | error   |  error   | error   | error

| `parser types`       | error     | error   |  error   | error   | error

| `control types`      | error     | error   |  error   | error   | error

| `package types`      | error     | error   |  error   | error   | error
```
When running asciidoctor using the included Makefile in the p4-16/spec directory, it gives this output:
```
$ make
./trim-asciidoc-tag-comments.py grammar.adoc > grammar.trimmed.adoc
time asciidoctor-pdf -v \
	-r asciidoctor-mathematical \
	-a pdf-fontsdir=resources/fonts \
	-a rouge-style=github P4-16-spec.adoc
asciidoctor: ERROR: P4-16-spec.adoc: line 2606: dropping cells from incomplete row detected end of table
23.21user 0.34system 0:23.51elapsed 100%CPU (0avgtext+0avgdata 369176maxresident)k
0inputs+10792outputs (0major+119635minor)pagefaults 0swaps
time asciidoctor -v \
	-r asciidoctor-mathematical \
	-a rouge-css=style P4-16-spec.adoc
asciidoctor: ERROR: P4-16-spec.adoc: line 2606: dropping cells from incomplete row detected end of table
0.46user 0.02system 0:00.48elapsed 100%CPU (0avgtext+0avgdata 78360maxresident)k
0inputs+5704outputs (0major+15052minor)pagefaults 0swaps
```

jafingerhut:
The relevant lines were added in this commit: https://github.com/p4lang/p4-spec/commit/ba0f0cfc2f0cec5247b88245e07d0340293746d2

@jaehyun1ee Could you take a look at those lines and see which entries you think need to be added?

ChrisDodd:
Probably an interaction/mis-merge with [`4517dafb`](https://github.com/p4lang/p4-spec/commit/4517dafb8183695149184590b6a4889526d78b40)

jaehyun1ee:
@jafingerhut Thank you for the ping, I agree with the changes made in the above PR, where `extern`s can be nested inside an array, but not for `parser`s, `control`s, and `package`s.

ChrisDodd:
Can we run this linting check as part of the CI rules?

[...]

PR #1378, opened 2026-01-21
fix type/container table columns

ChrisDodd:
Fix for #1377

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -2557,7 +2557,7 @@ arbitrary-precision integer, without a width specified.
 
 [.center,%autowidth]
 |===
-| Element type 5+^| Container kind  
+| Element type 6+^| Container kind
 
 |                | header    | header_union | struct or tuple | list | array [4] | header stack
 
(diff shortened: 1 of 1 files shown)


p4-16/spec/P4-16-spec.adoc
```diff
@@ -2557,7 +2557,7 @@ arbitrary-precision integer, without a width specified.
 
 [.center,%autowidth]
 |===
-| Element type 5+^| Container kind  
+| Element type 5+^| Container kind
```
jaehyun1ee: This should be `6+` instead of `5+`, since we now have six columns.

p4-16/spec/P4-16-spec.adoc
```diff
@@ -2595,15 +2595,13 @@ arbitrary-precision integer, without a width specified.
 
 | `list types`         | error     | error   |  error   | allowed | error [7] | error
 
-| `extern instances`   | error     | error   |  error   | erroe  | allowed [6] | error
+| `extern instances`   | error     | error   |  error   | error  | allowed [6] | error
```
jaehyun1ee: Because the column header is **Element type**, I think `extern types` better suits as the row title. Can we rename it to `extern types`?

ChrisDodd: I'm kind of mixed on this, though the meaning is the same.  The thing is that extern types (and control, parser, and package types) are not like other types.  You can't have values of these types, you can only instantiate them (which requires an explicit constructor call) and pass the instances as directionless args to things.  But we should at least be consistent with those 4.

jaehyun1ee: I see the point, thank you for the edits :)

jafingerhut (approved):
LGTM

jafingerhut:
@jonathan-dilorenzo @rcgoodfellow This is not urgent, but a good small fix to consider merging in, if you approve.

rcgoodfellow (approved):

jonathan-dilorenzo (approved):
LGTM. Will presume that Extern instances and extern types are not separately interesting for type-nesting rules. Which seems correct.

PR #1381, opened 2026-02-02
Clarify compile time constant requirements for directionless action p…

### LINKED ISSUE #1379: Need more Clarification in Section 6.8 that runtime values for directionless parameters apply only to call statements, not table properties
https://github.com/p4lang/p4c/issues/5042
The current text states that directionless parameters "need not be a compile-time known value" when provided by the P4 program. However, it does not distinguish between two different contexts:                                                                                                
                                                                                                                                                  
  Action call statements should allow runtime values, as the P4 program directly controls execution.                     
                                                                                                                                                  
  Table properties should require compile-time constants, as these values may be installed or modified by the control plane at runtime

### LINKED ISSUE #5042: Directionless action arguments provided by the P4 program should behave like `in`
The spec mentions in section 6.8. Calling convention that,

> For an action, a directionless parameter indicates that it is "action data". See [Section 14.1](https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-actions) for the meaning of action data, but its meaning includes the following possibilities:
>    - The parameter’s value is provided in the P4 program. In this case, the parameter behaves as if the direction were in. Such an argument expression need not be a compile-time known value.
>    - The parameter’s value is provided by the control plane software when an entry is added to a table that uses that action. See [Section 14.1](https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-actions).

But test [issue473.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_errors/issue473.p4) is categorized under `p4_16_errors`, expecting the frontend to reject the program with:

```p4
control cIngress(inout Parsed_packet hdr,
                 inout mystruct1 meta,
                 inout standard_metadata_t stdmeta) {
    action a1(in bit<32> x) {
        meta.b = meta.b + (bit<4>) x;
        meta.b = (bit<4>) ((bit<32>) meta.b + x);
    }
    action b(inout bit<32> x, bit<8> data) {
        meta.a = meta.a ^ (bit<4>) x ^ (bit<4>) data;
    }
    table t1 {
        key = { hdr.ethernet.srcAddr : exact; }
        actions = {
            a1((bit<32>) meta.a);
            b(meta.c);
        }
        default_action = b(meta.c, (bit<8>) meta.d); // HERE
    }

    apply {
        t1.apply();
    }
}
```

```plaintext
issue473.p4(70): [--Werror=type-error] error: (bit<8>)meta.d: action argument must be a compile-time constant
        default_action = b(meta.c, (bit<8>) meta.d);
                                   ^^^^^^^^^^^^^^^
```

I believe this is *not* in line with the current spec, since `b(meta.c, (bit<8>) meta.d)` is written within a P4 program (not by the control plane), thus the directionless argument `(bit<8>) meta.d` need *not* be a compile-time known value.

devalgupta404:
PR clarifies Section 6.8 to distinguish between Implicit invocations(Added that table properties require compile time constant values) and Explicit invocations (need not be compile time constant values).
 Closes https://github.com/p4lang/p4-spec/issues/1379                                                                             
 Closes https://github.com/p4lang/p4c/issues/5042

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -6810,12 +6810,14 @@ statements or expressions.
 
 Actions can be executed in two ways:
 
-- Implicitly: by tables during match-action processing.
-- Explicitly: either from a `control` block or from another `action`.
-  In either case, the values for all action parameters
-  must be supplied explicitly, including values for the directionless
-  parameters. In this case, the directionless parameters behave like `in`
-  parameters.
+- Implicitly: by tables during match-action processing. For action
+  invocations specified via table properties (such as `default_action`
+  or `entries`), directionless parameters must be compile-time constant
+  values, as these arguments are supplied by the control plane.
+- Explicitly: via a call statement, either from a `control` block or from another `action`.
+  In this case, the values for all action parameters must be supplied explicitly,
+  including values for directionless parameters. Directionless parameters
+  behave like `in` parameters and need not be compile-time constant values.
 
 [#sec-tables]
 === Tables


devalgupta404:
@jonathan-dilorenzo  can you please look into this?

fruffy:
@devalgupta404 We often discuss issues like this at the Language-Design Working Group.

You can see the schedule here: https://p4.org/working-groups/

PR #1385, opened 2026-03-01
Make spec builds fail if asciidoctor issues ERROR messages

jafingerhut:

PR #1386, opened 2026-03-01
Fix Asciidoctor error in P4-16 spec

jafingerhut:
Also modify Makefile so that if an Asciidoctor error occurs, the Github CI build will fail.

### DIFF
--- a/p4-16/psa/Makefile
+++ b/p4-16/psa/Makefile
@@ -8,17 +8,22 @@ all: ${SPEC}.pdf ${SPEC}.html charter
 
 ${SPEC}.pdf: ${SPEC}.adoc
 	asciidoctor-pdf -v \
+		--failure-level ERROR \
 		-a pdf-fontsdir=resources/fonts \
 		-a rouge-style=$(ROUGE_STYLE) $<
 
 ${SPEC}.html: ${SPEC}.adoc
-	asciidoctor -v -a rouge-css=$(ROUGE_CSS) $<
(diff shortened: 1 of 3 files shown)


jafingerhut:
@jaehyun1ee @ChrisDodd If you get a bit of time, would be excellent if you could review the spec changes in this PR.  I've added a few missing entries to the type nesting table, due to the recent addition of the "arrays" column to that table.

jafingerhut:
This PR addresses https://github.com/p4lang/p4-spec/issues/1377

It also modifies the asciidoctor commands so that they have non-0 exit status, and fail the Github CI build, if asciidoctor issues an ERROR message.  See https://github.com/p4lang/p4-spec/pull/1385 failed CI results for verification that the Makefile changes proposed in this PR (the same as that failing PR) do cause Github CI to fail if asciidoctor issues ERROR messages.

ChrisDodd (approved):
Looks like this subsumes #1378

jafingerhut:
> Looks like this subsumes #1378

Sorry, I lost track of your PR mentally.  I approved it a moment ago.  I am happy to have that one merged first, and then I can update this one to keep whatever additional changes remain.

jaehyun1ee (approved):
LGTM!

jafingerhut:
Sorry for any confusion.  Closing this PR in favor of Chris Dodd's earlier #1378, which appears to have an additional change that mine does not in the spec, so would be good to preserve that detail.

Issue #1390, opened 2026-04-06
Specify behavior of `+:` slice used as l-value


pacokwon:
The current working draft of the specification does not specify the behavior of `+:` style slices when they are used as l-values.

Here is the relevant description of `+:` slices in the specification.

> A slice may also be specified as [L+:W] where L is the lowest bit of the slice and W is the width of the slice. In this case, only W must be a non-negative local compile-time known numeric value. L must also be a numeric value and some architectures may require it to be compile-time known. It must also be non-negative and in-range for the type. Some architectures may allow variable indexing, in which case an out of range value will be equivalent to a shift of that amount. An index that is not compile-time known must be a bit<W> type in order to ensure that it is non-negative.

Earlier in the same paragraph in the specification is the behavior of `H:L` slices used as l-values described as follows:

>   Slices are also l-values, which means that P4 supports assigning to a slice: e[H:L] = x. The effect of this statement is to set bits H through L (inclusive of both) of e to the bit-pattern represented by x, and leaves all other bits of e unchanged.

Unlike `H:L` slices, the range referenced by `+:` could potentially go out of bounds. Hence a clear description of how it behaves in such edge cases also seems necessary.

jaehyun1ee:
Closing as #1391 is merged.

PR #1391, opened 2026-04-08
Specify behavior of +: slice used as l-value

pacokwon:
This PR adds a description of the behavior of `+:` slices when they are used as l-values.

Unlike `:` style slices that check the bounds of both `H` and `L` at compile time, `+:` slices can potentially reference an out of bounds range. While the behavior of `+:` slices with an *in bounds* range is straightforward, the behavior of the *out of bounds* range case deserves explicit clarification. However, neither of them are currently described by the specification.

There has been prior discussion on how such cases should be handled.

> We should define semantics of out-of-bounds slices. Similar to how we have it defined for header stack indexing, we should not leave it to the compiler.
> * we might consider the "out-of-range" bits to be undefined
> * or we might want then to be 0s
> * we probably want write to out-of-bounds bits to have no effect.

https://github.com/p4lang/p4-spec/pull/1316#discussion_r1939466081

This PR adopts the interpretation that out-of-bounds bits have no effect, or are discarded, when `+:` out-of-bounds slices are used as l-values.

Related Issue: https://github.com/p4lang/p4-spec/issues/1390

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -3674,6 +3674,12 @@ Bit-strings also support the following operations:
   in which case an out of range value will be equivalent to a shift of that amount.
   An index that is not compile-time known must be a `bit<W>` type in order to ensure that
   it is non-negative.
+  Slices specified with `[L+:W]` are also l-values, which means that P4 supports
+  assigning to such a slice: `e[L+:W] = x`.
+  The effect of this statement is to set `W` bits starting from bit `L` of `e`
+  to the bit-pattern represented by `x`, and leaves all other bits of `e` unchanged.
+  If `L+W` exceeds the width of `e`, then only the bits that fall within the valid
+  range of `e` are modified.
 * Concatenation of bit-strings and/or fixed-width signed integers, denoted by `++`.
   The two operands must be either `bit<W>` or `int<W>`, and they can be of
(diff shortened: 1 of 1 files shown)


ChrisDodd:
The spec does say that the behavior of reading out-of-bounds bits is the same as shifting -- so a variable index slice (read) can always be replaced by a fixed slice [width-1:0] on top of a shift.  That means they'll be 0 on a `bit` type, and duplicate the sign bit on an `int` type.

jafingerhut (approved):
This looks like the only "safe" specification of using this style of slice as an L-value that I can think of.

pacokwon:
> The spec does say that the behavior of reading out-of-bounds bits is the same as shifting -- so a variable index slice (read) can always be replaced by a fixed slice [width-1:0] on top of a shift. That means they'll be 0 on a `bit` type, and duplicate the sign bit on an `int` type.

Good point! My bad for not being clear enough, but the comments in the PR are specifically for when `+:` slices are being used as l-values. The changes in the PR add descriptions about the behavior of `+:` slices as l-values, and how the out-of-bounds edge case should be handled.

jonathan-dilorenzo:
@jafingerhut and @rcgoodfellow, I have a vague recollection that we discussed this at some point and decided to merge it. Since it's approved and looks like a sensible definition, will go ahead and do that, but if you have concerns we can address them in a follow-up.

PR #1392, opened 2026-04-13
Update P4 Spec with group_actions

matthewtlam:
Discussion from https://github.com/p4lang/p4runtime/issues/605, stemmed updating the Language spec with information regarding the semantics of group actions, which was introduced in P4Runtime in https://github.com/p4lang/p4runtime/pull/594 and https://github.com/p4lang/p4runtime/pull/604.

cc: @jonathan-dilorenzo, @smolkaj, @jafingerhut, @chrispsommers, @kheradmandG

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -8995,9 +8995,10 @@ The following three annotations can be used to give additional
 information to the compiler and control-plane about actions in a
 table. These annotations have no bodies.
 
-Note that with no annotation, the action can be used as (a) the action of any number of
-table entries, or (b) the table's defalt action, or both (a) and (b) simultaneously,
-but cannot be configured as (c) the action of any number of the action selector's groups.
+Note that with no annotation, the action can be used as (a) the action of any
+number of table entries, or (b) the table's defalt action, or both (a) and (b)
+simultaneously, but cannot be configured as (c) the action of any number of
+the action selector's groups.
 
 - `@tableonly`: actions with this annotation can only appear
   within the table, but never as default nor the group action.
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -8991,14 +8991,21 @@ target-dependent.
 [#sec-table-action-anno]
 ==== Annotations on the table action list
 
-The following two annotations can be used to give additional
+The following three annotations can be used to give additional
 information to the compiler and control-plane about actions in a
 table. These annotations have no bodies.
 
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):

rcgoodfellow:
LDWG notes: move this to a PR on the PSA spec since groups are not a part of the language itself.

jafingerhut:
@matthewtlam My apologies for not thinking of this earlier, but during the last language design work group meeting a comment was made about this proposed change that I think is valid: action selectors and action profiles are defined in some (but not all) P4 _architecture_ definitions.  They do not exist in the base language.

Thus it would make more sense to define the `@groupaction` annotation in an architecture specification, e.g. the PSA specification.  I can help with that, since I noticed that there isn't already a section in the PSA speciication with PSA-specific annotations.  At least there is not yet, but we can add one for this.

Once it is part of the SPA specification, any other architectures where it makes sense to define it (e.g. v1model) we can update their documentation to point at the definition in the PSA specification, if we want it to be the same definition (which I think in all cases for architectures "owned" by P4.org we would want it to be the same).

jafingerhut:
@matthewtlam I have looked through the PSA specifcation, and I think a good idea would be to add new text into Section 8.12 "Action Selector".  There currently is no separate section in the document mentioning all annotations defined in PSA, but we could add a new appendix, just before the last appendix "Revision History", with the new appendix titled "PSA Reserved Annotations", with contents similar to the appendix titled "P4 Reserved Annotations" in the P4 language spec [1].

[1] https://github.com/p4lang/p4-spec/blob/main/p4-16/spec/P4-16-spec.adoc#sec-p4-reserved-annotations

sanket-jadhav-cse:
Hi @matthewtlam, @jafingerhut, and @smolkaj,

I was looking through the discussion, and it seems like the plan is to move this documentation from the P4 language spec to the PSA spec, since action selectors and groups are architecture-specific.

I would like to work on the PSA spec by adding the `@groupaction` definition and explaining its semantics in Section 8.12, "Action Selector". I can also add a "PSA Reserved Annotations" appendix at the end if that is still a good approach.

Has anyone already started working on this? If not, can I take it up?

Thanks.

jafingerhut:
> Has anyone already started working on this? If not, can I take it up?

I am not aware of anyone working on this yet.  Please go ahead.

jafingerhut:
This alternative PR https://github.com/p4lang/p4-spec/pull/1417 making a similar change in the Portable Switch Architecture (PSA) specification was merged today.

For the reasons mentioned in earlier comments, no change is considered necessary or even desirable in the P4 language specification document for this topic, so closing this PR.

PR #1393, opened 2026-04-14
Clarify the equivalence of actions in default actions

### LINKED ISSUE #1388: To what extent do we check equivalence of table actions?
P4 tables may specify the `default_action` property where **Section 14.2.1.3. Default action** of the official spec states:

> If present, the default_action property must appear after the action property. It may be declared as const, indicating that it cannot be changed dynamically by the control-plane. The default action must be one of the actions that appear in the actions list.

The unclear part here is, how we should check the last part, *The default action must be one of the actions that appear in the actions list*. *"one of the actions"* is unclear, of whether it means syntactic equivalence or semantic equivalence. Examples are:

```p4
action a() {}
control c() {
    table t {
        actions = { .a; }
        default_action = a; /* semantically, a is the same as .a here, but not syntactically */
    }
    apply {}
}
```

```p4
control c()
{
  bit<2> tt;
  action a(in bit v1, in bit v2) { tt = v1++v2; }
  action a1(in s v) { tt = v.f0++v.f1; }
  table t
  {
    actions = { a1({ f0 = ext0(), f1 = ext1()}); }
    default_action = a1({ f1 = ext1(), f0 = ext0()}); /* also, semantically the same arguments, but not syntactically */
  }
  apply {}
}
```

For the latter case, it may be problematic if `ext0()` and `ext1()` is side-effecting, as the order of which they are called affects the program semantics. 

It is probably too onerous to precisely state what *the same* is in the spec, but at least there is an unclear part in the current spec we might want to discuss further.

Below are related tests in the `p4c` test suite:

* [p4c/testdata/p4_16_samples/issue2037.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2037.p4)
* [p4c/testdata/p4_16_samples/issue3671.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue3671.p4)

shyukahn:
Closes #1388
This PR is made from the April 2026 P4 LDWG meeting.

A `default_action` of a table must be one of the actions that appear in the `actions` list. The PR clarifies how to determine the equivalence of the default action and an action in the `actions` list.

- The name of the default action must be same as the action of the `actions` list, after resolving them both to full hierarchical names.
- The expressions passed as `in`, `out`, or `inout` parameters must be syntactically identical to the expressions used in the action of the `actions` list. A P4 implementation may choose to accept default actions  with arguments that are semantically equivalent to the expressions in the `actions` list, instead of requiring them to be syntacitcally identical.

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -7093,10 +7093,16 @@ not find a match for the supplied key.
 If present, the `default_action` property _must_ appear after the `action`
 property. It may be declared as `const`, indicating that it cannot
 be changed dynamically by the control-plane. The `default action`
-_must_ be one of the actions that appear in the actions list. In
-particular, the expressions passed as `in`, `out`, or `inout`
-parameters must be syntactically identical to the expressions used in
-one of the elements of the `actions` list.
+_must_ be one of the actions that appear in the actions list. That is,
+the default action must satisfy the following:
+
+- The name of the default action must be same as the action of the `actions`
+  list, after resolving them both to full hierarchical names.
+- The expressions passed as `in`, `out`, or `inout` parameters must be
+  syntactically identical to the expressions used in the action of the
+  `actions` list. A P4 implementation may choose to accept default actions 
+  with arguments that are semantically equivalent to the expressions in the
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):
LGTM

PR #1395, opened 2026-04-21
Add copyright and license attributions to many files

jafingerhut:


jafingerhut:
This is still incomplete, but it is a step towards adding copyright and license information to every file in the p4-spec repository, in hopes that in the not too distant future this can be completed for all files in this repository, and then we can use an open source tool recommended by the Linux Foundation called REUSE [1] that can be set up to automatically check all PRs pre-commit to ensure that any new files added to this repository also have copyright and license attributions, in a way that can be automatically checked (for their presence, at least, if not for correctness -- correctness requires human judgement).

Properties:

+ An open source reuse lint command can be used in CI to validate that every file in the repository has copyright and license information. This PR by itself does not get to that point, but it is a step in that direction.
+ Files that support comments have their copyright and license information added into comments inside the file, with specific supported syntax like "SPDX-License-Identifier: Apache-2.0" to specify the license the file is released under.
+ Files that do not support comments inside of them, e.g. the binary TrueType .ttf font files in the p4-spec repo, have a separate file named `<filename>.license` added that contains the copyright and license information.
+ All license files are stored in a top level directory named LICENSES, one license per file, with standard names for common licenses tracked by SPDX at https://spdx.org/licenses/
+ There is a `reuse annotate` command that can be used to add this information for one file, or many files at once, similar to Google's `addlicense` command in its features. The Bash script `tools/reuse-annotate-font-files.sh` in this PR demonstrates how this was done for the font files in the repository.

[1] https://reuse.software

rcgoodfellow (approved):
LGTM. Let's get a CI job as a follow up that uses the resue tool to check this so it does not get out of date.

Issue #1384, opened 2026-02-26
Dead spec links


fruffy:
Looks like the links to the specification and working draft on the README are dead. 

Or rather, there is no link to the working draft either.

jafingerhut:
I think #1398 may have fixed this, but leaving this issue open for now until issue creator confirms.

fruffy:
Maybe we should also link to the working draft in our README. @Devansh-567

PR #1398, opened 2026-05-09
Fix broken P4 specification link in README

Devansh-567:
Fixes broken link referenced in issue #1384.

Updated the README to point to the current valid P4 specification URL.


jafingerhut (approved):
LGTM

PR #1403, opened 2026-05-11
fix: add missing vars step and remove redundant -a flag in deploy job

Devansh-567:
The deploy job referenced `steps.vars.outputs.sha_short` but had no step with `id: vars` defined, causing every deploy commit message to read "docs for " with an empty SHA 

Also removed the `-a` flag from `git commit` since all files are
already explicitly staged by the preceding `git add`.

### DIFF
--- a/.github/workflows/asciidoc-build.yml
+++ b/.github/workflows/asciidoc-build.yml
@@ -96,6 +96,9 @@ jobs:
         name: api-charter
         path: api-charter
 
+    - name: Set short SHA
+      id: vars
+      run: echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
 
     - name: Copy files
       run: |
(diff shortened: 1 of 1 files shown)


jafingerhut:
I know this isn't something that you commented on or attempted to change, but one thing that I believe is true about the release scripts in this repository is that on every change to the specification source code, the new versions of generated PDF and HTML files are added to the repository history, causing the size of the .git directory to grow fairly quickly across many versions.  The p4runtime repository I believe does something different so that only the most recent version of generated PDF and HTML files are stored in the .git files.  If you could determine what the difference is there, and modify the p4-spec repo scripts to do the same, it would help eliminate that fairly rapid increase in p4-spec repo size over time.

Devansh-567:
I looked at the p4runtime repository's main-branch-uploads.yml. The key difference is that it uses EndBug/add-and-commit@v9 with commit: '--amend' and push: 'origin gh-pages --force', which amends the previous commit on gh-pages rather than adding a new one on every deploy. This keeps the gh-pages branch at a single commit permanently, preventing unbounded git history growth. I can implement the same approach in p4-spec's deploy job if that would be helpful.

jafingerhut:
> I can implement the same approach in p4-spec's deploy job if that would be helpful.

That would be wonderful!  If you are willing, please do so.

Devansh-567:
Hi @jafingerhut, just checking in on this PR. The gh-pages optimization discussed here has since been implemented in #1408. Whenever you get a chance, I'd appreciate a review. Thanks!

PR #1408, opened 2026-05-15
fix: use amend commit on gh-pages to prevent unbounded repo growth

Devansh-567:
Currently the deploy job adds a new commit to `gh-pages` on every specchange, causing the `.git` directory to grow unboundedly over time.
This adopts the same approach used in the `p4runtime` repository:
`EndBug/add-and-commit@v9` with `commit: '--amend'` and force push, keeping `gh-pages` permanently at a single commit regardless of how many times the spec is regenerated.

Also simplifies the deploy job by removing the manual git rebase steps, which are no longer needed since the action handles branch management internally.

Suggested by @jafingerhut in #1403.

### DIFF
--- a/.github/workflows/asciidoc-build.yml
+++ b/.github/workflows/asciidoc-build.yml
@@ -66,21 +66,8 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       with:
-        # Fetch all history for all branches
-        fetch-depth: 0
-
-    - name: Configure git
-      run: |
-        git config user.name "GitHub Actions"
-        git config user.email "p4lang@users.noreply.github.com"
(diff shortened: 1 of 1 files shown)


jafingerhut:
Thanks for the submission.  It would take me a fair amount of learning about Github Actions to review this carefully, which I may do if no one else is able to give this a proper review in the next week or two.

Devansh-567:
Hi @jafingerhut, just checking in on this PR. I know GitHub Actions changes can take some time to review, so no rush. Whenever you or someone else gets a chance to look at it, I'd be happy to address any feedback or make any needed changes. Thanks!

PR #1410, opened 2026-05-22
fix: replace Docker build with native asciidoctor install in any-branch-uploads

### LINKED ISSUE #1404: Inconsistent build strategies across CI workflows
- asciidoc-build.yml uses native asciidoctor install via install-asciidoctor-linux.sh
- any-branch-uploads.yml uses a Docker image p4lang/p4-spec-asciidoc:latest
- This means two different environments are building the same specs, potential for inconsistent output
- Propose unifying around one approach (Docker is the better pick, portable, reproducible)

Devansh-567:
Closes #1404

The any-branch-uploads workflow used a Docker image
p4lang/p4-spec-asciidoc:latest whose source is not publicly visible,
making the build environment difficult to audit.

This replaces the Docker-based build with the same native
install-asciidoctor-linux.sh approach used in asciidoc-build.yml,
making both workflows consistent and fully auditable.

Also updates actions/checkout from v3 to v4 and pins
ubuntu-latest to ubuntu-24.04 for reproducibility.

### DIFF
--- a/.github/workflows/any-branch-uploads.yml
+++ b/.github/workflows/any-branch-uploads.yml
@@ -1,21 +1,23 @@
 # SPDX-FileCopyrightText: 2024 The P4 Language Consortium
 #
 # SPDX-License-Identifier: Apache-2.0
-
 name: Any branch uploads
-
 on: workflow_dispatch
-
 jobs:
   any-branch-uploads:
     if: ${{ github.repository == 'p4lang/p4-spec' }}
-    runs-on: [ubuntu-latest]
+    runs-on: ubuntu-24.04
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Install asciidoc
+      run: sudo ./p4-16/spec/install-asciidoctor-linux.sh
     - name: Build spec
       run: |
-        docker run -v `pwd`/docs/v1:/usr/src/p4-spec p4lang/p4-spec-asciidoc:latest make
-        ls docs/v1/build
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):
LGTM

jafingerhut:
Out of curiosity, does the p4lang/p4runtime repository use a different docker image than the p4-spec repo is, or is it a different one?  https://github.com/p4lang/p4runtime/blob/main/.github/workflows/main-branch-uploads.yml#L20

Devansh-567:
> Out of curiosity, does the p4lang/p4runtime repository use a different docker image than the p4-spec repo is, or is it a different one? https://github.com/p4lang/p4runtime/blob/main/.github/workflows/main-branch-uploads.yml#L20

Yes, they are different images. The p4-spec repo was using p4lang/p4-spec-asciidoc:latest while p4runtime uses p4lang/p4rt-asciidoc. Both are separate Docker images. The p4runtime workflow at that link still uses the Docker-based approach, it might be worth applying the same native install approach there as well if consistency is desired, but that would be a separate PR in the p4lang/p4runtime repository.

Devansh-567:
This is ready to merge whenever you have a chance. Thanks for the review!

jafingerhut:
@rcgoodfellow Any preferences on this?  I do not recall the source or work flow used to create the Docker image that is currently used to build the Asciidoc source of the spec documents.  This PR stops using that Docker image, instead installing Asciidoctor on each CI run.  It takes a little bit of time on the CI machine, but the whole process is still at most 5 minutes long, so seems reasonable to me.

PR #1412, opened 2026-05-24
Remove Dockerfiles that are no longer used

Devansh-567:
Both tools/Dockerfile.asciidoc_v1 and tools/Dockerfile.asciidoc_v2
had broken RUN command chaining in the gem install and rouge build
steps.

Issues fixed:
- gem install commands used `\` (line continuation) instead of `&&\`,
  meaning failures would be silently ignored and subsequent installs
  would still run
- `cd rouge \` did not persist to the next line, so `gem build` and
  `gem install` were running in the wrong directory
- Replaced hardcoded `rouge-4.4.0.gem` with `rouge-*.gem` wildcard
  to match whatever version was built from source

These bugs would cause the Docker image to silently produce an
incorrect build environment.


jafingerhut:
Since you recently made changes so that the Docker images are not used to run asciidoctor, I guess the bigger question is whether there is any reason to keep these files around any longer, or should they just be removed?

Devansh-567:
> Since you recently made changes so that the Docker images are not used to run asciidoctor, I guess the bigger question is whether there is any reason to keep these files around any longer, or should they just be removed?

That's a fair point. Since the any-branch-uploads.yml workflow no longer uses the Docker image, these Dockerfiles appear to have no remaining users in the repository. I am happy to replace this PR with one that removes both files entirely if that is preferred. Just let me know and I will update the branch

jafingerhut:
> > Since you recently made changes so that the Docker images are not used to run asciidoctor, I guess the bigger question is whether there is any reason to keep these files around any longer, or should they just be removed?
> 
> That's a fair point. Since the any-branch-uploads.yml workflow no longer uses the Docker image, these Dockerfiles appear to have no remaining users in the repository. I am happy to replace this PR with one that removes both files entirely if that is preferred. Just let me know and I will update the branch

Please do, and thanks.  I'd rather not maintain scripts that likely no one is using now, nor likely to use in the future.

jafingerhut (approved):

Devansh-567:
This is ready to merge whenever you have a chance. Thanks for the review!

jafingerhut:
> This is ready to merge whenever you have a chance. Thanks for the review!

It seems best to wait to merge this until after https://github.com/p4lang/p4-spec/pull/1410 is approved and merged.

Issue #1389, opened 2026-03-16
What is the semantics of a `ternary` match on `bool` and `error` types?


jaehyun1ee:
Two tests in `p4c` use a `ternary` match on `bool` and `error` types. However, it is unclear what its semantics should be.

The specification mentions about `ternary` as:

> a ternary match kind on a key field means that the field in the table specifies a set of values for the key field using a value and a mask. The meaning of the (value, mask) pair is similar to the P4 mask expressions, as described in Section [8.15.3](https://p4.org/wp-content/uploads/sites/53/2024/10/P4-16-spec-v1.2.5.html#sec-cubes): a key field k matches the table entry when k & mask == value & mask.

Thus, a `ternary` match makes sense for numeric types, but I'm not sure what its semantics is for non-numeric types, and whether they should be allowed at the first place.

Below are the two tests in `p4c` that exhibit such use case:

* [`metrics/metrics_test_4.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/metrics/metrics_test_4.p4)

```p4
struct meta_t {
        /* ... omitted ... */
        error err_field; // error code
}

control ingress(
    inout headers_t hdr,
    inout meta_t meta,
    inout standard_metadata_t standard_metadata)

    table custom_table {
        key = {
            /* ... omitted ... */
            meta.err_field : ternary;
        }
        actions = {
            set_output;
            set_headers;
            my_drop;
        }
        default_action = my_drop();
    }
}
```

`meta.err_field` has `error` type, matched with `ternary` in `custom_table`.

* [`psa-bool-ternary-const-entry-bmv2.p4`](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/psa-bool-ternary-const-entry-bmv2.p4)

```p4
    table test_table {
        key = {
            hdr.ethernet.isValid() : ternary;
        }
        actions = {
            DummyAction;
            NoAction;
        }
        const entries = {
            (true) : DummyAction();
        }
    }
```

`hdr.ethernet.isValid()` has `bool` type, matched with `ternary` in `test_table`.

jafingerhut:
Doing a ternary match against a field with type `error`, or any non-serializable enum type, doesn't semantically make sense from the language spec perspective, since the spec does not define any bit pattern representation for those types, explicitly and by design.  It would be better if attempting to compile such a program gave an error.

[...]

PR #1413, opened 2026-05-27
Clarify that Booleans are serializable

jaehyun1ee:
Following the LDWG discussion on #1389, this PR clarifies in **7.1.4. The Boolean Type** that boolean types are serializable, where `true` means `1` and `false` means `0`.

Booleans were actually already implied to be serializable in the original spec, as `bool` was allowed within `header` types. Thus, this PR makes this point explicit by adding a sentence to the spec.


jafingerhut (approved):
LGTM

Issue #1327, opened 2024-11-04
Allow division and modulo on fixed-width integers


vlstill:
Currently, the P4<sub>16</sub> spec v1.2.5 only defines `/` and `%` operation (division and modulo) for arbitrary-width integers (`int`): see [8.8. Operations on arbitrary-precision integers](https://p4.org/wp-content/uploads/2024/10/P4-16-spec-v1.2.5.html#sec-varint-ops).

However, the P4C reference compiler allows these operations also on `bit<N>` type (fixed-width unsigned), and can perform constant folding and even strength reduction of them. For that reason, the user can write a code like this:

```p4
my_hdr1.fld_a = my_hdr0.fld_b / 16;
```

and the code is strength-reduced to 

```p4
my_hdr1.fld_a = my_hdr0.fld_b >> 4;
```
and compiles just fine. Indeed, if the target does not invoke `P4::CompileTimeOperations` midend pass, or otherwise prohibit these operations, and it implements proper code-generation for them, the target may choose to allow fully-runtime division and modulo. Of course, such an operation may not be available on all targets.

---

I think a more reasonably way to specify this is to allow division and modulo on `bit<N>` type, and to explicitly state in spec that on certain targets, only division/modulo by a constant (possibly a power-of-two constant) may be allowed. I don't see any strong reason to prohibit targets from supporting these ops.

Furthermore, for the sake of symmetry and elimination of exceptions, I would like to extend these operations on `int<N>` type too. Here, however, we need to define precisely how division modulo should behave for negative numbers as different languages have different behavior. I am not sure if there is one definition which is considered better.

jonathan-dilorenzo:
Discussed in the P4-LDWG, we decided that there are 3 levels of this:
1. Allow only constant folded away bit<N> divisions and modulo.
2. Allow any bit<N> divisions and modulo, noting that targets may restrict this behavior arbitrarily. **Suggestion from P4 LDWG is to send a PR to support this.**
3. Allow int<N> modulo and division as well.

linuxrocks123:
@jonathan-dilorenzo what would you do about runtime division by zero for levels (2) and (3)?

rcgoodfellow:
Consensus here is that levels (2) and (3) for exceptional conditions like divide by zero, this is architecture dependent. Future major versions of P4 may decide to formalize exceptional behavior, but this cannot be addressed in P4-16 beyond architecture-specific behavior.

PR #1414, opened 2026-05-27
Allow division and modulo operations on fixed-width bit (unsinged) types

jaehyun1ee:
This is a follow-up PR from #1327.

The current specification only allows division and modulo for arbitrary-precision integer types (`int`), but not for `bit<W>` or `int<S>` types. Yet, there were uses of division and modulo on `bit<W>` and `int<S>` types in p4c tests (where p4c allows such, under certain p4c-specific restrictions).

This was first discussed in Nov-2025 LDWG meeting, where @jonathan-dilorenzo summarized as:

> Allow any bit divisions and modulo, noting that targets may restrict this behavior arbitrarily. Suggestion from P4 LDWG is to send a PR to support this.

However, no follow-up PR was made, thus this has gone through another round of discussion in May-2026 LDWG meeting. The consensus is the same: allow division and modulo for `bit<W>` types, yet disallow for `int<S>` types since there is no clear consensus on how division and modulo should be defined regarding signedness and rounding.

This PR adds division and modulo to **8.7. Operations on fixed-width bit types (unsigned integers)**, also mentioning that P4 architectures may restrict division and modulo.

### DIFF
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -3636,6 +3636,12 @@ to bit-strings of the same width:
   and is computed by truncating the result to the output's width.
   P4 architectures may impose additional restrictions --- e.g., they may only
   allow multiplication by a non-negative integer power of two.
+* Truncating integer division between positive values, denoted by `/`.
+  The result has the same width as the operands. Similar to multiplication,
+  P4 architectures may impose additional restrictions.
+* Modulo between positive values, denoted by `%`.
+  The result has the same width as the operands. Similar to multiplication,
+  P4 architectures may impose additional restrictions.
 * Bitwise "and" between two bit-strings of the same width, denoted by `&`.
 * Bitwise "or" between two bit-strings of the same width, denoted by `|`.
 * Bitwise "complement" of a single bit-string, denoted by `~`.
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -3636,7 +3636,7 @@ to bit-strings of the same width:
   and is computed by truncating the result to the output's width.
   P4 architectures may impose additional restrictions --- e.g., they may only
   allow multiplication by a non-negative integer power of two.
-* Truncating integer division between positive values, denoted by `/`.
+* Truncating integer division, denoted by `/`.
   The result has the same width as the operands. Similar to multiplication,
   P4 architectures may impose additional restrictions.
 * Modulo between positive values, denoted by `%`.
--- a/p4-16/spec/P4-16-spec.adoc
+++ b/p4-16/spec/P4-16-spec.adoc
@@ -3637,6 +3637,7 @@ to bit-strings of the same width:
   P4 architectures may impose additional restrictions --- e.g., they may only
   allow multiplication by a non-negative integer power of two.
 * Truncating integer division, denoted by `/`.
+  The denominator must be positive.
   The result has the same width as the operands. Similar to multiplication,
   P4 architectures may impose additional restrictions.
 * Modulo between positive values, denoted by `%`.


p4-16/spec/P4-16-spec.adoc
```diff
@@ -3636,6 +3636,12 @@ to bit-strings of the same width:
   and is computed by truncating the result to the output's width.
   P4 architectures may impose additional restrictions --- e.g., they may only
   allow multiplication by a non-negative integer power of two.
+* Truncating integer division between positive values, denoted by `/`.
```
jafingerhut: The numerator for division or modulo should be allowed to be 0, not only positive, yes?  I agree it is reasonable to require the denominator to be positive, and not attempt to define the result of such an operation with a denominator of 0.

jafingerhut (changes requested):

p4-16/spec/P4-16-spec.adoc
jaehyun1ee: Thank you for the suggestion, I made the following edit:

```
* Truncating integer division, denoted by `/`.
  The denominator must be positive.
```

jonathan-dilorenzo (approved):

jonathan-dilorenzo:
LGTM, anything else you want to change here @jafingerhut or shall we resolve the conversation and merge this?

p4-16/spec/P4-16-spec.adoc
```diff
@@ -3636,6 +3636,13 @@ to bit-strings of the same width:
   and is computed by truncating the result to the output's width.
   P4 architectures may impose additional restrictions --- e.g., they may only
   allow multiplication by a non-negative integer power of two.
+* Truncating integer division, denoted by `/`.
+  The denominator must be positive.
+  The result has the same width as the operands. Similar to multiplication,
+  P4 architectures may impose additional restrictions.
+* Modulo between positive values, denoted by `%`.
```
jonathan-dilorenzo, line 3643: Allow the numerator to be 0 here.

jafingerhut: Recommendation: Change "the denominator must be positive." with "The denominator must not be 0."  The same restriction should exist for the modulo operator (the current language suggests that the numerator cannot be 0, but we should allow that).

p4-16/spec/P4-16-spec.adoc
jonathan-dilorenzo, line 3789: Ideally allow 0 for numerator both here and in truncating integer division.

jafingerhut (approved):
LGTM, subject to Jonathan's final review on small changes discussed in the 2026-Sep-14 LDWG meeting.

p4-16/spec/P4-16-spec.adoc
jaehyun1ee: Thanks! Addressed as follows:

```adoc
* Truncating integer division, denoted by `/`.
  The denominator must not be zero.
  The result has the same width as the operands. Similar to multiplication,
  P4 architectures may impose additional restrictions.
* Modulo, denoted by `%`.
  The denominator must not be zero.
  The result has the same width as the operands. Similar to multiplication,
  P4 architectures may impose additional restrictions.
```

p4-16/spec/P4-16-spec.adoc
jaehyun1ee: Also addressed here:

```diff
- * Truncating integer division between positive values, denoted by `/`.
- * Modulo between positive values, denoted by `%`.
+ * Truncating integer division, denoted by `/`.
+   The numerator must be non-negative and the denominator must be positive.
+ * Modulo, denoted by `%`.
+   The numerator must be non-negative and the denominator must be positive.
```

jaehyun1ee:
Thank you for the comments! @jonathan-dilorenzo, can you take a look?

jafingerhut (approved):

jonathan-dilorenzo:
Love it! Thanks @jaehyun1ee and @jafingerhut!

PR #1417, opened 2026-07-05
psa: document @groupaction annotation for ActionSelector groups

sanket-jadhav-cse:
## Summary

Documents the `@groupaction` annotation in the PSA specification for use with ActionSelector group actions in P4Runtime one-shot action selector programming.

The feature was added to P4Runtime, and the follow-up discussion concluded that this annotation is architecture-specific and should be documented in the PSA specification rather than the base P4 language specification.

## Changes

- Adds a `Group Actions` subsection to PSA Section 8.12, `Action Selector`.
- Documents that group actions execute logically before the selected member action.
- Defines `@groupaction` as required for group actions and restricts it to group-action use only.
- Documents that `@groupaction` cannot be combined with `@tableonly` or `@defaultonly`.
- Documents the one-shot P4Runtime limitation.
- Adds a `PSA Reserved Annotations` appendix entry for `@groupaction`.

## Validation

- `git diff --check`
- `make -C p4-16/psa`

The build completed successfully. The environment emitted optional dependency warnings for `rouge` and `prawn-gmagick`; these do not affect this documentation change.

## Related

- p4lang/p4runtime#594
- p4lang/p4runtime#605
- p4lang/p4-spec#1392


p4-16/psa/PSA.adoc
```diff
@@ -2437,6 +2437,29 @@ to select a member from the group, and the action specification
 corresponding to that member is applied. The dynamic selection algorithm is
 specified as a parameter when instantiating the action selector.
 
+==== Group Actions
+
+This feature is currently supported only for P4Runtime{P4RuntimeAPI}
+one-shot action selector programming.
+
+A group may be configured with a _group action_ that executes logically before
```
matthewtlam, line 2445: Is  `_group action_` a formatting thing?

p4-16/psa/PSA.adoc
```diff
@@ -2437,6 +2437,29 @@ to select a member from the group, and the action specification
...
+the selected member action. An action used as a group action must be annotated
+with `@groupaction`.
+
+An action annotated with `@groupaction` may only be used as a group action. It
+must not be used as a table entry action or as a table default action.
```
matthewtlam: Maybe we should make it explicit since english is not an obvious language. But we would say it cannot be used as a table entry and cannot be used a a table default action. The other way could be misinterpreted depending on the brackets or how the user reads it

p4-16/psa/PSA.adoc
```diff
@@ -2437,6 +2437,29 @@ to select a member from the group, and the action specification
...
+
+An action that is not annotated with `@groupaction` must not be used as a
+group action.
+
+The `@groupaction` annotation is not allowed in combination with `@tableonly`
+or `@defaultonly`. (For the semantics of `@tableonly` and `@defaultonly`,
+see the P4 language specification.)
+
+Because a single named action cannot be used both as a group action and as
```
matthewtlam: `single named` --> `non-group`

matthewtlam (commented):
Thanks for a nice first attempt @Sanketjadhav31. The language and semantics of this is a bit tricky

matthewtlam:
@jfingerh PTAL to see if we have the same understanding

p4-16/psa/PSA.adoc
jafingerhut: According to AsciiDoc documentation, it is used to make the enclosed text italic: https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/

p4-16/psa/PSA.adoc
jafingerhut: Perhaps the following might be more explicit, as a suggested replacement for the last sentence in the paragraph above?

As specified in the P4Runtime specification, attempting to use an action annotated with `@groupaction` as a table entry action or as a table default action must fail and return an appropriate error status.

p4-16/psa/PSA.adoc
jafingerhut: It also seems clear to simply say "Because an action ..." with no adjectives before "action"

p4-16/psa/PSA.adoc
matthewtlam: +1 to what the suggestion replacement

jafingerhut (approved):
Looks good to me.

jafingerhut:
@matthewtlam Please take a look at the most recent version to see if you approve.

matthewtlam (approved):
Thanks @Sanketjadhav31. LGTM!

